PR #3006, opened 2026-04-21
fix: add duplicate checks to junction entity services

### LINKED ISSUE #2977: Feature: Properly handle role/permission creation as upsert (update or insert)
### Context
In a system setup combining Authup and a given application, the setup is such that on start-up, the application tries to set up its role/permissions model in Authup. Each permission and role has a unique name defined from the application domain, which thus serves as identifier for the given permission/role. The application tries to perform an "upsert" (update or insert) using HTTP `PUT` semantics (https://www.rfc-editor.org/rfc/rfc9110.html#name-put) and the unique permission name as "external ID" (i.e. not the one assigned internally by Authup, but by the application as client)

`PUT .../permissions/{unique_permission_name}  { ... }`

These upsert semantics are in principle correctly handled by Authup, but when the requests lead to an actual update (instead of insert), the Authup logs show "errors" like:

... query failed: INSERT INTO auth_role_permissions ...
error: Error: Duplicate entry ... 

These seeming "errors" are misleading from the sys admin view, given that they create the feeling that something is wrong in the system, when in fact the upsert operation was successful.

### Suggested behaviour

How about avoiding these misleading false positive "errors" by making the upsert logic truly part of the Authup application logic?
From the current behaviour, it seems to be that every incoming PUT request is directly translated to an SQL `INSERT` without further checks whether the attribute used as external (client) ID is already present, such that an SQL `UPDATE` would be required instead. Accordingly, the `INSERT` then fails.

### Alternative

Alternatively, the "duplicate entry" messages should be suppressed or logged on another level (e.g. WARN?) to avoid confusing sys admins with false positive "errors".

tada5hi:
closes #2977

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents duplicate assignments across multiple entity types (roles, permissions, scopes, mappings) with pre-create uniqueness checks and clearer conflict errors.
  * Fixes realm/validation edge cases to reject invalid cross-realm mappings.

* **New Features**
  * Introduces a dedicated service and validation for identity-provider role mappings with full CRUD and permission-aware checks.
  * New/expanded validators improve input validation and error messages.

* **Refactor**
  * Centralized creation/update/delete authorization and persistence into services; controllers now delegate to services.

* **Tests**
  * Expanded unit coverage for services, validation, and conflict/permission scenarios.


tada5hi:
@coderabbitai review

PR #3018, opened 2026-04-26
perf(server-core): batch role permission bindings to remove N+1

### LINKED ISSUE #3016: RoleRepositoryAdapter.getBoundPermissionsForMany has N+1 queries
## Pre-existing issue surfaced during PR review

**Reported in:** PR #3015, review comment by @Copilot
**File:** `apps/server-core/src/app/modules/database/repositories/role/repository.ts#154`

## Description

`getBoundPermissionsForMany(entities)` calls `getBoundPermissions(entity)` once per role and flattens the results — N+1 queries for N roles. This is on the permission-evaluation hot path (every authenticated request hits it via `IdentityPermissionProvider.getForUser`/`getForRobot`/`getForClient`).

The 60 s redis cache on each `auth_role_permissions` lookup mitigates steady-state load, but the cold path (cache miss after a write or process start) issues N round-trips.

The same shape was inherited from the legacy `RoleRepository` (now retired); this issue is being filed as a follow-up rather than expanding the scope of the per-entity-port refactor.

## Suggested fix

Single query that loads all junction rows for the given role ids, plus a batched `findDescendantsTreeById` to avoid the secondary N+1 on the policy side. Sketch:

```ts
async getBoundPermissionsForMany(entities: (string | Role)[]): Promise<PermissionPolicyBinding[]> {
  const ids = entities.map((e) => (typeof e === 'string' ? e : e.id));
  if (ids.length === 0) return [];
  const entries = await this.repository.manager
    .getRepository(RolePermissionEntity)
    .find({ where: { role_id: In(ids) }, relations: { permission: true } /* + cache key per id-set */ });
  // ... collect distinct policy_ids, batch-load, map to bindings
}
```

This may require batched policy-tree loading (separate concern).

tada5hi:
closes #3016

### DIFF
--- a/apps/server-core/src/app/modules/database/repositories/bindings.ts
+++ b/apps/server-core/src/app/modules/database/repositories/bindings.ts
@@ -5,6 +5,9 @@
  * view the LICENSE file that was distributed with this source code.
  */
 
+import {
+    In,
+} from 'typeorm';
 import type {
     EntityManager,
     EntityTarget,
@@ -30,6 +33,8 @@ type PermissionJunction = {
     policy_id?: string | null;
 };
 
+const IN_CLAUSE_CHUNK_SIZE = 500;
+
 export async function loadBoundPermissions<E extends PermissionJunction & ObjectLiteral>(
     options: LoadBoundPermissionsOptions<E>,
 ): Promise<PermissionPolicyBinding[]> {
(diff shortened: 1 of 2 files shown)


tada5hi:
Reverting this batched approach — it traded a small cold-path win for a hot-path regression by dropping the per-role Redis cache. The right fix for #3016 needs a custom caching layer (replacing TypeORM's QueryResultCache with an ICache-backed abstraction over hydrated entities + invalidation hooks across all 23 subscribers), which is wider scope than this PR. Closing in favor of a follow-up that does the cache refactor first, then revisits #3016 with a clean hybrid (per-id cache check + batched miss-fill).

PR #3022, opened 2026-04-28
fix: enforce state for public clients in code flow at /authorize

tada5hi:
## Updates

* Public clients using OAuth2 authorization code flow must now include the state parameter; requests without it will be rejected
* Issuer URLs are normalized to consistently remove trailing slashes when constructing realm-specific endpoints


tada5hi:
@coderabbitai review

PR #3066, opened 2026-05-20
refactor!: drop user.client_id, backfill canonical identifier columns

tada5hi:
## Summary

- Drops the unused `client_id` / `client` columns and relation from `User` / `UserEntity`. The column has been functionally dead since migration `1766830857009` deleted every row where it was non-null ("user client_id referenced wrong table"). The provisioning synchronizer drops the now-redundant `client_id` lookup key.
- Adds paired `postgres` + `mysql` migration `1779267068441-Default` that bundles:
  - **Canonical-name backfill** (consolidates plan 008): collision pre-check that aborts on any duplicate group after `LOWER(TRIM(name))`, then `UPDATE … = LOWER(TRIM(…))` across nine name tables (`auth_clients`, `auth_robots`, `auth_users`, `auth_realms`, `auth_roles`, `auth_scopes`, `auth_permissions`, `auth_policies`, `auth_identity_providers`) and `auth_users.email`. Pre-check `GROUP BY` mirrors each entity's `@Unique([...])` declaration.
  - **Column drop**: auto-generated DDL dropping the `auth_users.client_id` FK, index, and column.
- `down()` restores the column / index / FK only; canonical `UPDATE`s are not reversed (lossy by definition — original casing is not recoverable).

## Breaking change

`User.client_id` and `User.client` are removed from `@authup/core-kit`. Note the `[key: string]: any` escape hatch on the `User` interface means any consumer reading these fields still typechecks but returns `undefined` at runtime.

## Test plan

- [x] `npm run build --workspace=apps/server-core` — clean
- [x] `npm run test --workspace=apps/server-core` — 799/799 passed
- [x] `npx eslint --fix` on all changed files — clean
- [x] Smoke-tested migration `up` / `down` / re-`up` against fresh Postgres + MySQL containers — see below
- [x] Verified canonical UPDATE rewrites mixed-case rows correctly on both DBs
- [x] Verified collision pre-check aborts cleanly on a deliberately seeded duplicate

### Smoke-test details

- Postgres: fresh DB → migration run (8/8 applied) → `auth_users.client_id` confirmed dropped → revert → column restored with correct FK target (`auth_clients`, not the originally-buggy `auth_realms`) → re-apply → column dropped again. ✅
- MySQL: same cycle. ✅
- Canonical UPDATE on Postgres: seeded `'TestRealm'` + `'Foo@Bar.COM'`, ran migration → names and emails canonicalized to lowercase. ✅
- Collision detection on Postgres: seeded `'CollideMe'` + `'collideme'` (Postgres is case-sensitive, both rows persist) → migration aborted with actionable error: `Canonical-name migration aborted: "auth_realms" has 1 collision group(s) on LOWER(TRIM(name))`. ✅
- **Bug caught during MySQL smoke test** (fixed in `ec8972dfa`): MySQL's default `utf8mb4_0900_ai_ci` collation treats `'TestRealm' = 'testrealm'`, so the original `WHERE name <> LOWER(TRIM(name))` predicate was always false on mixed-case rows and the UPDATE silently no-op'd. Switched the WHERE clause to `WHERE BINARY name <> LOWER(TRIM(name))` (and same for `email`) to force byte-level comparison. Re-verified end-to-end. ✅

## Notes

- Migration bundles two unrelated changes (column drop + canonical backfill). Considered splitting; kept bundled because the canonical backfill was already ledger'd in plan 008 awaiting consolidation, and shipping both in one timestamp avoids a second migration run for operators.
- `down()` for canonical backfill is intentionally a no-op (cannot recover original casing); subsequent writes continue to canonicalize, so the schema stays consistent.
- ⚠️ **CI does not exercise migration files** — the test runner uses `dataSource.synchronize()` instead. The MySQL `BINARY` bug above was only caught by manual smoke testing. Opened #3067 to add a dedicated CI job that runs migrations end-to-end. Until that lands, every migration PR should be smoke-tested manually against both DBs.

## Summary by CodeRabbit

* **Bug Fixes**
  * User names and email addresses are now automatically normalized (converted to lowercase with whitespace trimmed) for improved data consistency.

* **Refactor**
  * Simplified user record structure by removing client associations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/authup/authup/pull/3066?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/server-core/src/adapters/database/domains/user/entity.ts
+++ b/apps/server-core/src/adapters/database/domains/user/entity.ts
@@ -18,8 +18,7 @@ import {
     Unique,
     UpdateDateColumn,
 } from 'typeorm';
-import type { Client, Realm, User } from '@authup/core-kit';
-import { ClientEntity } from '../client/index.ts';
+import type { Realm, User } from '@authup/core-kit';
 import { RealmEntity } from '../realm/index.ts';
 
 @Entity({ name: 'auth_users' })
@@ -176,19 +175,6 @@ export class UserEntity implements User {
     @ManyToOne(() => RealmEntity, { onDelete: 'CASCADE' })
     @JoinColumn({ name: 'realm_id' })
     realm: Realm;
-
-    // ------------------------------------------------------------------
-
-    @Index()
-    @Column({ nullable: true })
-    client_id: Client['id'] | null;
-
-    @ManyToOne(() => ClientEntity, {
-        onDelete: 'CASCADE',
-        nullable: true, 
-    })
-    @JoinColumn({ name: 'client_id' })
-    client: Client | null;
 
     // ------------------------------------------------------------------
 
--- a/apps/server-core/src/adapters/database/migrations/mysql/1779267068441-Default.ts
+++ b/apps/server-core/src/adapters/database/migrations/mysql/1779267068441-Default.ts
@@ -43,7 +43,7 @@ export class Default1779267068441 implements MigrationInterface {
             await queryRunner.query(`
                 UPDATE \`${table}\`
                 SET \`name\` = LOWER(TRIM(\`name\`))
-                WHERE \`name\` <> LOWER(TRIM(\`name\`))
+                WHERE BINARY \`name\` <> LOWER(TRIM(\`name\`))
             `);
         }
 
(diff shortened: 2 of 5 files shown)


tada5hi:
@coderabbitai review

PR #3104, opened 2026-06-09
feat: per-realm web client login, realm chooser & backend-served auth workflow UI

tada5hi:
## Summary

This branch turns the login entrypoint into a per-realm flow and makes server-core fully self-sufficient for every auth workflow (headless deployments included).

### Per-realm `web` client + realm-selection login
- Auto-provisions a public OAuth2 client named `web` for **every** realm — at startup (incl. pre-existing realms) and on runtime realm creation. Public (`is_confidential: false`, PKCE required), `built_in`, `scope: global openid`, one `<origin>/**` redirect entry per trusted app origin. No migration — provisioning is idempotent (MERGE, guarded on `built_in`) and reconciles on every startup.
- client-web `/login` is now a Keycloak-style realm chooser (single realm auto-selects; client-side search above 8 realms). Selecting a realm runs authorization-code + PKCE against `client_id=web&realm_id=<id>`; the global routing interceptor verifies `state` and exchanges the code.
- `built_in` clients auto-consent in the SSR `/authorize` UI (with fallback to the manual Allow/Deny form if the auto-submit fails). `web`/`system` are reserved client names; the validator strips `built_in` so no API caller can self-assign it.
- `ADDITIONAL_DOMAINS` config extends the trusted app origins; this set drives **both** the `web` client's redirect allowlist and the CORS origin allowlist.

### Backend-served auth workflow UI + status endpoint
- `/register`, `/activate`, `/password-forgot`, `/password-reset` now serve SSR pages from the embedded UI (`GET`), while `POST` on the same paths stays the JSON API. Pages render a localized "disabled" notice instead of 404 when a workflow is off (stale mail links shouldn't dead-end).
- New public status endpoint `GET /` → `{ version, date, features: { registration, passwordRecovery, emailVerification } }` (`StatusResponse` in `core-http-kit`, `client.status.get()`); the same flags ride the SSR hydration payload.
- New kit form components (`ALoginForm`, `ARegisterForm`, `AActivateForm`, `APasswordForgotForm`, `APasswordResetForm`) + shared `AAuthShell` chrome; client-web auth pages use a dedicated chrome-less `auth` layout.
- Workflow links carry a same-origin `redirect` param so "back to login" restores the original `/authorize` request; `sanitizeRelativeRedirect()` rejects absolute / protocol-relative URLs (open-redirect guard).

### Mail-template renderer
- Workflow mails go through a new `IMailTemplateRenderer` port: localized copy (en/de/fr/es, BCP-47-narrowed), html-escaped interpolation, branded shell, multipart (html + text).
- When `publicUrl` is set, mails deep-link to the SSR pages (`/activate?token=…`, `/password-reset?token=…&realm_id=…`); the raw code stays in the body for copy/paste, no PII in URLs.

### Locale ownership refactor
- `@vuecs/locale` is now the source of truth for the UI locale (cookie-backed `vc-locale`, browser auto-resolve, `<html lang>` sync); ilingo follows one-way via `syncTranslatorLocaleFromManager`. The kit's language switcher works with or without vuecs-locale installed.

### Fixes & hardening
- XSS / PKCE / realm / config hardening across the auth workflow flows; redirect + mail-link escaping.
- Tokens issued via a client keep the user's global roles.
- Realm grid no longer flashes the untranslated "no results" state during SSR/first paint (skeleton is gated on first load completion).

## ⚠️ Behavioral change (deploy note)

CORS previously **reflected all origins**; it is now restricted to `publicUrl` + `additionalDomains`. In **production**, if client-web is served from a different origin than the API's `PUBLIC_URL`, you **must** add that origin to `ADDITIONAL_DOMAINS` or both CORS and the login redirect (`<origin>/**`) will reject it. In non-production, `http://localhost:3000` is seeded automatically.

Security: the `web` client is `built_in` + `global` scope, so any allowlisted origin can obtain a full-permission user token after login — only list origins you control.

## Audit notes (non-blocking)

Verified end-to-end: `/authorize` verifier resolves the per-realm `web` client via `findOneByIdOrName(client_id, realm_id)`; `GET /realms` is unauthenticated so the logged-out realm grid loads; `<origin>/**` matches `/login/callback`; token-exchange param shapes match. Known minor smells left as-is:
- `ARealmGrid` caps at 100 realms (alphabetical; client-side search mitigates, no paging).
- `exchangeAuthorizationCode` doesn't emit `LOGGED_IN` (no current subscribers; latent).
- Runtime realm-create logs and swallows a web-client provisioning failure (self-heals at next startup) rather than rolling back the realm.

## Test plan

- [x] `npm run test --workspace=apps/server-core` green
- [x] Builds pass, eslint clean
- [ ] Manual: click through `/login` realm selection → consent auto-submit → token exchange → land on home, in both dev and a prod-like split-origin setup
- [ ] Manual: confirm `ADDITIONAL_DOMAINS` correctly gates CORS + redirect in production
- [ ] Manual: register / activate / password-forgot / password-reset SSR pages, incl. disabled-feature notices and mail deep links

## Summary by CodeRabbit

* **New Features**
  * Expanded auth UX: multi‑realm login, callback redirect, chrome‑less auth layout, hero graphic, realm picker, full register/activate/password‑forgot/reset flows, and auto‑consent for system clients
  * Improved OAuth2 UX: PKCE-based flow, persisted authorize state, smoother token exchange, and automatic per‑realm public web client provisioning

* **Locale & Theme**
  * Cookie-backed locale and color‑mode with language switcher and auth gadgets

* **Security**
  * Safer relative-redirect validation and safer email deep‑link handling when public URL set

* **Documentation**
  * Updated deployment and provisioning guidance for trusted origins

* **Tests**
  * New/updated unit tests covering origins, provisioning, mail templates, redirects, and status features


tada5hi:
@coderabbitai pause

Issue #3159, opened 2026-06-29
isSuperset / grantDominates does not compare policy content — latent over-permit on differing policy-bound grants


tada5hi:
## Summary

`IdentityPermissionProvider.isSuperset` (and its helper `grantDominates`) gates role/identity assignment by checking, per permission, that every grant in the target (child) is **dominated** by some grant the actor (parent) holds. The realm reach is compared exactly, but the **policy content is not compared** — any two policy-bound grants are treated as mutually dominating. This is a latent **over-permit** (fail-open) when the actor and target hold *different* policies on the same permission.

This is **pre-existing** (the collapsed `isSuperset` used the same `parentHasPolicy && !childHasPolicy` approximation); PR #3158 preserved it verbatim per-grant when moving to the disjunction model. Filing as a follow-up rather than expanding that PR's scope.

## Where

`apps/server-core/src/core/identity/permission/module.ts`

```ts
// grantDominates (~line 226)
function grantDominates(parent: PermissionGrant, child: PermissionGrant): boolean {
    if (compareRealmScope(parent.realm_scope, child.realm_scope) < 0) {
        return false;
    }
    if (parent.policy && !child.policy) {   // <-- only this approximation
        return false;
    }
    return true;                            // both policy-bound => "dominated", regardless of content
}
```

## Why it matters (concrete)

- Actor (parent role) holds `user_update` gated by policy `P_actor` = `ATTRIBUTES { department: "X" }`.
- Target role (child) grants `user_update` gated by policy `P_target` = `ATTRIBUTES { department: "Y" }` (or any non-subset policy).
- `grantDominates`: `parent.policy` present **and** `child.policy` present → not `(parent && !child)` → returns `true` → the child grant is considered dominated.
- Result: `isSuperset` permits the assignment, so the actor confers reach over `department=Y` users it does not itself hold. **Privilege escalation across disjoint policy scopes.**

The reverse direction is already handled correctly: a policy-free child is *not* dominated by a policy-bound parent (the actor cannot grant unrestricted from a restricted holding).

## Why it isn't a contained fix

Deciding `P_actor ⊇ P_target` for arbitrary policy trees (composite / attributes / identity / realm-match / time, with `invert`) is undecidable in general. Options to scope a fix:

1. **Exact identity:** require the same `policy_id` (or structural equality) for domination — simplest, conservative (may under-permit legitimate equal-but-distinct policies).
[...]

Issue #3155, opened 2026-06-27
Merge of mixed policy-free + policy-bound grants under-grants realm reach


tada5hi:
## Problem

When an actor holds two grants for the **same** permission with different scope+policy — e.g. `(realm_scope: own, no policy)` and `(realm_scope: any, IDENTITY policy)` — `mergePermissionPolicyBindings` collapses them into a single unrestricted binding whose `realm_scope` is folded from the **policy-free subset only** (`own`). The `(any ∧ IDENTITY)` reach is lost, so the actor is denied cross-realm access it legitimately holds.

This is **fail-closed** (an under-grant, not a leak), which is why it was deprioritised — but it is a correctness gap.

## Why it's not a contained fix

The correct semantics are a disjunction: access iff `∃ binding . (realm_scope_i matches resource) ∧ (policies_i pass)`. Representing that requires one of:

1. **Multiple merged bindings per permission key** — but all four consumers of `mergePermissionPolicyBindings` assume **one binding per key**:
   - `PermissionBindingPolicyEvaluator` → `bindingsMerged[0]`
   - `IdentityPermissionProvider.isSuperset` → `parentMerged.find(...)`
   - `IdentityPermissionProvider.resolveJunctionGrant` → `merged[0].policies[0]`
   - `MemoryPermissionProvider.setMany` → `reduce` to `prev[key] = current`
   Changing the invariant silently breaks all four.
2. **Encoding `realm_scope` into the policy expression** — reintroduces the original fail-open widening the enum was introduced to fix (a policy-free binding drops the whole policy composite under the AFFIRMATIVE merge, which would also drop the scope condition).

A contained option is an `additionalGrants: {scope, policies}[]` field on the merged binding that **only** the evaluator ORs (preserving one-per-key for the other three consumers); `isSuperset` would still under-grant the mixed case (fail-closed).

## References

- PR #3151 — introduced `realm_scope` + the fail-open merge fix that surfaced this.
- `packages/access/src/permission/helpers/merge.ts` (the policy-free-subset scope fold)
- `apps/server-core/src/core/security/policy/evaluator.ts`, `core/identity/permission/module.ts`, `packages/access/src/permission/provider/memory.ts` (the four consumers)
- Pairs with #3154 (PolicyData / evaluation-context redesign).

PR #3164, opened 2026-06-30
fix(server-core)!: select junction grant relative to requested reach (#3160)

### LINKED ISSUE #3160: resolveJunctionGrant collapses the actor's grant disjunction into one ceiling — mixed-grant actors under-propagate
## Summary

`IdentityPermissionProvider.resolveJunctionGrant` returns a single `{ policy, realmScope }` "ceiling" used to cap junction propagation. `selectCeilingGrant` picks the actor's **most-permissive single grant** (highest realm reach, policy-free preferred on a tie). When an actor holds *mixed* grants for one permission, that single ceiling cannot represent all of them, so the consumer caps reach to the ceiling and **inherits the ceiling's policy** — which can be more restrictive than another grant the actor genuinely holds.

This is **fail-closed** (a mixed-grant actor under-propagates; it never confers more than it holds), so it is not a security leak. It is the write-side analogue of the realm-reach disjunction work (#3155 / PR #3158, plan 036): the read path (evaluator) and `isSuperset` now evaluate the full disjunction, but `resolveJunctionGrant` is the one place that still collapses it to a single value. Lower priority than #3159 (which is fail-open).

## Where

- `apps/server-core/src/core/identity/permission/module.ts` — `resolveJunctionGrant` (~L83-129) and `selectCeilingGrant` (~L242)
- `apps/server-core/src/core/identity/permission/types.ts` — `ResolveJunctionGrantResult`
- Consumers (cap + inherit-policy logic, create + update): `role-permission`, `client-permission`, `robot-permission`, `user-permission` services (8 sites)

## Why it matters (concrete)

Actor holds `user_update` as **both** `(own, no-policy)` and `(any, IDENTITY-policy)`.

`selectCeilingGrant` → `(any, IDENTITY)` (highest reach). Creating a junction with requested `realm_scope = own` and no explicit policy:

```ts
validated.realm_scope = minRealmScope([own, any]);              // = own  (correct)
if (grant.realmScope !== ANY || grant.policy) {                 // any !== any => false; grant.policy truthy => true
    validated.policy_id = grant.policy ? grant.policy.id : null; // = IDENTITY.id  (forced on)
}
// junction = (own, IDENTITY)
```

The actor holds `(own, no-policy)` and could legitimately confer it ungated, but the ceiling forces the `IDENTITY` policy onto the own-scoped junction. The recipient gets a **more restricted** grant than the actor was entitled to give.

## Why it isn't a contained fix

A faithful version must choose the source grant by the *requested* `(scope, policy)`, not a global "ceiling". That means either:

1. Pass the requested `(realm_scope, policy_id)` into `resolveJunctionGrant` and return the least-restrictive **held** grant that dominates it (then the consumer just validates, not re-derives), or
2. Return the full grant disjunction and move the cap/inherit decision into a shared helper the 8 call sites use.

Either changes the `ResolveJunctionGrantResult` contract and the cap/inherit logic across the 8 junction-service sites. Pairs with #3159 (the `isSuperset` policy-content approximation) under a single "make write-side propagation disjunction-faithful" effort.

## Acceptance

- A mixed-grant actor can propagate the least-restrictive grant it actually holds that covers the requested junction (no spurious policy inheritance / reach narrowing).
- Still fail-closed: never confer a `(scope, policy)` not dominated by a real held grant.
- Extend `apps/server-core/test/unit/core/identity/permission/provider.spec.ts` and the junction-service specs with mixed-grant propagation cases.

tada5hi:
## Summary

Closes #3160.

`IdentityPermissionProvider.resolveJunctionGrant` collapsed the actor's grant
**disjunction** into a single global *ceiling* (`selectCeilingGrant`: highest
`realm_scope`, policy-free preferred on a tie). When an actor held *mixed* grants
for one permission, that single ceiling could not represent all of them, so the
consumer capped reach to the ceiling and **inherited the ceiling's policy** —
which can be more restrictive than another grant the actor genuinely holds.

This was the last write-side site still folding the disjunction to one value —
the read path (evaluator) and `isSuperset` already evaluate the full disjunction
(#3155 / #3158, #3159).

### Concrete bug

Actor holds `user_update` as **both** `(own, no-policy)` and `(any, IDENTITY-policy)`.
Creating an `own`-scoped junction with no explicit policy → `selectCeilingGrant`
picks `(any, IDENTITY)` (highest reach) → the junction is capped to `own` but
**forced to carry the `IDENTITY` policy**. The actor could legitimately confer
`(own, no-policy)` ungated; the recipient gets a more restricted grant than the
actor was entitled to give. (Fail-closed, so not a leak — but wrong.)

## Fix

- `resolveJunctionGrant` now selects the grant **relative to the requested reach**
  (`selectGrantForRequest`): each held grant is ranked by its `realm_scope`
  **capped to** the requested scope (highest capped reach wins, policy-free preferred
  on a tie), so a lower-scoped policy-free grant beats a higher-scoped policy-bound
  one when both cap to the same reach. The selected grant is returned **uncapped**,
  preserving the consumer's own `min` cap and its "actor is unrestricted-`any`?"
  check that decides whether an explicit `policy_id` may stand.
- `ResolveJunctionPolicyOptions` gained `realmScope?` — the requested junction reach.
- The four junction services (`role`/`user`/`client`/`robot-permission`) pass it on
  **create** (`validated.realm_scope`) and **update** (`data.realm_scope ?? entity.realm_scope`).

For the example above: an `own` request now selects `(own, no-policy)` (ungated);
an `any` request selects `(any, IDENTITY)` (reaching `any` requires the gated grant).

### Additional fix — update-path fail-OPEN (surfaced while implementing)

In the junction **update** path, `realm_scope` was capped only when
`data.realm_scope` was present. A restricted actor (e.g. a `realm_admin` holding
only `(own, no-policy)`) doing a **policy-only** update on an existing `(any, P)`
binding kept `realm_scope = any` while dropping/replacing the policy — persisting
`(any, no-policy)`, a binding **no held grant dominates** (privilege escalation).
Now a restricted actor that touches the binding re-caps the existing reach down to
its own grant. Unrestricted (`any`, policy-free) actors are unaffected.

## Tests

- `provider.spec.ts`: 8 request-aware `resolveJunctionGrant` cases incl. mixed-grant
  own-vs-any selection and a composite-policy avoidance case.
- `user-permission/service.spec.ts`: regression tests for the update fail-open and
  its asymmetry (unrestricted actor may still drop policy without narrowing reach).
  Mutation-checked: the regression test fails when the re-cap is removed.
- Full `apps/server-core` suite green: **938 passed**.

## Notes

Marked `!` for consistency with the realm-scope series (#3151/#3157/#3158/#3159):
this changes runtime authorization outcomes for mixed-grant actors and closes a
fail-open update path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Permission create/update now correctly applies `realm_scope` capping and `policy_id` inheritance/clearing, preventing unintended broader access when policy settings change.
  * Junction grant selection is now based on the requested reach (with safer fail-closed behavior when only non-propagatable policies apply).
* **Tests**
  * Expanded unit test coverage for junction grant creation and update-data building, including edge cases for policy-only updates, realm-scope reach behavior, member gating, and input validation.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -919,14 +919,14 @@ An actor with both `admin` (unrestricted) and `realm_admin` (restricted) grants 
 
 ### Junction Policy Propagation
 
-When creating any permission-binding junction (role-permission, user-permission, client-permission, robot-permission):
+When creating or updating any permission-binding junction (role-permission, user-permission, client-permission, robot-permission):
 
-1. The service calls `this.identityPermissionProvider.resolveJunctionGrant(identity, { name, realmId, clientId })`.
-2. It aggregates the actor's grants for that permission and selects the **ceiling** = the actor's most-permissive single grant (`selectCeilingGrant`: highest `realm_scope`, policy-free preferred on a tie).
-3. The new junction is capped to that ceiling: `realm_scope = min(requested, ceiling.realm_scope)`, and the ceiling's **own** `policy` (its `id`) is propagated as `policy_id` — never the target's. (If the ceiling is policy-restricted but its policy is not a propagatable `Policy` — e.g. an id-less composite — it fails closed to `realm_scope: none`.)
-4. If `data.policy_id` is already set explicitly, propagation is skipped.
+1. The service calls `this.identityPermissionProvider.resolveJunctionGrant(identity, { name, realmId, clientId, realmScope })`, passing the **requested** reach (`validated.realm_scope ?? own` on create; `data.realm_scope ?? entity.realm_scope` on update — the *resulting* junction reach, so a policy-only update can't silently widen).
+2. It aggregates the actor's grants for that permission and selects the grant **relative to the requested reach** (`selectGrantForRequest`, #3160): each grant is ranked by the reach it can confer *for this request* — its `realm_scope` capped to `realmScope` — so a lower-scoped policy-free grant beats a higher-scoped policy-bound grant when both cap to the same requested reach (highest *capped* reach, policy-free preferred on a tie). This is **not** a global "ceiling" — a mixed-grant actor (e.g. `own`+no-policy *and* `any`+policy) propagates its policy-free `own` grant for an `own` request instead of inheriting the wider grant's policy.
+3. The selected grant is returned **uncapped**; the new junction is then capped by the consumer: `realm_scope = min(requested, selected.realm_scope)`, and the selected grant's **own** `policy` (its `id`) is propagated as `policy_id` — never the target's. (If the selected grant is policy-restricted but its policy is not a propagatable `Policy` — e.g. an id-less composite — it fails closed to `realm_scope: none`. A clean lower-scoped grant covering the request avoids that fail-closed.)
+4. Returning the selected grant uncapped preserves the "only an unrestricted (`any`, policy-free) actor may set an explicit `policy_id`" rule: that check reads the selected grant's uncapped `realm_scope`/`policy`, so it still fires exactly when the actor genuinely holds an `any` policy-free grant.
 
-This prevents privilege escalation: a `realm_admin` cannot create unrestricted permission bindings. Because the actor only ever propagates its *own* policy (not the target's), this path needs no policy-content comparison — the asymmetry with the superset check, which must compare against fixed target grants.
+This prevents privilege escalation: a `realm_admin` cannot create unrestricted permission bindings, and (post-#3160) a mixed-grant actor neither under-propagates (spurious policy inheritance on a narrow request) nor over-propagates (riding a wider grant's reach with a narrower grant's policy). Because the actor only ever propagates its *own* policy (not the target's), this path needs no policy-content comparison — the asymmetry with the superset check, which must compare against fixed target grants.
 
 ## Self-Edit Pattern (declarative field denylists)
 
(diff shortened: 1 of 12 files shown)


tada5hi:
@coderabbitai review

tada5hi:
### Follow-up commit: dedup + update-path hardening

A multi-agent audit of the first commit surfaced a pre-existing fail-open on the junction **update** path (a restricted actor could keep a wider pre-existing `realm_scope` on a policy-only edit). That is fixed, and the security-critical cap/inherit/re-cap rule — previously copy-pasted across all four `*-permission` services — is now a single shared pure module (`core/identity/permission/junction-grant.ts`), so the privilege-escalation boundary is audited and changed once rather than four times.

Also in this commit:

- **Member-permission gate on `update()`** — an actor may only modify a binding for a permission it itself holds (mirrors `create()`; previously absent on update).
- **UPDATE-group validator on update data** — rejects a malformed `realm_scope` / non-UUID `policy_id` (instead of silently coercing) and strips create-only FKs.
- **Deterministic tie-break** in `selectGrantForRequest` — selection between equally-reaching policy-bound grants no longer depends on repository order.

A second independent audit of this commit returned **no security/correctness findings** (3 confirmed items, all `nit`: a benign stricter-validation contract change plus test-coverage notes — both now covered). Full `apps/server-core` suite: **954 passing**, including new exhaustive unit coverage for the shared helpers, the tie-break, the member gate, and update-validation rejection.

tada5hi:
@coderabbitai review

PR #3168, opened 2026-06-30
feat(client-web-kit): realm_scope UI follow-up — labels + assignment-time scope (plan 034)

tada5hi:
Plan 034 — realm_scope UI follow-up, **goal G1 (labels + hints)**.

## Problem

The junction realm-scope selector in `APermissionPolicyBindingButton` rendered the **raw enum values** as button labels — admins literally saw `none` / `own` / `ownOrNull` / `any`. For an authorization-**reach** control that's a misconfiguration hazard, and `none` (a grant that matches no realm) was offered as a selectable option.

## Change

- New i18n keys (`REALM_SCOPE_OWN` / `_OWN_OR_NULL` / `_ANY` + matching `*_HINT`) on `TranslatorTranslationClientKey`, authored across **all four locales** (en/de/fr/es; the locale-parity test enforces this). "realm" follows each locale's existing term (de `Realm`, fr `domaine`, es `dominio`).
- The selector now renders the **localized label** per button, the **hint as a tooltip**, and a **hint legend** for the currently-selected scope (keyed off the string value of `REALM_SCOPE`, so `ownOrNull` maps correctly).
- **`none` is no longer offered** — it remains a valid stored/migration value but is not user-selectable (a dead grant shouldn't be a one-click option).

## Scope

`realm_scope` lives only on the four `*-permission` junctions, surfaced through this one shared control, so this is the entire UI surface for G1.

## Follow-up

Goal **G2 (assignment-time scope)** — letting a user pick the reach when *assigning* a permission (today the junction is created at the `own` default and edited afterwards) — is a separate change with a UX design choice still to settle (where the picker lives); it will follow once that's decided.

## Verification

i18n build + locale-parity test green (77); client-web-kit builds clean; lint clean.

## Summary by CodeRabbit

* **New Features**
  * Permission assignment controls now support both creating new bindings and editing existing ones more smoothly.
  * Added realm-scope options with localized labels and hints in multiple languages.
* **Bug Fixes**
  * Improved permission assignment behavior so new bindings can be saved even when no existing record is present.
  * Existing permission changes now update immediately and reflect the latest saved state in the UI.

### DIFF
--- a/packages/client-web-kit/src/components/entities/client-permission/AClientPermissionAssignment.vue
+++ b/packages/client-web-kit/src/components/entities/client-permission/AClientPermissionAssignment.vue
@@ -8,7 +8,7 @@
 <script lang="ts">
 import { EntityType } from '@authup/core-kit';
 import type { ClientPermission } from '@authup/core-kit';
-import { defineComponent } from 'vue';
+import { computed, defineComponent } from 'vue';
 import {
     defineEntityManager,
     defineEntityVEmitOptions,
@@ -61,14 +61,28 @@ export default defineComponent({
             return manager.delete();
         };
 
+        const handleCreated = (entity: ClientPermission) => {
+            manager.created(entity);
+        };
+
         const handleUpdated = (entity: ClientPermission) => {
             manager.updated(entity);
         };
 
+        // The existing junction in edit mode, else a stable create template (FK base) carrying
+        // no id → the binding control treats it as create mode. Memoized so the template keeps a
+        // stable reference while unassigned.
+        const bindingEntity = computed<Partial<ClientPermission>>(() => manager.data.value || {
+            client_id: props.clientId,
+            permission_id: props.permissionId,
+        });
+
         return {
             manager,
             handleChanged,
+            handleCreated,
             handleUpdated,
+            bindingEntity,
             EntityType,
         };
     },
@@ -82,10 +96,10 @@ export default defineComponent({
             @changed="handleChanged"
         />
         <APermissionPolicyBindingButton
-            v-if="manager.data.value"
-            :key="manager.data.value.id"
+            :key="manager.data.value?.id || 'create'"
             :entity-type="EntityType.CLIENT_PERMISSION"
-            :entity="manager.data.value"
+            :entity="bindingEntity"
+            @created="handleCreated"
             @updated="handleUpdated"
         />
     </span>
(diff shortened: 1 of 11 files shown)


tada5hi:
### Added: G2 — assignment-time `realm_scope`

`APermissionPolicyBindingButton` now drives **both** create and edit through a single `entity` prop (the form-initialValue pattern):

- **edit mode** — `entity` has an `id` → an existing binding; each selection patches immediately (unchanged behavior).
- **create mode** — `entity` has no `id` → a *template* carrying the FK base fields; the modal stages a chosen `(realm_scope, policy_id)` and the control **performs the create itself** via the injected client, emitting `created` for the parent to sync (`manager.created`).

So a permission can be assigned with the intended reach instead of being born at the `own` default and edited afterwards.

**Two design points addressed in review:**
- The control no longer self-manages *update* while merely emitting for *create* — create is performed in the control too, symmetric with update.
- No second prop: create context rides on the same `entity` prop as a no-`id` template (a memoized computed in each assignment component keeps the template reference stable, so the staged selection isn't reset on re-render).

The four `*-permission` assignment components render the control in both states and forward `created`/`updated` to `manager`.

client-web-kit builds clean; lint clean; i18n parity green.

tada5hi:
### Audit + fixes

A multi-agent audit of this PR found a **blocker** in the G2 create flow (now fixed):

- **Blocker — create never transitioned to edit mode.** The record entity manager's `created()` only emitted; unlike `updated()`/`deleted()`/`create()` it never set `entity.value`. So the new create-via-binding-button path persisted the junction server-side but left `manager.data` undefined — the row stayed in create mode (toggle off), and a second Add / toggling on issued a **duplicate create**. Fixed by having `created()` adopt the record as `entity.value` (consistent with the other mutators and the socket `onCreated`, which is wired to `created`).
- **Medium — a11y.** The icon-only binding trigger had no accessible name. Added a mode-aware `aria-label`/`title`.

Verified-clean by the audit: edit-mode behavior unchanged (no regression); create payload spread is correct; types are sound (`Partial<…Permission>` template, `RealmScopeValue`); i18n keys/parity correct.

Left as follow-ups (not blocking):
- **Design (low):** two create affordances now appear on each unassigned row (the quick toggle = default `own`, and the sliders = staged scope). Defensible as progressive disclosure, but worth a maintainer call on whether to keep both.
- **i18n (nit):** the German realm-scope copy uses "Realm" while the de entity noun is "Organisation" — a pre-existing cross-namespace term mismatch in `de/client.ts` (en/fr/es are consistent). Out of scope here.

> Note: the blocker fix is reasoning-verified and consistent with the manager's own patterns, but the create→edit transition has not yet been clicked through in a running app.

PR #3020, opened 2026-04-28
fix: enforce OAuth2 client authentication and PKCE per RFC 6749/7636

tada5hi:
## Release Notes

* **New Features**
  * Added OAuth2 client authentication across authorization, password, and refresh token grant flows.
  * Implemented PKCE (Proof Key for Code Exchange) requirement for public clients during authorization.
  * Enhanced refresh token security with client binding and authentication enforcement.

* **Improvements**
  * Improved client credential extraction and validation across token endpoints.
  * Added comprehensive error handling for OAuth2 validation failures.

* **Tests**
  * Expanded test coverage for client authentication scenarios and PKCE validation.

PR #3177, opened 2026-07-02
fix: realm-scoping and provisioning-validation follow-ups

tada5hi:
Implements all six follow-up items deferred from #3175 (plan 038).

## 1. Fail-closed realm-key predicates (repo-wide)

New `IRealmRepository.resolveId(key)` — UUID pass-through (binding an unknown UUID matches zero rows), canonicalizing name lookup, `null` ⇒ caller fails closed. Swept across all 12 resolve-then-filter blocks in the 8 repository adapters, the permission/policy checker services, and the realm-resolver middleware. Previously 9 of those blocks silently **dropped** the realm filter for unknown realm keys, so `GET /realms/<unknown-uuid>/users/<name>` returned a cross-realm name match; it now 404s (pinned by tests).

## 2. Unified realm-hint semantics across token grants

`authorization_code` and `refresh_token` now share the password grant's realm-hint semantic via a common `readRealmHint` helper: `realm_id ?? realm_name`, canonicalized at the ingress, resolved once with master fallback, scoped onto client-by-name resolution. This makes the refresh leg deterministic — a password login via a name-identified client refreshes against the same realm's client instead of an unscoped lookup matching an arbitrary same-named client (every realm has a built-in `web` client, so collisions were guaranteed). The refresh grant resolves lazily; a bare refresh does no realm lookup. `client_credentials` / `robot_credentials` and the `/authorize` code-request verifier keep raw `realm_id` semantics (documented non-goal).

## 3. client-web-kit test infrastructure

vitest + @vue/test-utils + happy-dom stood up in `packages/client-web-kit` (first component tests in the monorepo). Regression suites pin the interactive-login realm plumbing that shipped the same one-line bug class twice in #3175: picker-selected realm transmitted; `codeRequest.realm_id` transmitted including a late-arriving prop; empty realm omitted from the grant; plus a DOM-free spec for the `store.login()` gate.

## 4. Realm lookup caching on `/token`

`RealmRepositoryAdapter.findOneById` is query-cached (60s, `CachePrefix.REALM` id key — byte-identical to the existing realm subscriber invalidation key). Name-key caching stays deferred (rename invalidation prerequisite). The new invalidation tests surfaced a real pre-existing bug: `EntitySubscriber.afterRemove` read `event.entity`, whose primary key TypeORM strips on remove — so id-keyed cache invalidation (and DELETED event payloads) were broken for all subscribers. Fixed by preferring `event.databaseEntity`.

## 5. UUID identifiers skip the realm predicate — decided: intended

Documented (architecture.md + resolver comment) and pinned by test: UUIDs are globally unique, the realm hint constrains NAME resolution only, and the issued token carries the identity's true realm. Enforcing the predicate would break realm-less id-identified logins because the master fallback erases hint explicitness.

## 6. File-provisioning validation was a no-op

Nested provisioning validators ran group-less, executing zero core-kit field validators. New `ValidatorGroup.PROVISIONING`: identifier fields mounted `[CREATE, PROVISIONING]`, `built_in` only under PROVISIONING (API groups keep stripping it), user email optional under PROVISIONING (synchronizer backfills). Nested runs pass the group explicitly, assign validated output back, and prefix array indices onto issue paths. Top-level `policies` — documented but silently dropped from the validated output until now — are validated via the new `PolicyProvisioningValidator` and provisioned (exercised end-to-end by the integration suite). `BaseProvisioningSynchronizer.canonicalizeName` stays as defense in depth for the non-validating default/programmatic sources.

## Verification

- Full monorepo build green; `npm run test` green across all workspaces (server-core: 97 files / 992 tests, incl. new HTTP, service-level, and pure validator suites; client-web-kit: 7 new tests; core-kit: 6 new group tests)
- Also de-flaked a pre-existing LDAP test race (two specs sharing one `cn=foo` account on the shared OpenLDAP container)

## Known follow-up (out of scope)

The kit's silent-refresh path sends a bare `refresh_token` without `client_id`, so sessions with client-bound tokens already die at the RFC 6749 §6 binding check (pre-existing since #3020). A kit-side fix (send `client_id` + realm on refresh) can now build on this PR's scoped name lookup and test infrastructure.

BREAKING CHANGE: a name-identified client on the authorization_code or refresh_token grant without a realm hint now always resolves in the master realm (pass `realm_id`/`realm_name` or the client UUID for other realms). Invalid file-provisioning entities now abort startup instead of being silently accepted, and attribute keys outside the entity schema are stripped.

## Summary by CodeRabbit

* **New Features**
  * Added support for startup provisioning fields like `built_in`, improved realm-scope settings, and richer policy/permission relations in provisioning files.
  * Login and token flows now accept realm hints more consistently across password, refresh, and authorization-code flows.
  * Added client-web-kit test tooling and login form coverage.

* **Bug Fixes**
  * Realm-scoped lookups now fail closed when a realm can’t be resolved, preventing cross-realm matches.
  * Token, permission, and provisioning behavior now preserves validated data more reliably, including delete handling and cache updates.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -556,6 +556,32 @@ The provisioning system declaratively synchronizes entities (permissions, roles,
   - `composite/`: Merges multiple sources with dedup by composite key (`name:realm_id:client_id`)
 - **app/modules/provisioning/module.ts**: `ProvisionerModule` — creates shared repository adapter instances and wires them to synchronizers
 
+### File-Source Validation (`ValidatorGroup.PROVISIONING`)
+
+Only the **file** source validates (the default and programmatic sources are
+code-built and bypass validation — which is why
+`BaseProvisioningSynchronizer.canonicalizeName` stays as defense in depth).
+`FileProvisioningSource` runs `RootProvisioningValidator` with
+`ValidatorGroup.PROVISIONING` and **uses the validated output**: every nested
+entity run (`createProvisioningEntitiesValidator`,
+`core/provisioning/entities/utils.ts`) passes the group explicitly (zod
+check closures don't inherit the validup run group), assigns the result back
+(so validator transforms — canonicalization, stripping — reach the
+synchronizers), and prefixes the array index onto issue paths.
+
+The PROVISIONING group lives in the core-kit entity validators alongside
+CREATE/UPDATE: identifier fields (`name`, `realm_id`, policy `type`) are
+mounted `[CREATE, PROVISIONING]`; `built_in` is mounted **only** under
+PROVISIONING (the API groups deliberately strip it — no HTTP service ever
+runs the PROVISIONING group); `user.email` is optional under PROVISIONING
+(the user synchronizer backfills a placeholder) while staying required at
+CREATE. Consequences for file configs: invalid entities now fail startup
+(fail-closed — the load throws before anything synchronizes), unmounted
+attribute keys are stripped, and top-level `policies` (previously silently
+dropped from the validator output) are validated via
+`PolicyProvisioningValidator` (attributes + `extraAttributes` + recursive
+`children`) and provisioned.
+
 ### Synchronization Order
 
 `ProvisionerModule` runs (1) `GraphProvisioningSynchronizer`, (2) backfill via `assignDefaultPolicy` (config-gated, deprecated).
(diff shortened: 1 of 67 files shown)


tada5hi:
@coderabbitai pause

PR #3187, opened 2026-07-06
feat: session-management API (list + revoke sessions)

tada5hi:
Plan 016 **PR G (backend + typed client)** — a REST surface over `auth_sessions` so identities can see and revoke sessions ("active devices" / force logout). Stacked on #3186 (refresh-token rotation); **base is `feat/refresh-token-rotation`**, so this diff is the session-management change only. The client-web UI (`/settings/sessions`, admin `/users/:id/sessions`) is a **follow-up (PR G2)**.

## Endpoints

Dual-mounted `@DController(['/sessions', '/realms/:realmId/sessions'])`:

| Route | Behavior |
|---|---|
| `GET /sessions` | List. **Self-service** without `SESSION_READ` (force-scoped to own sessions via a mandatory owner `WHERE`); with `SESSION_READ`, realm-gated per row. |
| `GET /sessions/:id` | Read one. Own → no permission; else `SESSION_READ` + realm-match. `@me`/`@self` → the caller's current session. |
| `DELETE /sessions/:id` | Revoke one. Own → no permission; else `SESSION_DELETE` + realm-match. |
| `DELETE /sessions` | Revoke every own session **except the current one** ("log out my other devices"). Self-service; the current session id comes from the bearer token, never the client. |

## Design

- **Dedicated `SESSION_READ` / `SESSION_DELETE`** (not reusing `USER_*`/`CLIENT_*`/`ROBOT_*`, because a session's subject is polymorphic). Auto-provisioned (enum-iterated); granted to `admin` (`any`) and `realm_admin` (`ownOrNull` read / `own` delete).
- **Ownership** = `session.sub === actor.identity.data.id && session.sub_kind === actor.identity.type`, so self-service is inherent (no extra permission).
- `SessionRepository` gains `findMany` / `findAllByOwner` (list bypasses the id-keyed cache). Delete routes through the cache-aware `remove` → the DB delete **cascade-drops the session's `auth_session_tokens` rows**, so a force-logout also invalidates the subject's refresh tokens.
- Typed `client.session` API in `@authup/core-http-kit`.

## Security hardening (found by an adversarial review of this PR)

`SessionRepository.findMany` **force-selects `realm_id` / `sub` / `sub_kind`** (`qb.addSelect`) regardless of the client `fields` projection. The per-row realm gate reads `realm_id`, and rapiq honors a `fields` projection over `default` — so without this, an own/ownOrNull-scoped `SESSION_READ` holder could request `?fields[session]=id,sub` (omitting `realm_id`) and neutralize the `realm_scope` reach factor, leaking cross-realm sessions. `session-realm-isolation.spec.ts` is the regression guard (verified: it fails without the `addSelect`).

> **Note (out of scope, pre-existing):** the same field-projection pattern affects the existing `robot` / `client` / `user` `getMany` paths (their `resourceRealmMatch` per-row gate reads a `realm_id` a client can project away). Not fixed here — flagged for a separate hardening pass.

## Testing

- Service spec (13) — self-scoping, per-row realm drop, self vs. permission for get/delete, delete-all-but-current.
- HTTP specs (4) — admin list + per-session revoke, delete-all-but-current, non-admin self-scoping, and the cross-realm field-projection regression.
- Full server-core suite **1071 tests pass**; full monorepo build clean.

Docs: `.agents/architecture.md` → *Session Management API*.

## Summary by CodeRabbit

* **New Features**
  * Added session management API (list, get, and revoke) with `@me`/`@self` support and force-logout of other devices.
  * Extended the client with a dedicated sessions domain API.
  * Added `SESSION_READ` and `SESSION_DELETE` permission names, and documented the Session Management API.

* **Bug Fixes**
  * Enforced realm-aware session access and projections isolation during session listing/reads.
  * Ensured revocation keeps the active session when revoking “all others”.

* **Tests**
  * Added unit and regression tests for authorization, revocation behavior, and realm isolation.


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

PR #3194, opened 2026-07-07
feat: realm-bind the authorize and token flow

tada5hi:
## Problem

A downstream app lets the user pick a realm and redirects to `/authorize` with that realm's client. If the browser still carries an authup session for **another** realm (e.g. the user logged in to RealmA earlier and logged out only in the downstream app — authup-origin cookies are untouched), authup silently continued as the RealmA identity: because every realm's `web` client is `built_in`, the consent step auto-submitted, and the server performed **no** identity-realm ↔ client-realm check. The downstream app asked for RealmB and received RealmA tokens — a confused deputy whose artifact carried a RealmA `iss`/signing-key with a RealmB `aud`.

## Fix

An identity may only authorize, redeem a code for, or refresh a token against a client in **its own realm**, enforced server-side at three independent points (the kit UI is UX only — the server is authoritative):

1. **`POST /authorize` issuance** — `OAuth2Authorization.authorize()` throws `login_required` (`OAUTH_LOGIN_REQUIRED`, HTTP 400, no identity data in the body — no realm-enumeration oracle) when the identity realm ≠ the client realm.
2. **`/token` code redemption** — the code verifier's new `realmId` option (fed `client.realm_id`) rejects `code.realm_id != client.realm_id` with `invalid_grant`. Covers identity-provider-callback codes and any in-flight pre-deploy codes.
3. **`/token` refresh parity** — a **public** client refreshing a token whose realm differs from the client's realm → `invalid_grant` (kills legacy cross-realm `web`-client refresh tokens). Confidential clients stay exempt — the secret proves identity, and the documented cross-realm password grant relies on it.

Also included:
- **Realm-mismatch UI** — the SSR `/authorize` page renders a "sign in to {realm}" / "return to application" card (the latter only when the `redirect_uri` matched a registered pattern) instead of auto-consenting, gated so a `built_in` client can't auto-consent before the store resolves the identity's realm.
- **Clickjacking guard** — all SSR auth pages send `Content-Security-Policy: frame-ancestors 'none'` + `X-Frame-Options: DENY`.
- **Deterministic client resolution** — a name-identified client at `/authorize` now requires a realm hint (`invalid_request` otherwise); every realm has a `web` client, so a bare name was ambiguous.

## Testing

Full `apps/server-core` suite (1111 tests) green, plus `@authup/i18n` locale parity (77) and `@authup/client-web-kit` (11). New tests: scenario-1 reproduction (`login_required`) + same-realm control, the code-verifier realm branch, the refresh-parity reject + control, and the name-hint guard. Two pre-existing tests that minted codes via a cross-realm master-admin authorize were updated to authorize as a same-realm user (their fixture relied on the now-closed path).

## Breaking change

Master-realm admins can no longer authorize into another realm's app via the built-in `web` client. Previously-issued cross-realm artifacts were malformed (RealmA `iss`/key + RealmB `aud`), so intentional reliance is implausible — use realm-local accounts.

## Scope / follow-ups

This is PR A of the plan in `.agents/plans/041-authorize-session-binding.md`. It fully closes the cross-realm scenario. The account-switch UX (`prompt=select_account` + chooser) and cross-app logout via the standard `end_session_endpoint` are PRs B and C — the latter is the RP-agnostic mechanism that ends a lingering authup session on a downstream logout (kit and non-kit RPs alike), so this PR deliberately does not add a kit-specific session-delete to `store.logout()`.

## Summary by CodeRabbit

* **New Features**
  * Added realm-aware sign-in handling, including a clearer “different realm” message and a way to switch accounts.
  * Auth screens now show the target realm and whether a redirect back to the app is safe.

* **Bug Fixes**
  * Prevented authorization, code exchange, and refresh flows from crossing realm boundaries in unsupported cases.
  * Added stronger protection for server-rendered auth pages against being embedded in other sites.

* **Chores**
  * Added new user-facing error messages and translations for multiple languages.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1064,6 +1064,47 @@ Two-layer rejection:
 
 `AttributeNamesPolicyValidator` reads the policy's `names` field from extra-attributes (`policy_attributes`). For top-level policies bound directly to permissions, the policy is loaded as the root of a closure-table descendants tree. `EATreeRepository.findDescendantsTree()` calls `extendOneWithEA(entity)` after building the children — without that, the root entity's EA fields stay unloaded and the validator fails with "value_invalid". Both Layer 1 (`PermissionDatabaseProvider`) and Layer 2 (`bindings.ts`) depend on this fix.
 
+## Authorize Realm Binding (plan 041)
+
+The authenticated identity's realm MUST equal the client's realm — an identity
+cannot authorize (or redeem a code / refresh a token) against a client in
+another realm. Without this an identity with a lingering session for realm A,
+redirected to `/authorize` for realm B's `web` client (a downstream app's realm
+picker), silently minted realm-A tokens against realm B's client (confused
+deputy; the artifact carried realm-A `iss`/signing-key + realm-B `aud`).
+Enforced server-side at **three** points — the kit UI (realm-mismatch card in
+`Authorize.vue`) is UX only:
+
+1. **`POST /authorize` issuance** — `OAuth2Authorization.authorize()` throws
+   `OAuth2LoginRequiredError` (`ErrorCode.OAUTH_LOGIN_REQUIRED` / OIDC
+   `login_required`, HTTP 400, **no identity data in the body** — no
+   realm-enumeration oracle) when `identity.data.realm.id !== data.realm_id`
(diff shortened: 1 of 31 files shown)


tada5hi:
@coderabbitai pause

PR #3196, opened 2026-07-07
feat: add rp-initiated logout (end_session_endpoint)

tada5hi:
Stacked on #3195 (prompt surface). PR C of the plan in `.agents/plans/041-authorize-session-binding.md`. **Review/merge after #3195.**

## Problem

A downstream app that logs the user out locally had no standard way to also end the authup session — the "logout doesn't reach authup" half of the reported issue. This is deliberately *not* solved with a kit-specific `store.logout()` session delete (that stays local-only); the RP-agnostic answer is the OIDC standard `end_session_endpoint`.

## What this adds

`GET`/`POST /logout` (OIDC RP-Initiated Logout, advertised as `end_session_endpoint` in discovery, **no feature flag**). Core logic is `OAuth2EndSessionService`. It's the mechanism any RP — kit or non-kit React/other — redirects to on its own logout.

Security posture (all enforced, full unit-test matrix):

- **`id_token_hint`** is verified by the token verifier with a new **`ignoreExpiry`** option — signature, `nbf` and **`kind`** still apply; only `exp` is skipped (a logout hint is routinely expired). It threads down to server-kit's `verifyToken` (`validateExp: false`).
- A hint whose **`kind !== id_token` is rejected** — access/refresh tokens also carry `session_id`, so accepting them would let a leaked access token force a logout. `aud` is cross-checked against `client_id` when both are present.
- A verified hint carrying `sid` → the session is revoked **immediately**, but **only after** its `sub`/`sub_kind` match the hint's subject (never revoke someone else's session). Without a hint the endpoint mutates nothing — the SSR page's sign-out is a click-gated, bearer-authenticated `store.logout()`.
- **`post_logout_redirect_uri`** is honored only when it is absolute http(s) AND matches a registered client `redirect_uri` pattern (open-redirect guard); `state` rides only with a validated redirect, and an unverified hint's claims are never reflected.

Also: the SSR logout page (`AEndSessionForm` + `apps/server-core/ui/src/pages/logout.vue`), the discovery `end_session_endpoint`, and the typed `buildEndSessionURL`.

## Testing

Full `apps/server-core` suite (1136) + `@authup/server-kit` (10, the `verifyToken` change) + `@authup/i18n` parity + `@authup/client-web-kit` green. New `OAuth2EndSessionService` matrix: valid hint verified; forged hint rejected; **access-token-as-hint rejected** (kind check); `aud`/`client_id` mismatch rejected; registered redirect honored; unregistered + `javascript:` redirect dropped; sub-match revoke vs sub-mismatch no-revoke. Plus discovery `end_session_endpoint` assertion.

## Deferred follow-up

A dedicated `post_logout_redirect_uri` client column + migration + a `WebClientProvisioner` default + the client-web end-session round-trip — until then `/logout` validates against `redirect_uri` patterns.

## Residual (documented)

Keycloak parity: a *leaked* valid id_token can force-logout its own session (an annoyance, not privilege escalation) — mitigated by the sub-match requirement and the short id_token TTL.

## Summary by CodeRabbit

* **New Features**
  * Added OIDC RP-initiated logout support, including a new logout endpoint, session verification/revocation flow, and logout redirect handling.
  * Exposed logout metadata in OpenID provider configuration so clients can discover the logout endpoint.
  * Added a logout page and confirmation form in the UI, with support for immediate sign-out when the session was already revoked.
  * Added logout URL building support for client-side sign-out redirects.

* **Bug Fixes**
  * Tightened redirect validation to allow only safe absolute HTTP(S) logout redirects.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1142,6 +1142,39 @@ session rather than a freshly-created fallback one (documented residual).
 `prompt_values_supported` and **fixes** `revocation_endpoint` from `…/token` to
 `…/token/revoke` (RFC 7009 — an RFC 7009 POST to `/token` never worked).
 
+### RP-Initiated Logout — `end_session_endpoint` (plan 041 PR C)
+
+`GET`/`POST /logout` (discovery `end_session_endpoint`, **no feature flag**) is
+the RP-agnostic session-termination mechanism — the intended way a downstream
+app (kit or non-kit) ends a lingering authup session on its own logout, so
+`store.logout()` never needs a kit-specific session delete. Core logic is
+`OAuth2EndSessionService` (`core/oauth2/end-session/`), wired via
+`createLogoutController`. Security posture (all enforced, unit-tested matrix):
+
+- **id_token_hint** is verified by `OAuth2TokenVerifier` with a new
+  `ignoreExpiry` option — signature, nbf and (crucially) **kind** still apply;
+  only `exp` is skipped (a logout hint is routinely expired). The option threads
+  down to server-kit's `verifyToken` (`validateExp: false`). A hint whose `kind
+  !== id_token` is **rejected** (access/refresh tokens also carry `session_id`,
+  so accepting them would let a leaked access token force a logout). `aud` vs
+  request `client_id` cross-checked when both present.
+- A signature-verified hint carrying `sid` → the referenced session is revoked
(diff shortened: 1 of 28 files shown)


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

PR #3199, opened 2026-07-08
feat: drop implicit & hybrid oauth2 response types

tada5hi:
Plan 042 item **3**. Decided with the plan: fix the "implicit tokens land in the redirect **query** string" defect (RFC 6749 §4.2.2 requires the fragment) at the source — by dropping the implicit/hybrid response types entirely, per the OAuth 2.1 draft (which removes the implicit grant) and RFC 9700 (OAuth 2.0 Security BCP: implicit MUST NOT be used).

## What changes

- **`response_type=code` is the only supported response type.** The code-request validator rejects `token`, `id_token`, `none`, and any hybrid combination; `OAuth2Authorization.authorize()` enforces the same as defense in depth (`unsupported_response_type`).
- **The authorization response never carries tokens.** `confirm()` loses its `access_token`/`id_token` `searchParams` writes — previously those tokens leaked via browser history, proxy/access logs, and `Referer` (compounding the id_token-hint replay surface, see #3198). An openid-scoped code still mints its id_token onto the code blob (with `c_hash`); the `/token` exchange returns it — unchanged.
- **Discovery** `response_types_supported` now advertises only `code` (previously `code`, `token`, `none`).
- **Verifier tightened:** PKCE + `state` for public clients are now unconditional — the `willIssueCode` gate is dead code with a code-only pipeline (every call site is validator-guarded, including the identity-provider flow, which validates the embedded code request before verifying).
- `OAuth2AuthorizationResult` loses `accessToken`/`idToken`; the now-unused `accessTokenIssuer` was removed from the authorize controller/authorization contexts and wiring.
- client-web's client **URL generator** page now emits `response_type=code`.

## Error surface

At the HTTP layer a dropped response type yields the standard 400 validation shape (a `response_type` issue carrying the `unsupported_response_type` message) — consistent with how every other invalid `/authorize` parameter has always behaved. The precise OAuth2 error object is thrown by the core module (unit-tested). A spec-conformant *error redirect* back to the RP is deliberately not part of this PR — that redirect infrastructure lands with the `prompt=none` work (plan 042 item 10).

## Tests

- unit: rejection matrix (`token`, `id_token`, `none`, `code token`, `id_token token`) + code-issuance assertion in `authorization/module.spec.ts` (spec reworked onto `response_type=code` with a working code-issuer stub)
- unit: verifier spec's "implicit flow skips state" test flipped to a rejection pin
- e2e: `authorize.spec.ts` rewritten — code flow still works, each dropped type 400s with a `response_type` issue, and discovery advertises only `code`

Full server-core suite green (1163 tests; the known pre-existing `grant-authorize` FK flake documented in #3198 surfaced once across four runs, on an untouched code path).

**BREAKING:** any third-party RP using the implicit or hybrid flow against authup must migrate to the authorization code flow with PKCE. Authup's own kit, client-web, SSR pages, and federated IdP flow have always used the code flow — no first-party impact.

## Summary by CodeRabbit

* **New Features**
  * OAuth sign-in now uses authorization code flow only, with updated client links and discovery metadata to reflect the supported response type.
  * Public clients must always use PKCE and provide a state value.

* **Bug Fixes**
  * Redirects no longer include access or ID tokens.
  * Unsupported or implicit/hybrid response types are now rejected consistently.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1105,6 +1105,19 @@ mechanism. `store.logout()` stays local-only (token/cookie cleanup); it does not
 call `DELETE /sessions/@me` (that endpoint remains the session-management API for
 revoking a specific session from the sessions UI).
 
+### Response types — code only (plan 042 item 3)
+
+`response_type=code` is the **only** supported response type (OAuth 2.1
+posture). The implicit/hybrid response types (`token`, `id_token`, `none`) are
+rejected by the code-request validator (a `response_type` issue → 400) and,
+defense in depth, by `OAuth2Authorization.authorize()`
+(`unsupported_response_type`). The authorization response never carries tokens
+— an openid-scoped code stores its id_token on the code blob and the `/token`
+exchange returns it. Discovery `response_types_supported` advertises only
+`code`. Consequently the code-request verifier requires PKCE + `state` for
+public clients **unconditionally** (the former `willIssueCode` gate is gone —
+every verified request issues a code).
+
 ### OIDC prompt surface & id_token claims (plan 041 PR B)
 
 `/authorize` accepts the OIDC Core §3.1.2.1 params `prompt`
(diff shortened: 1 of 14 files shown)


tada5hi:
@coderabbitai review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

PR #3200, opened 2026-07-08
feat: mint the id_token at the token exchange for an authoritative sid

tada5hi:
Plan 042 item **6**. Stacked on #3199 (drop implicit) — base branch `feat/oauth2-drop-implicit`; retarget to `master` once #3199 merges.

## The defect

The id_token was minted at `/authorize` time from `options.sessionId` and passed through the auth-code blob verbatim to `/token`. Two consequences:

1. **Stale `sid`.** The `authorization_code` grant's `resolveSession` can back the issued tokens with a *different* session than the code's `session_id` — the #3191 sub/realm-mismatch fallback, or a session deleted in flight. The id_token, minted earlier, kept the old `sid` → revoke-by-`sid` (RP-initiated logout, #3196) targeted the wrong session.
2. **No id_token for federated logins.** Only `OAuth2Authorization.authorize()` minted an id_token, but the external-IdP callback (`IdentityProviderController`) issues its code directly, bypassing that path — so an `openid`-scoped federated login received **no id_token at all**.

## The fix

Mint the id_token inside `OAuth2AuthorizeGrant.runWith`, **after** `resolveSession`:

- `sid = session.id` (the real backing session) — authoritative in the reuse branch, the fallback-create branch, and the session-less federated flow alike.
- `auth_time` is captured at authorize time onto the auth-code blob as a new `OAuth2AuthorizationCode.auth_time` field (**replacing** the removed `id_token` field — the code is a Redis cache blob, so **no migration**), and read back at the exchange.
- `at_hash` (over the freshly-issued access token) and `c_hash` are recomputed at the exchange; `nonce` rides from the code.
- `OAuth2Authorization.authorize()` no longer mints the id_token, and drops its `openIdTokenIssuer` / `identityResolver` dependencies (and the `codeIssuer.updateIdToken` method is gone). The `openIdTokenIssuer` is instead wired into the `TokenController`'s authorize grant.

## Tests

- **Core grant unit** (`grant-types/authorize.spec.ts`): no id_token without `openid` scope; `sid` = reused session (not the code's `session_id` by coincidence) with `auth_time`/`nonce`/`at_hash` asserted; `sid` = the freshly-created session on the fallback branch (≠ the stale code id); id_token minted for a session-less (federated) openid code.
- **Authorization module unit**: `auth_time` = session `created_at` for an interactive authorize, `= now` for a session-less one, passed through to the code issuer.
- **e2e `grant-authorize.spec.ts`**: strengthened the existing `sid`/`auth_time` test to assert `id_token.sid === introspect(access_token).session_id` — the id_token and access token share one real session.
- **e2e `grant-idp-authorize.spec.ts`**: new test — an `openid` federated login now returns an id_token whose `sid` matches the access token's session (previously: no id_token).

Full server-core suite green (1170 tests).

## Summary by CodeRabbit

* **New Features**
  * OAuth and OpenID Connect now use authorization code flow only.
  * ID tokens are issued during the token exchange, with session-based `sid` and accurate `auth_time` claims.

* **Bug Fixes**
  * Redirects no longer include access tokens or ID tokens.
  * Public clients now consistently require PKCE and `state`.

* **Documentation**
  * Updated API guidance and discovery details to reflect code-only authorization and PKCE requirements.


tada5hi:
@coderabbitai review

tada5hi:
@coderabbitai full review

PR #3202, opened 2026-07-08
feat: dedicated post_logout_redirect_uri client column

tada5hi:
Plan 042 item **9** (was the "deferred follow-up" noted in #3196's PR C). Independent of #3198–#3201 — branches off master.

## The conflation

`OAuth2EndSessionService.isValidPostLogoutRedirect` validated the RP's `post_logout_redirect_uri` against the client's **login** `redirect_uri` patterns, conflating two distinct redirect surfaces. A dedicated column separates them.

## What changes

- **`post_logout_redirect_uri`** nullable `varchar(2000)` on `ClientEntity` + core-kit `Client` type + `ClientValidator` (comma-separated wildcard patterns, mirroring `redirect_uri`; mounted in every group; **not** in the `system.client-names-self-manage` denylist; **not** in the trimmed `ClientSummary` DTO) + core-http-kit `ClientCreatePayload`
- added to the client repository's `fields.default` allow-list (both `findMany` and `findOne`) so reads return the column
- `isValidPostLogoutRedirect` now matches against `client.post_logout_redirect_uri` — a URI matching the login `redirect_uri` but **not** the post-logout list is now rejected
- `buildWebClientAttributes` sets it to the same `<origin>/**`-per-app-origin patterns as `redirect_uri`; `WebClientProvisioner`'s MERGE widens hand-edited `web` clients on the next startup (release note)

## Migration

Per the maintainer's direction, folded into the **still-unreleased** `1783325495597-Default.ts` (both `mysql` and `postgres`) rather than a new file, so local sqlite can be lazy-patched. Verified the full **up → revert-all → up** round-trip on live Postgres and MySQL containers (exactly what the `tests-migrations` CI job runs): the column is added as `varchar(2000)`, dropped cleanly by `down()`, and re-added idempotently.

## Tests

- unit: a new separation test in `end-session/service.spec.ts` — a `post_logout_redirect_uri` matching `redirect_uri` but not the dedicated column is dropped; existing stubs updated to set the new field
- e2e: the `/logout` matrix client now sets `post_logout_redirect_uri`, so the happy-path redirect test exercises the column end-to-end (loaded from DB, matched)

Full server-core suite green (1154; the known pre-existing `grant-authorize` FK flake surfaced once on an untouched path). Not included: a client-form UI field for `post_logout_redirect_uri` — manageable via the API + auto-provisioned for the `web` client; a UI affordance is a separate nice-to-have.

## Summary by CodeRabbit

* **New Features**
  * Added support for a separate post-logout redirect allow-list, improving logout redirect handling for OIDC flows.
  * Clients can now store and manage a dedicated post-logout redirect setting alongside existing login redirect settings.

* **Bug Fixes**
  * Logout redirects are now validated against the correct allow-list, preventing unintended fallback to login redirect rules.
  * Updated validation to reject invalid post-logout redirect values more consistently.


tada5hi:
@coderabbitai review

tada5hi:
@coderabbitai review

PR #3210, opened 2026-07-09
fix: scope the id_token_hint exp-bypass so an expired token cannot poison the claims cache

tada5hi:
## Problem

An expired `id_token` verified through the RP-initiated-logout `ignoreExpiry` path poisons the shared claims cache, so `/token/introspect` reports an already-expired token as `active` for up to an hour — an RFC 7662 violation.

Root cause chain:

1. `OAuth2EndSessionService.verify()` calls `tokenVerifier.verify(hint, { ignoreExpiry: true, skipActiveCheck: true })` — `core/oauth2/end-session/service.ts:53`.
2. `ignoreExpiry` → `verifyToken(..., { validateExp: false })`, so an already-expired token passes crypto verification.
3. `OAuth2TokenVerifier.verify()` then `saveWithSignature(payload, token)` **unconditionally** — `core/oauth2/token/verifier/module.ts:125`.
4. `buildTTL(exp)` for a past `exp` computes `ttl ≤ 0` and falls back to `3_600 * 1000` (1h) — `app/modules/oauth2/repositories/token/repository.ts:143`.
5. The cache-first branch of `verify()` returns the cached payload with **no `exp` re-check** (only `jti` + optional `isInactive`) — `verifier/module.ts:38-52`. `/token/introspect` (which calls `verify` with `skipActiveCheck: true`) then derives `active` solely from `isInactive(jti)`, which the logout flow never sets.

Result: after a `/logout` carrying an expired `id_token_hint` (default `endSessionHintGracePeriod=0` honors any expired hint), that token — and any expired token submitted on an `ignoreExpiry` path — introspects as `active: true` for ~1h.

## Fix

Skip `saveWithSignature` when `options.ignoreExpiry` is set, so the exp-bypass stays scoped to the single end-session call and never leaks into the shared signature-keyed claims cache.

The only `ignoreExpiry` caller is `OAuth2EndSessionService.verify`, which uses the returned payload directly and never relied on the cache write — no behavioral loss.

## Tests

- New regression test in `verifier/module.spec.ts` — an expired token verified with `ignoreExpiry` is not written to the cache and is not retrievable by signature.
- Targeted verifier + end-session specs green (33/33); full server-core suite green.

## Notes

Found during a deep review of the plan 041/042 OAuth2/OIDC + RP-logout change set. Remaining low/info follow-ups from that review are tracked separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token verification so expiry-bypassed checks no longer re-save expired tokens into shared cache, preventing them from being treated as active later.
  * Restricted the expiry bypass to the single logout verification flow, reducing the chance of unintended reuse.

* **Documentation**
  * Clarified the logout verification guidance to highlight the cache-related security requirement.

* **Tests**
  * Added a regression test covering expired token verification with expiry bypass enabled.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1241,7 +1241,14 @@ app (kit or non-kit) ends a lingering authup session on its own logout, so
 - **id_token_hint** is verified by `OAuth2TokenVerifier` with a new
   `ignoreExpiry` option — signature, nbf and (crucially) **kind** still apply;
   only `exp` is skipped (a logout hint is routinely expired). The option threads
-  down to server-kit's `verifyToken` (`validateExp: false`). A hint whose `kind
+  down to server-kit's `verifyToken` (`validateExp: false`). The `ignoreExpiry`
+  verify path **must NOT populate the shared signature-keyed claims cache**
+  (`OAuth2TokenVerifier.verify` skips `saveWithSignature` when `ignoreExpiry` is
+  set): otherwise an expired token re-caches with `buildTTL`'s 1h fallback (a
+  past `exp` → non-positive ttl → 3600s) and the cache-first branch returns it
+  with no `exp` re-check on every later verify — so `/token/introspect` would
+  report an expired token as `active` for up to an hour (RFC 7662). The
+  exp-bypass stays scoped to the single end-session call. A hint whose `kind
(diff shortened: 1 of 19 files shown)


tada5hi:
@coderabbitai pause

PR #3225, opened 2026-07-10
fix: post-review hardening for the store stack

tada5hi:
## Summary

Fixes for the confirmed findings of the multi-agent review of `f138d4c..master` (the plan-045 store stack #3214/#3215/#3218/#3219/#3221, #3216, #3217). One scoped commit per finding.

### Store/hook zombie-commit gaps (the two high-severity races)

- **`refreshSession` had no generation guard**: a logout interleaved with the store's own refresh round-trip was silently undone — the late grant re-wrote the cleared store, cookies re-persisted, and the follow-up revalidation committed against the post-logout generation (full session resurrection, e.g. on a shared machine). The refresh now snapshots the generation before the round-trip and drops + best-effort-revokes a late grant; a cleanup that already interleaved is not run twice.
- **The hook's `REFRESH_FINISHED` stale-drop protected only the store**: the hook had already stamped the dropped session's bearer onto every attached client (mixed identity after an account switch), re-armed its timer, and the dropped pair's live server session was never revoked. `core-http-kit` now applies the response *before* emitting (so subscribers can override), and the kit's drop branch re-syncs header/timer from store state and revokes the dropped pair; the expire-date listener clears the timer when the date is unset.
- Adjacent guard gaps: `cleanup()` returns the generation it bumped to (two concurrent interactive flows could otherwise both pass the commit guard — the loser's tokens were clobbered unrevoked), and `commitSession` clears `resolutionStale` only for a non-superseded token (a mid-staging hook refresh keeps its promised awaited revalidation).

### Kit

- **`APermissionCheck` was fail-open since inception**: `computed(() => checkerFn(...))` truth-tested the returned `Ref<boolean>` object — the permission-gated slot always rendered (UI disclosure only; the server still enforced). Now a single setup-time checker call, gated on the returned ref — and the checker accepts a reactive context getter, so `name`/`input`/`options` prop changes re-trigger the evaluation (the broken `computed` never provided that either: it re-ran, but the outcome was hardwired truthy).
- **`ARoleForm`/`AScopeForm` silently dropped the `:realm-id` prop** the app pages pass (`realmManagementId`) — both now follow `AUserForm`'s locked-realm pattern (prop wins, picker hidden, `form.realm_id` written).
- **Cookie hydration violated the pinned expire-date-before-token order**: the write-back echo re-persisted the access-token cookie with no `maxAge` (session cookie). Hydration now writes the expire date first.

### server-core

- **`buildOAuth2TokenHash` TypeError'd on a legacy key row with NULL `signature_algorithm`** (nullable since the initial migration; `alg.slice(-3)` ran before the guard). A missing alg now derives SHA-256 — matching the signer's `*256` defaults for exactly those rows; an unrecognized alg still fails loud.

### Docs

- Removed a committed `<<<<<<< HEAD` merge-conflict marker from the `StoreAuthStatus` JSDoc (shipped into the emitted `.d.ts`).
- Restored the `.agents/structure.md` phase-model row #3221's stack merge reverted after #3219 added it (updated to the `unauthenticated` rename + RT-only `RESTORING` + the new refresh drop guard).
- JSDoc corrections: `RESTORING` is presence- not progress-derived; `StoreAuthOrigin` never had the 'null while unauthenticated' invariant; `RESOLVED` does not fire for a failed `resolve()`; `architecture.md`'s stale `userResolved` reference replaced with the current `userSettled` mechanism.

Deliberately **not** changed (accepted residuals): `interactionInFlight` stays a single ref (overlapping interactions can briefly flap `status`), the promise-share wrapper keeps its macrotask-deferred clear, and `RESTORING` gets a documented caveat instead of a new `settled` input.

## Test plan

- New specs: logout-during-store-refresh stays final (lifecycle), hook drop re-syncs the attached client's bearer + revokes the dropped pair (authentication-hook), `APermissionCheck` deny/allow rendering (new component spec), hydration write-back carries a derived `maxAge` (install-cookies), null-alg at_hash (helpers).
- `npm run test --workspace=packages/client-web-kit` — 17 files, 102 tests green
- `npm run test --workspace=packages/core-http-kit` — 19 tests green
- `npm run test --workspace=apps/server-core` — 121 files, 1223 tests green; `check:types` clean
- `npm run build` green for core-http-kit, client-web-kit, server-core

## Summary by CodeRabbit

- **New Features**
  - Role and scope forms can now lock realm selection to a specified realm.
  - Permission-gated content now updates reliably as permissions (and related context) change.

- **Bug Fixes**
  - Improved authentication handling during concurrent login, refresh, logout, and session restoration, including stale refresh cleanup and prevention of resurrecting discarded tokens.
  - Fixed refresh timing behavior after logout and when no access-token expire date is available.
  - Improved cookie-based session restoration ordering.
  - OAuth token hashing now supports legacy requests without an algorithm value.

- **Documentation**
  - Clarified authentication state semantics, lifecycle expectations, and account fallback behavior.


tada5hi:
@coderabbitai pause

PR #3227, opened 2026-07-11
feat: security hardening quick wins

tada5hi:
## Overview

First batch ("wave 0") of the roadmap quick wins — four independent, migration-free hardening items. One commit per item.

### 1. Cross-realm disclosure via `fields` projection (security fix)

rapiq honors a client `fields` projection over the repository default, so `GET /<entity>?fields[<entity>]=id,name` stripped `realm_id` from the SQL SELECT and the per-row `realm_scope` gate evaluated against a missing resource realm:

- an `ownOrNull`-scoped reader received **every realm's rows** (cross-realm disclosure; for clients, `fields[client]=id,secret` could disclose foreign-realm **plaintext client secrets**),
- while an `own`-scoped reader was wrongly denied its own rows.

Fix: shared `applyRealmScopeSelect` helper force-selects the gate-read columns after `applyQuery` in the user, robot, client, role-attribute and user-attribute adapters; the session adapter (fixed in #3187) is migrated onto the same helper. The regression spec pins all four exploitable directions and was verified to fail without the fix.

Audit note: role/scope/permission/policy list reads carry **no per-row realm gate at all** (their `resourceRealmMatch` usages are write paths), so there was nothing to bypass there — but their ungated list reads are a pre-existing gap worth a separate follow-up decision.

### 2. Meter `/authorize` + `/token`

The prometheus middleware `skip()` excluded exactly the security-critical endpoints. Only the `/metrics` self-scrape and root status excludes remain; both metered paths are fixed strings (no label-cardinality risk).

### 3. Configurable minimum password length

`UserValidator` accepted 3-character passwords and the reset flow enforced an inconsistent `min(5)`. New `passwordMinLength` config (env `PASSWORD_MIN_LENGTH`, default **10**) threads into user create/update, registration and password reset. Length-only, no composition rules (NIST 800-63B). New writes only — existing hashes keep verifying.

### 4. Enforce `Client.grant_types`

The column existed end-to-end but was documented "metadata only". A non-null (space/comma-delimited) list is now an opt-in allowlist enforced after client resolution in the `authorization_code` / `refresh_token` (incl. the bound-client-from-token path) / `client_credentials` / `password` grants and in the `/authorize` code-request verifier (deny before the consent UI). Violations fail with the new `unauthorized_client` error code (RFC 6749 §5.2, i18n entries in all four locales). `null` = allow-all, so upgrades are backward compatible.

## Upgrade notes (for the next release)

- New password writes reject values shorter than 10 characters unless `PASSWORD_MIN_LENGTH` is lowered; file-provisioned users with shorter passwords now fail provisioning validation at startup.
- Deployments that stored decorative `grant_types` values will see those clients restricted to the listed grants (e.g. a client listing only `authorization_code` fails refresh with `unauthorized_client`). The provisioned per-realm `web` client already lists `authorization_code refresh_token`.

## Verification

- Full monorepo test suite green (15 projects; server-core 125 files / 1254 tests, incl. the new regression + enforcement specs).
- `check:types` (tsconfig.test.json) and eslint clean.

## Summary by CodeRabbit

- **New Features**
  - Added configurable minimum password length for registration, password recovery, and user updates, with environment and configuration support.
  - OAuth2 clients can now restrict permitted grant types; unauthorized grant requests return clear localized errors.
  - Improved realm isolation so projected queries continue hiding data from other realms.

- **Bug Fixes**
  - Corrected Prometheus metrics collection to include authorization and token endpoints while excluding self-scraping noise.

- **Documentation**
  - Updated configuration and architecture guidance for password policies, OAuth2 grant restrictions, and realm isolation.


tada5hi:
@coderabbitai review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

PR #3233, opened 2026-07-12
feat: session auth-method recording with amr/acr claims + step-up

tada5hi:
Part **2/5** of the MFA family (plan 050). Stacked on #3232 — targets `feat/mfa-authenticator-devices`.

## What
Records **how** a subject authenticated and surfaces it as OIDC `amr`/`acr`, plus `acr_values` step-up.

- **auth_method:** `SessionAuthMethod` (`pwd|ldap|ext|client|robot`) stamped at all five session-creation sites; the federated IdP callback threads it through the authorization-code blob so the `/token` fallback-create inherits it.
- **Claims:** `amr`/`acr` on `OAuth2TokenPayload`, derived at the `/token` mint from the resolved session (`deriveAmrAcr`) — `pwd`/`ext`, `+otp` and `acr=urn:authup:mfa` once `session.mfa_at` is set. Emitted on every token kind (resource servers read the method without parsing an id_token); a pre-column session emits none.
- **acr_values:** mounted on `/authorize` (persisted on the code blob), advertised via discovery `acr_values_supported`. Voluntary per OIDC Core §5.5.1.1 — unknown tokens ignored; `urn:authup:mfa` acts as a **step-up trigger** enforced only while the user holds a confirmed factor, requiring `session.mfa_at` within `mfaFreshnessMaxAge`.

## Note on the default
`mfaFreshnessMaxAge` defaults to **60s**, deliberately not the plan's 0 — a 0-window could never absorb the hosted challenge round-trip and would loop the ladder. Documented + pinned by tests.

## Tests
`deriveAmrAcr` table + step-up matrix, per-grant `auth_method` stamping, discovery `acr_values_supported`, and an end-to-end id_token/access-token claim-decode spec (pwd → mfa after a second factor).

## Summary by CodeRabbit

* **New Features**
  * Added OpenID Connect `amr` and `acr` claims to access and ID tokens.
  * Added support for requesting password or MFA authentication contexts with `acr_values`.
  * Added MFA step-up enforcement when the existing second-factor verification is no longer fresh.
  * OpenID Connect discovery now advertises supported authentication contexts.

* **Configuration**
  * Added configurable MFA verification freshness, defaulting to 60 seconds.

* **Documentation**
  * Documented authentication claims, step-up behavior, and the new configuration setting.


tada5hi:
@dependabot full review

tada5hi:
@coderabbitai review

PR #3243, opened 2026-07-13
fix: dedupe realm-scope gate selects - postgres list queries with include failed

tada5hi:
## Problem

Reported against a production deployment (`authup/authup` + `postgres:18`): `GET /users?include=realm` — and every other realm-gated list combining `include=` with an overlapping `fields` projection, e.g. `/robots?include=realm` or `/users?include=realm&fields=id,name` — returns 500 with `column reference "user_id" is ambiguous`.

## Root cause

`applyRealmScopeSelect` (plan 039) re-added `realm_id` / extra gate columns even when the rapiq `fields` projection had already selected them. TypeORM's `addSelect` is **not** a no-op for an already-selected column — it emits a second identically-aliased select — and under a join + take (TypeORM's DISTINCT id-subquery wrapper) postgres rejects the wrapper's `ORDER BY "user_id"` as ambiguous (mysql: duplicate column name in the derived table). Not reachable in CI before: no spec exercised `include=` on a `getMany`.

## Changes

- `applyRealmScopeSelect` now dedupes against the already-applied projection (`qb.expressionMap.selects`); its doc comment previously (wrongly) claimed `addSelect` deduped.
- Regression tests: `include=realm` collection cases in `user.spec.ts` / `robot.spec.ts`, including a `fields` + `include` combination.
- `.agents/architecture.md` plan-039 paragraph updated accordingly.

## Verification

- Reproduced the 500s by booting the built server against a scratch postgres DB created by the migration CLI; fixed endpoints return 200 with correct joined data.
- Full server-core suite passes on sqlite; the affected specs also pass against postgres and mysql; `check:types` clean.

## Out of scope

The same report also surfaced `GET /robot-roles?include=robot` failing on postgres (`operator does not exist: uuid = character varying`): migration-created schemas carry `auth_robot_roles.robot_id` as `character varying` without FK (plus the legacy orphaned `client_id` join column) from the initial migration, while the entity joins robot via `robot_id`. A schema-alignment migration was deliberately dropped from this PR — the robot entity is slated for removal, at which point the table goes away entirely.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1662,9 +1662,15 @@ secret_encrypted` — the plaintext-secret gate precondition), `role-attribute`,
 no `fields`, so the call is pinning defense in depth there). `role` / `scope` /
 `permission` / `policy` list paths carry no per-row realm gate (their
 `resourceRealmMatch` usages are write paths on server-loaded data), so they are
-deliberately not force-selected. Regression specs:
+deliberately not force-selected. The helper **dedupes against the already-applied
+projection** (`qb.expressionMap.selects`) — TypeORM's `addSelect` is NOT a no-op
+for an already-selected column: it emits a second identically-aliased select, and
+under a join + take (TypeORM's DISTINCT id-subquery wrapper) postgres rejects the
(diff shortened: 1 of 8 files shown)


tada5hi:
@coderabbitai pause

PR #3262, opened 2026-07-15
fix(server-core): prevent cross-realm audit event disclosure

### LINKED ISSUE #3250: realm_admin can read other realms' entity-CRUD audit events (null-realm attribution + ownOrNull)
## Summary

A `realm_admin` (whose `EVENT_READ` reach is `ownOrNull`) can read entity-CRUD audit events belonging to other realms, because rows whose realm cannot be read from a top-level `realm_id` column are stamped realm-less.

## Root cause

`EntityEventHandler` (`apps/server-core/src/core/entities/event/entity-event-handler.ts`) attributes realm with:

```
const realmId = ctx.content.data?.realm_id ?? null;
```

Entities without a bare `realm_id` column are therefore recorded as `realm_id = null`:
- **Junction rows** — role-permission, user-role, client-role, robot-role, etc.
- **identity_provider_accounts** — keys realm under `user_realm_id` / `provider_realm_id`, not `realm_id`.

The per-row gate in `EventService.getMany` uses `resourceRealmMatch`; `ownOrNull` matches resource realm `=== actor realm` **or** `=== null`. So every null-realm audit row passes for a `realm_admin` in *any* realm.

## Impact

A realm-A `realm_admin` can enumerate junction membership-change events (ref_type/ref_id/actor/IP/time) for realm-B subjects, and read PII carried in diffs (`user.email`, `first_name`/`last_name`; external-IdP `provider_user_email` / `provider_user_id`). No realm-bound entity row leaks, but the null-realm attribution combined with `ownOrNull` is a real cross-realm read.

## Fix direction

Either derive junction/EA-account event `realm_id` from the owner side (e.g. `role_realm_id`, `user_realm_id`) rather than the missing top-level column, or exclude null-realm rows for scoped (`ownOrNull` / `own`) readers.

## Related (event read API, same PR)

- `EventService.getMany`'s drop-unauthorized-rows shape decrements `total` only for rows denied on the current page, leaking an approximate cross-realm event count via the pagination total. (Pre-existing pattern shared with SessionService/RobotService, repeated here.)
- The `/realms/:realmId/events` dual-mount ignores the route `:realmId` — the service never reads `getRequestRealmID(event)`, so the segment is cosmetic. Not an authz hole (the per-row reach gate still isolates) but a filter/scoping gap worth fixing for correctness.

Introduced in #3229 (commit 5a30950a4). Unreleased.

tada5hi:
## Summary

- attribute entity audit events to their canonical owner realm, including junction records and identity-provider accounts
- enforce `/realms/:realmId/events` as a mandatory read constraint through explicit `{ realmId }` service options
- expose only the visible page count after per-row authorization

## Testing

- `npx eslint --fix` on all changed TypeScript files
- `npm run test --workspace=apps/server-core -- test/unit/core/entities/event/entity-event-handler.spec.ts test/unit/core/entities/event/service.spec.ts test/unit/http/controllers/entities/event-entity-bridge.spec.ts`
- `npm run build -w apps/server-core`

Closes #3250

## Summary by CodeRabbit

- **Bug Fixes**
  - Event logs are now correctly restricted to the selected realm.
  - Individual events from other realms are hidden from unauthorized readers.
  - Event records now use the owning resource’s realm for attribution, including junction records.
  - Partial updates retain the correct previous realm attribution.
  - Event totals now reflect only visible, authorized records.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -2519,9 +2519,10 @@ hub lacks: a **closed taxonomy** (`EventName`/`EventScope` enums in
   repository force-selects the gate columns (`applyRealmScopeSelect`, plan-039
   discipline). Nested `/realms/:realmId/events` reads add a mandatory
   repository realm predicate (and single-record reads fail as not found on a
-  mismatch). Because collections still apply asynchronous per-row policies
-  after the database query, their response total is the visible page count
-  rather than the unauthorized backing-table total. `EVENT_READ`
+  mismatch). Flat collection reads derive the actor's coarse realm reach from
+  `EVENT_READ` and apply it (plus an own-row alternative) before database
+  pagination, keeping cross-realm rows out of both pages and totals. Residual
+  per-row policy denials decrement the repository total. `EVENT_READ`
   auto-provisions via `Object.values(PermissionName)`:
   `admin` = `any`, `realm_admin` = `ownOrNull` (deliberately NOT in the OWN
   override list). Typed client: `client.event.getMany/getOne`.
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -872,7 +872,7 @@ Realm-scoped controllers are dual-mounted via `@routup/decorators` array paths:
 export class UserController { ... }
 ```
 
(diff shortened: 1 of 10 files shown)


tada5hi:
@coderabbitai full review

PR #3261, opened 2026-07-15
feat!: add OAuth mutual TLS authentication

tada5hi:
## Summary

- replace the client `is_confidential` flag with explicit `auth_method` and `token_binding_method` fields
- authenticate OAuth clients through trusted RFC 9440 or forwarded certificate evidence with offline PKI validation
- parse, thumbprint, and verify certificate paths with `@peculiar/x509` plus an injectable WebCrypto provider
- issue and enforce RFC 8705 certificate-bound access and refresh tokens in Authup and the reusable server adapters
- add discovery metadata, admin UI controls, migration backfills, and NGINX/Traefik deployment documentation

## Breaking change

`Client.is_confidential` is replaced by `auth_method` (`none`, `secret`, or `tls`) and `token_binding_method` (`none` or `tls`). The MySQL and PostgreSQL migrations preserve existing behavior by mapping `false` to `none` and `true` to `secret`.

## Verification

- server-core: 1,677 tests passing
- client-web-kit: 141 tests passing
- affected core, HTTP, specs, and server-adapter workspace tests passing
- server and documentation builds, TypeScript checks, and lint passing
- MySQL and PostgreSQL migration up/down/reapply and legacy-row backfill verified

## Summary by CodeRabbit

* **New Features**
  * Added OAuth client authentication modes (`none`, `secret`, `tls`) and token binding modes (`none`, `tls`).
  * Added certificate-bound access/refresh tokens with proof-of-possession (`cnf`) validation.
  * Added RFC 8705 mTLS endpoint support and OpenID discovery metadata for certificate-bound access tokens.
  * Added configurable certificate ingestion/validation via `CERTIFICATE_SOURCE` and `MTLS_PUBLIC_URL`.
* **Documentation**
  * Added/expanded guides for client certificates, trust anchors, NGINX, and Traefik mTLS setups.
* **Bug Fixes**
  * Improved fail-closed validation for token-bound requests (including thumbprint checks) and updated client authentication behavior.
* **Chores**
  * Updated UI/client provisioning and tests to use the new auth/binding fields instead of the prior confidential flag.

PR #3267, opened 2026-07-16
fix(server-core): harden mTLS/key-store/oauth2 review findings + core error taxonomy

tada5hi:
Addresses findings from a security review of the recently-merged OAuth mTLS (#3261), realm key store (#3255), key management (#3256/#3257/#3260), access-policy + persisted consent (#3246/#3254), and MFA pending-ticket (#3244) work — plus a `core/**` error-taxonomy change.

## Security hardening

- **Certificate path (DoS + RFC 5280):** bound the certification-path search with a signature-verification budget — an attacker could otherwise present 10 mutually-issuing certs (shared key, identical DNs) and force ~factorial signature verifications per unauthenticated `POST /token`. Also reject unhandled **critical** extensions (which fails name-constrained CAs *closed* rather than trusting a chain whose name constraints we don't evaluate — closes a cross-tenant client-impersonation vector) and **weak** (SHA-1/MD*) edge signatures; continue to the next trust anchor on a path-assertion failure; add `IClientCertificateValidator`; let infra faults surface as 5xx instead of masquerading as a bad certificate.
- **`/token/revoke` (RFC 7009):** always write the inactive marker from the verified payload — previously a cache flush/eviction made an access-token revoke a silent no-op (202 but still active until natural expiry).
- **Key store:** widen `auth_keys.{decryption,encryption}_key` to `text` (a KEK-wrapped RSA-4096 import overflows `varchar(4096)` → 500 on strict SQL / silent truncation on non-strict MySQL); log the lazy wrap-on-read write-back failure; reject an import masquerading as `wrapped.v1.` material; `isUUID`-guard `resolveById` (a garbage JWT `kid` was a postgres 500, now a clean miss); belt-and-braces null of private material on key reads.
- **Consent:** honour the `/realms/:realmId/consents` route realm as a mandatory predicate; `getOne`/`delete` fail not-found on realm mismatch (same class the audit-event fix #3262 applied to `/events`).
- **MFA login ticket:** claim the ticket before minting so two concurrent completions can't both issue a token pair.
- **Refresh grant:** explicit token-kind guard (defense in depth alongside the existing session-row check).
- **Provisioning file source:** fail closed on a non-object root (top-level list/scalar/empty file); dispatch module-vs-data unwrap by file extension instead of a `default`-key sniff.
- **Config:** fail loud on a set-but-invalid boolean for security toggles (`MFA_ENABLED` / `MFA_REQUIRED` / `LOGIN_ATTEMPT_THROTTLE_ENABLED` / `EVENT_LOG_ENABLED` / `EVENT_LOG_ENTITY_ENABLED`) — envix's `readBool` silently swallowed e.g. `MFA_REQUIRED=yes`.
- Misc: `core-http-kit` nullifies empty enroll payload fields; bump the composite install action to `actions/setup-node@v7`; refresh stale `MFA_ENCRYPTION_KEY` doc references.

## Core error taxonomy

- Add `ValidationError` (+ `isValidationError`) to `@authup/errors` — a transport-agnostic domain error that maps to HTTP 400 via `httpStatusFromCode`.
- Replace `BadRequestError` / `UnauthorizedError` throughout `apps/server-core` `core/**` with `ValidationError` / semantic domain errors; HTTP-status-named error classes now belong to the HTTP layer only. `conventions.md` updated.

## Not included (deliberate)

- **Event/consent count metadata** and **refresh binding status (invalid_client vs invalid_grant):** the current behaviour is deliberately tested; left as-is (documented limitations).
- **`TokenVerifier.verify()` cnf enforcement:** would change a released public signature; the request/socket wrappers already enforce binding for all documented usage.
- **Key/trust-anchor lifecycle audit events:** already a documented follow-up; deferred (needs an event-taxonomy decision).

## Verification

`npm run check:types`, `npm run test` (server-core: 1700 tests), the changed adapter/errors packages, `npm run lint`, and `npm run build` all pass. Migration ALTERs are covered by the `tests-migrations` job.

## Summary by CodeRabbit

* **Bug Fixes**
  * Consent records are now consistently restricted to the active realm.
  * Revoked access tokens become inactive immediately, including after cache refreshes.
  * MFA login tickets can no longer be reused.
  * Refresh-token flows reject tokens of an unexpected type.

* **Security**
  * Client certificate validation now rejects weak signatures and unsupported critical extensions.
  * Private key material is hidden from read responses and supports larger encrypted values.
  * Invalid identifiers no longer trigger unnecessary database lookups.

* **Improvements**
  * Configuration now reports unrecognized boolean environment values.
  * Provisioning rejects malformed or unsupported file contents.
  * Validation errors are reported more consistently across the API.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

PR #3270, opened 2026-07-16
fix(server-adapter)!: enforce certificate binding inside TokenVerifier.verify()

### LINKED ISSUE #3268: server-adapter-kit: enforce RFC 8705 certificate binding inside TokenVerifier.verify()
Found during the review + hardening of PR #3267.

## Problem

Certificate binding (RFC 8705 `cnf`/`x5t#S256`) is enforced only in the `verifyRequest` / `verifySocket` wrappers (`@authup/server-adapter-node` / `-web` / `-socket-io`), not in the primitive `TokenVerifier.verify()` itself. The raw `verify()` fails **open** on a bound token: a resource server that calls `verify()` directly (bypassing the documented wrappers) accepts a bound, stolen token with no certificate — silently losing the proof-of-possession the binding exists to provide.

## Suggested fix

- Add an optional thumbprint input to `verify(token, options?)` (`options.certificateThumbprint`: a value or a provider) and enforce binding inside `verify()` via `assertTokenCertificateBinding`.
- Update the three request/socket wrappers to forward the presented thumbprint into `verify()` and drop their own post-verify `assertTokenCertificateBinding` call.
- Move the wrapper unit tests' binding-enforcement assertions into `server-adapter-kit`'s `verifier.spec.ts` (testing the real `TokenVerifier`); the wrapper tests then assert only that the wrapper forwards the thumbprint provider.

## Caveat (deliberate break)

Making `verify(token)` (no thumbprint) throw on a bound token is a behaviour change for any direct caller that currently gets the token back — but that is exactly the vulnerable behaviour, so fail-closed is the intended semantics. It changes a released public signature, which is why it was pulled out of #3267 into its own focused change.

tada5hi:
## Problem

RFC 8705 `cnf`/`x5t#S256` enforcement lived only in the `verifyRequest` / `verifySocket` wrappers (`@authup/server-adapter-node` / `-web` / `-socket-io`). The primitive `TokenVerifier.verify()` failed **open** on a bound token: a resource server calling `verify()` directly accepted a bound, stolen token with no certificate — silently losing the proof-of-possession the binding exists to provide.

## Changes

- `@authup/server-adapter-kit`: `ITokenVerifier.verify(token, options?)` gains `options.certificateThumbprint` — the presented certificate's SHA-256 DER thumbprint as a value or a lazy provider (invoked only when the payload carries `cnf`). `verify()`, `verifyLocal()`, and `verifyRemote()` all assert the binding via `assertTokenCertificateBinding` on every verification — **cached results included**, since the binding is per-request evidence, not a property of the token alone. The former public bodies moved to protected `resolveLocal`/`resolveRemote`.
- The three wrappers forward a lazy per-request/per-socket thumbprint provider into `verify()` and drop their own post-verify assert. Wrapper-level behavior is unchanged.
- Binding-enforcement tests moved onto the real `TokenVerifier` in `server-adapter-kit`'s `verifier.spec.ts` (no-thumbprint, mismatch, value match, async provider, provider laziness for unbound tokens, cache-hit enforcement, remote introspection); the wrapper specs now assert only that the lazy provider is forwarded.
- Docs: the three middleware pages and `.agents/architecture.md` note that enforcement lives in the primitive.

## Breaking change (deliberate)

`TokenVerifier.verify(token)` without a thumbprint now fails closed (`JWTError`) on a certificate-bound token instead of returning it. That is exactly the vulnerable behavior this fixes; direct callers must pass `verify(token, { certificateThumbprint })`. Unbound tokens are unaffected.

closes #3268

## Summary by CodeRabbit

* **New Features**
  * Added certificate-bound token verification using certificate thumbprints.
  * Token verification now supports direct or lazy, asynchronous thumbprint providers.
  * Added enforcement across HTTP request and socket authentication flows.
  * Bound tokens fail closed when the required thumbprint is missing or incorrect.

* **Documentation**
  * Clarified certificate-binding behavior and direct verification requirements across middleware guides.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1732,7 +1732,15 @@ realm trust is only an authentication concern. Refresh must present the same
 thumbprint, and rotation preserves the claim. Authup's authorization
 middleware and `server-adapter-*` consumers fail closed on a bound token unless
 the resource request supplies that thumbprint; resource servers compare only
-the signed thumbprint and do not repeat chain validation.
+the signed thumbprint and do not repeat chain validation. On the adapter side
+the enforcement lives inside `@authup/server-adapter-kit`'s
+`TokenVerifier.verify(token, options?)` itself (issue #3268):
+`options.certificateThumbprint` takes the presented thumbprint as a value or a
+lazy provider (invoked only when the payload carries `cnf`), and the check runs
+on every verify — cached results included, since the binding is per-request
+evidence. The `verifyRequest`/`verifySocket` wrappers merely forward a lazy
+per-request provider; a direct `verify(token)` call on a bound token without a
+thumbprint fails closed (`JWTError`), never open.
 
 ### PKCE for public clients
 
--- a/docs/src/sdks/javascript/server-adapter/node-middleware.md
+++ b/docs/src/sdks/javascript/server-adapter/node-middleware.md
@@ -43,6 +43,11 @@ Behind a proxy, derive the value from the proxy's authenticated certificate
 contract instead. Never hash an unchecked public header; the proxy must remove
 and overwrite it, and the backend listener must be private.
 
+Enforcement lives inside `TokenVerifier.verify()` itself — the middleware only
+forwards a lazy thumbprint provider. Calling `verify(token)` directly on a
+bound token therefore also fails closed; pass the presented thumbprint via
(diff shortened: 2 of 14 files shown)


tada5hi:
@coderabbitai pause

PR #3271, opened 2026-07-16
feat(server-core): emit audit events for key & trust-anchor lifecycle operations

### LINKED ISSUE #3269: server-core: emit audit events for key & trust-anchor lifecycle operations
Found during the review + hardening of PR #3267; the architecture doc already notes this as a follow-up.

## Problem

Keys and trust anchors deliberately have **no** entity subscriber (to keep private key material off the realtime/audit bus), and `KeyService` / `TrustAnchorService` make no explicit `EventService.record()` calls. As a result, security-sensitive operations are invisible in `auth_events` while routine entity edits (e.g. a role change) are logged:

- creating an **enabled** CA trust anchor (turns on mTLS client auth for a realm),
- disabling / deleting a signing key (mass token invalidation),
- force crypto-shredding an encryption key (`?force=true`).

## Suggested fix

Add explicit, **metadata-only** `EventService.record()` emits (never key material) on create / update (incl. status change) / delete in `KeyService` and `TrustAnchorService`:

- inject `eventService?` into `KeyServiceContext` / `TrustAnchorServiceContext` and wire it from the controller factory (`container.resolve(DatabaseInjectionKey.EventService)`),
- derive actor attribution from the `ActorContext`, realm from the entity,
- pick the event name/scope (generic `created`/`updated`/`deleted` under `EventScope.ENTITY` with a `key` / `trust_anchor` ref type, or a dedicated `KEY_*` set),
- payload limited to id / name / use / status — **never** `decryption_key` / `certificate` bytes.

tada5hi:
## Summary

Closes #3269.

Keys and trust anchors are deliberately subscriber-less (private material must stay off the realtime/audit bus), which left their security-sensitive mutations — enabling a CA trust anchor (turns on mTLS client auth for a realm), disabling/deleting a signing key, force crypto-shredding an encryption key — invisible in `auth_events` while routine entity edits are logged.

`KeyService` / `TrustAnchorService` now record explicit, **metadata-only** `EventService.record()` emits on create / update / delete:

- Generic `created`/`updated`/`deleted` under `EventScope.ENTITY` with `ref_type: key` / `trustAnchor` (the entity-event handler's realm-key map already carried both entries).
- **Actor** attribution from the `ActorContext`, **realm** from the entity, **request** attribution (path / method / IP / user-agent) via the same injected `useRequestEventContext` AsyncLocalStorage getter the entity-CRUD bridge uses; both wired from the controller factories (`DatabaseInjectionKey.EventService`).
- **Payload**: keys carry `name`/`use`/`status` + a scalar `diff` (name/priority/status) on update + `force: true` on a forced crypto-shred; trust anchors carry `name`/`enabled` + diff. Never `decryption_key` / `encryption_key` / `certificate` bytes — enforced centrally by the sanitizer's allowlist-first write boundary.
- Emits ride the **default (long) event retention** and are not gated by `eventLogEntityEnabled` — these are security events, not entity churn.

### Drive-by fix

The MFA lifecycle events (`mfaEnrolled`/`mfaRemoved`/…) pass `data: { kind }`, but `kind` was never in the sanitizer allowlist, so their data was silently stripped to `null` at the write boundary. `kind` is now allowlisted, with spec coverage.

## Tests

- Service-level `audit events` blocks in the key + trust-anchor specs: attribution, exact payloads, status/enabled diffs, force flag, no-emit-on-failure, no-material assertions.
- One end-to-end HTTP test per entity pinning the factory wiring (create → update → delete yields three attributed rows via `client.event.getMany`).
- Sanitizer spec extended for the new allowlist keys.
- Full server-core suite green (1725 passed; the 2 failing LDAP spec files fail identically on an untouched checkout — local LDAP container state, unrelated).

## Docs

`.agents/architecture.md`: key-store + Stage C sections no longer defer the emits as a follow-up; the Security Event Log emit-sites list now includes the key / trust-anchor lifecycle.

## Summary by CodeRabbit

* **New Features**
  * Added security event logging for key and trust-anchor creation, updates, and deletion.
  * Events include actor and request details plus relevant metadata and change differences.
  * Sensitive key and certificate material is excluded from event records.
  * Forced key deletion is clearly marked in the audit event.

* **Documentation**
  * Updated architecture documentation to describe lifecycle event logging and security event coverage.

* **Tests**
  * Added coverage for lifecycle events, metadata sanitization, attribution, diffs, and failure handling.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -2306,8 +2306,15 @@ every read and the create response null `decryption_key` (authentik
 CVE-2024-42490 is the cautionary tale). Typed client: `client.key.*`
 (`KeyCreatePayload`/`KeyUpdatePayload`; `delete(id, { force })`). Keys have
 **no entity subscriber** (deliberate — `afterInsert` content would carry raw
-private material onto the realtime bus), so no entity-CRUD audit rows;
-explicit `EventService` emits are a follow-up.
+private material onto the realtime bus); the audit trail comes from
+explicit, metadata-only `EventService.record()` emits inside `KeyService`
+(issue #3269): ENTITY-scope `created`/`updated`/`deleted` rows with
+`ref_type: key`, actor from the `ActorContext`, request attribution via the
+injected `useRequestEventContext` getter, `data` limited to
+`name`/`use`/`status` plus a scalar `diff` on update and `force: true` on a
+forced crypto-shred — never `decryption_key`/`encryption_key`/`certificate`.
+The emits ride the default (long) event retention, not the short
+entity-churn TTL, and are not gated by `eventLogEntityEnabled`.
 
 **Imported certificates (plan 071 Stage B):** create may attach an immutable
 PEM certificate chain only to an **imported signature key** (never generated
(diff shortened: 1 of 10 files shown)


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

Issue #3279, opened 2026-07-19
tracking: rapiq v2 follow-up migrations, gated on upstream rapiq releases


tada5hi:
## Context

The rapiq v2 migration (#3276) surfaced six upstream enhancement requests (tada5hi/rapiq#774–#779) plus one major upstream feature in flight (tada5hi/rapiq#780, entity-derived schemas). Each of them unblocks a concrete simplification or capability in authup. This issue tracks those follow-ups; each item activates once the referenced rapiq change ships in a released `@rapiq/*` version.

## Phase 1 — mechanical cleanups (independent, one small PR each)

> ✅ **Phase 1 complete** — delivered grouped: deps bumps #3281 (beta.2, merged) + #3282 (cleanups + beta.3 bump, merged). rapiq `2.0.0-beta.3` is the floor: the adoption surfaced two upstream regressions, both fixed there — tada5hi/rapiq#789 (index-signature relation keys) and tada5hi/rapiq#790 (DEPTH not threaded into the string-key arms → `TS2590` in client-web's typecheck on the beta.2 baseline).

- [x] **Guards** (gated on tada5hi/rapiq#774): replace the hand-rolled duck-type guards with core exports — `isQuery` in `packages/core-http-kit/src/helpers/query.ts`, `isSortsNode` in `packages/client-web-kit/src/core/query/sort.ts`. *(→ #3282; `isSortsNode` had no caller and was deleted, `isCondition` stays as a composition of the native `isFilter`/`isFilters` — rapiq ships no union guard.)*
- [x] **Case-sensitive policy evaluation** (gated on tada5hi/rapiq#775): drop `collectConditionFields` in `packages/access/src/policy/built-in/attributes/evaluator.ts` in favor of `compileFilters(condition, { caseSensitive: true })`. Existing case-exactness guard tests must stay green unchanged. *(→ #3282; guard tests green unchanged.)*
- [x] **Schema key types** (gated on tada5hi/rapiq#777): remove the two remaining `@ts-expect-error` markers (client → `accessPolicy`, policy → `realm` nullable relations) and the `'data' as never` cast in the event schema. *(→ #3282; the two interim markers the #777 fix initially forced onto `user-permission`/`user-role` — tada5hi/rapiq#789 — are gone again on beta.3.)*
[...]

PR #3282, opened 2026-07-20
refactor: adopt the rapiq v2.0.0-beta.2 native surfaces

tada5hi:
The four mechanical cleanups gated on the rapiq v2 beta bumps (#3281) — phase 1 of #3279, one commit each — plus the `2.0.0-beta.3` bump that unblocks them:

1. **Guards + build-input type** (`client-web-kit`, `core-http-kit`): the duck-typed `isQuery` guards are replaced by `@rapiq/core`'s `isQuery` (tada5hi/rapiq#774); the unused `isSortsNode` is dropped; `isCondition` stays as a composition of the native `isFilter`/`isFilters` (rapiq ships no union guard); the hand-rolled `QueryInput` mapped type collapses onto `QueryBuildInput<T, 3>` (tada5hi/rapiq#776 — closes the depth-bounded-aliases item too).
2. **Case-sensitive attributes evaluator** (`access`): `collectConditionFields` is gone — `compileFilters(condition, { caseSensitive: true })` (tada5hi/rapiq#775) pins byte-exact string equality globally. The case-exactness guard tests stay green unchanged.
3. **Schema key-type workarounds** (`server-core`): the client/policy nullable-relation `@ts-expect-error` markers and the event schema's `'data' as never` cast are removed (tada5hi/rapiq#777).
4. **Native parameter mask** (`server-core`): the local `maskQuery` re-slice is deleted; `ParseQueryOptions.parameters` (tada5hi/rapiq#778) is forwarded to `queryCodec.decode`. Unlisted parameters are neither parsed nor defaulted, so the session bulk-revoke (`parameters: ['filters']`) keeps its no-truncation guarantee (schema `pagination.maxLimit` never materializes). `ApplyRequestQueryParameter` now aligns with rapiq's `Parameter` vocabulary (`'sort'`; the local `'sorts'` alias had no caller).
5. **`fix(deps)`: bump the six `@rapiq/*` pins to `^2.0.0-beta.3`**, resolving the two issues this work surfaced:
   - tada5hi/rapiq#790 (DEPTH now threaded into the string-key arms) — `apps/client-web`'s `nuxi typecheck` is **green again** (the beta.2 baseline tripped `TS2590` on the self-recursive `Client`/`Policy` graphs; the kit's declared depth 3 now actually binds).
   - tada5hi/rapiq#789 (relation keys of index-signature targets restored) — the two interim `@ts-expect-error` markers on the `user-permission`/`user-role` relation lists are removed again.

## Verification (on beta.3)

- full monorepo build green (22 projects, including `client-web`'s `nuxi typecheck`)
- test suites green: `packages/access` (91), `packages/core-http-kit` (18), `packages/client-web-kit` (150), `apps/server-core` (full suite — only the pre-existing environmental LDAP-container spec files fail, tests skipped)
- `check:types` green on `apps/server-core`

Closes the phase-1 checklist of #3279.


tada5hi:
@coderabbitai pause

PR #3285, opened 2026-07-20
feat(server-core): validate entity schemas against typeorm metadata at boot

tada5hi:
## Context

Outcome of the #3279 phase-2 evaluation (entity-derived schemas, gated on tada5hi/rapiq#780 — the APIs shipped in rapiq `2.0.0-beta.3`). A full prototype was built in a worktree: schema options moved beside the repository adapters, `core/query`'s registry left empty and composed at `DatabaseModule.setup` via `defineSchemaWithEntity`, services decoding by `EntityType` name.

**The prototype settled the question against derivation, per the issue's own keep-the-status-quo criterion:**

- Under `EntityType` naming, every derivable contribution must be overridden anyway: the derived name is `camelCase(className)` (`userEntity` ≠ `user`), the derived `schemaMapping` targets those class names, and the derived `relations.allowed` defaults to *all* relations (authup restricts them deliberately). Net contribution of the metadata: **validation only**.
- `defineSchemaRegistryWithDataSource` derives a schema for **every** DataSource entity (session tokens, EA tables, ...) with no allow-lists — a fail-open query surface for entities that must not have one.
- A boot-populated registry turns static imports into a lifecycle: decode-by-name throws before registration, so DataSource-free consumers (service unit specs) need per-worker seeding via a vitest setup file — whose import chain immediately broke `vi.mock('@authup/server-kit')` in the signer/verifier crypto specs (module registry beats the hoisted mock). Fixable, but a standing constraint every future schema import has to respect.

So the schemas stay exactly where #3283 put them (static, colocated in core, the read-side analog of the validators), and this PR extracts the one genuine win:

## Changes

- `validateEntitySchemas(dataSource)` (`app/modules/database/repositories/schema-validation.ts`), run by `DatabaseModule.setup` after `dataSource.initialize()`: every registered schema's allow-listed keys (fields default/allowed, filters, sort, relations) must exist on its entity's TypeORM metadata — plain keys as column property paths (or relation names for `relations.allowed`), dotted keys headed by a relation. Schema/entity drift (e.g. a renamed column) fails the boot fail-fast instead of surviving as a dead filter or a runtime adapter error.
- The schema-name → entity-class map is explicit; a mapped name missing from the registry fails via `getOrFail`, so a renamed/removed schema surfaces too.
- Spec: `test/unit/app/modules/database/schema-validation.spec.ts` over a metadata-only in-memory DataSource — the "accepts every registered entity schema" case doubles as proof there is no existing drift; negative cases cover unknown keys, dotted keys, and relation-vs-field misuse.
- `.agents/architecture.md` Query IR flow section documents the validation and records the phase-2 conclusion.

The upstreaming (tada5hi/rapiq#800) shipped same-day in rapiq `2.0.0-beta.4` and is **adopted here**: the `@rapiq/*` pins are bumped and the local assert is replaced by `@rapiq/typeorm`'s `assertSchemaMatchesEntity` — a superset covering allow-lists, fields/sort defaults and the filters default condition tree. `schema-validation.ts` keeps only the schema-name → entity-class map and the `validateEntitySchemas` loop; other consumers (PrivateAIM/hub#1758) get the helper for free.

## Verification

Full server-core suite: 155/157 spec files pass (every HTTP suite boots through the validation); the 2 failures are the LDAP specs failing identically on a clean checkout (stale local LDAP container). Lint + both tsconfig typechecks clean.

Refs #3279

## Summary by CodeRabbit

* **New Features**
  * Added startup validation to detect mismatches between query schemas and database entities.
  * The application now fails early when configured fields, filters, sorting options, or relations reference unavailable entity properties.

* **Tests**
  * Added coverage for valid schemas, invalid properties, relation paths, and relation allow-list rules.

* **Documentation**
  * Documented schema drift validation and its role in keeping query allow-lists consistent.


tada5hi:
@coderabbitai pause

Issue #3286, opened 2026-07-20
RFC: data-availability-driven policy evaluation (tri-state engine, derived preEvaluate)


tada5hi:
## Problem

`preEvaluate` is currently defined as `evaluate` with a hardcoded exclusion list — `policiesExcluded: [ATTRIBUTES, ATTRIBUTE_NAMES, REALM_MATCH]` (`packages/access/src/permission/evaluator/module.ts`) — and the engine turns an excluded policy into a neutral pass `{ success: true }` (`packages/access/src/policy/engine/module.ts:82-100`). Two structural problems:

1. **The phase split is a hand-maintained type list.** Whether a policy can be evaluated *before data exists* is a property of the policy (what data it needs), not of a list in the caller. Every new policy type must be remembered in that list — or the pre-gate silently evaluates it against missing data. The question we actually want to answer is: **when can which policy be evaluated?**

2. **"Assume true" is a lossy encoding of "unknown" and breaks under negation.** Masking a subtree to `true` is over-approximating (a sound early-deny gate) only while there is no inversion above it. With `composite { invert: true, children: [attributes(...), identity(...)] }`, `preEvaluate` masks `attributes` → true and evaluates `NOT(identity)`: an actor satisfying the identity check is **denied at the pre-gate**, even though the full policy `NOT(attributes AND identity)` allows them whenever the row's attributes fail. Mask-then-negate ≠ negate-then-mask: under an odd number of inversions the gate flips to under-approximating and produces spurious denials. Fail-closed, but wrong — and currently untested.

## Proposal: data-availability-driven evaluation (tri-state)

> A policy leaf is evaluable at instant T iff its declared data requirements are satisfied by what is known at T. Until then it is not "true" — it is **pending**.

### Evaluator contract

```ts
interface IPolicyEvaluator {
    /** PolicyData keys this policy needs to settle (e.g. identity, attributes, clock). */
    requires: PolicyDataKey[];

    evaluate(policy, ctx): Promise<PolicyEvaluationResult>;

    /** Optional: express the policy as a condition over row attributes (rapiq ICondition)
     *  instead of waiting for the row — enables WHERE pushdown. */
    lower?(policy, knowns): ICondition;
}
```

### Engine result

[...]

PR #3290, opened 2026-07-21
feat(access): derive preEvaluate from data availability (tri-state policy engine)

tada5hi:
## What

Phase 1 of #3286 — the policy engine becomes **tri-state** (`settled(true) | settled(false) | pending`), and `preEvaluate` is derived from data availability instead of a hand-maintained type exclusion list. Public API shapes are unchanged.

### `@authup/access`

- `PolicyEvaluationResult` gains `pending?: boolean`. Pending rides on `success: false` (fail-closed: any consumer that ignores the flag behaves like a deny), so full `evaluate()` needs no change — the historical missing-data deny is preserved with a pending-flavored `DATA_MISSING` issue.
- `IPolicyEvaluator` gains optional `requires?(value)`: the PolicyData keys the policy needs before it can settle. The engine checks it after the include/exclude neutral-pass, before invoking the evaluator; any missing key → pending. Like the neutral-pass, `invert` is never applied to a pending result — unknown stays unknown under negation.
- Built-in `requires`: identity → `IDENTITY`; attributes / attributeNames → `ATTRIBUTES`; permissionBinding → `PERMISSION_BINDING`; realmMatch → `IDENTITY` in scope mode (the `realmMatch` resource key stays the neutral-pass discriminator), `IDENTITY, ATTRIBUTES` in attribute mode; date / time / composite → none.
- Composite algebra: pending children are UNKNOWN — never counted, never masked to a settled value. Sound early returns stand (AFFIRMATIVE settles true on a settled-true child, UNANIMOUS settles false on a settled-false child); the tail settles despite pendings only when no resolution could change the outcome (CONSENSUS: `count − p > 0` / `count + p ≤ 0`); otherwise the composite returns pending, uninverted.
- `PermissionEvaluationOptions` gains `pendingPolicies?: 'deny' | 'permit'` (default `deny`). `preEvaluate` passes `permit` instead of `policiesExcluded: [ATTRIBUTES, ATTRIBUTE_NAMES, REALM_MATCH]` — the gate is now derived: a pending grant tree passes, only a tree that settles false with the current bag denies. `policiesIncluded`/`policiesExcluded` remain functional for explicit callers (e.g. the event service's reach derivation).

### `apps/server-core`

- `PermissionBindingPolicyEvaluator` declares `requires: [PERMISSION_BINDING]` — IDENTITY deliberately NOT declared, so a missing identity stays a settled `DATA_MISSING` deny and a scope-restricted / anonymous bearer still fails the pre-gate through `system.default`.
- Its grant loop propagates a **pending junction (Layer-2) policy as a pending grant term** (the disjunction settles false only when every term settled). Without this, an attributes-bound junction grant would settle to deny at the pre-gate — the Layer-2 counterpart of the behavior-preservation invariant, which the issue's notes only covered for Layer-1. The direct scope-mode realm-match call stays engine-bypassing, so realm reach can never pend or be filtered away.

## The bug this fixes

The old pre-gate masked excluded policy types to `{ success: true }`. Under an odd number of inversions the over-approximation flips to under-approximation: `composite { invert, children: [attributes, identity] }` with a satisfying identity was **spuriously denied** at the pre-gate (`NOT(masked-true AND true)`), even though the full policy `NOT(attributes AND identity)` allows whenever the row's attributes fail. With tri-state, the attributes child stays pending, the composite stays pending, `invert` is skipped, and the gate permits — the enriched-bag `evaluate()` remains the authority.

## Behavior preservation (verified)

- attributes-only grant at pre-gate: old neutral-pass ≡ new pending+permit — same outcome, both layers (Layer-1 binding policies and Layer-2 junction policies).
- `AND(identity, attributes)` at pre-gate: identity-false still denies early; identity-true → pending → permit.
- Full `evaluate` with missing data: denied before (evaluator `DATA_MISSING`), denied now (engine pending, `success: false`).
- Third-party evaluators without `requires`: behave exactly as today.
- Scope-restricted bearer (no identity in bag): still denied at the pre-gate via `system.permission-binding`'s settled identity-missing deny.

## Tests

- `test/unit/policy/engine.spec.ts`: requires-gate (pending on missing key, invert not applied to pending, settles when present) + the #3286 inversion regression at engine level, incl. full-bag assertions that the inverted semantics still settle correctly.
- `test/unit/policy/composite.spec.ts` (new): tri-state algebra matrix across AFFIRMATIVE / UNANIMOUS / CONSENSUS + invert-never-applies-to-pending.
- `test/unit/permission/pre-evaluate.spec.ts` (new): derived pre-gate (pending permits, settled-false denies at the gate, full-evaluate semantics intact, explicit `policiesExcluded` backcompat) + the regression (`preEvaluate` no longer spuriously throws).

Verified: `@authup/access` 114/114; server-core `check:types` clean + full suite 1671/1671 (the 2 failed LDAP suite files are the documented pre-existing stale-container bind failures); client-web-kit 150/150; ESLint clean.

Part of #3286 (phases 2–3 — `lower()` pushdown and service integration — follow separately).

## Summary by CodeRabbit

* **New Features**
  * Policy evaluations now distinguish between settled and pending results when required data is unavailable.
  * Added configurable handling for pending policies during pre-evaluation.
  * Composite policies now preserve pending states across decision strategies and inversion.
  * Added structured issues identifying missing evaluation data.

* **Bug Fixes**
  * Prevented incomplete data from producing premature denials or permits.
  * Improved inverted composite policy behavior while required data remains pending.

* **Documentation**
  * Clarified realm matching, pending evaluation, and data availability behavior.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -876,8 +876,9 @@ factor, ANDed with the junction's `policyId` policies) by invoking the
 `RealmMatchPolicyEvaluator` in **SCOPE MODE**: a grant's `realmScope` is matched
 against the resource realm supplied under the **`realmMatch` PolicyData key** (a single id,
 `null`, or an array of ids — `realmScopeMatches` requires the scope to reach every listed
-realm). The realm-match call is made **directly** (not via the policy engine — `realmMatch`
-is in `policiesExcluded`) and stays **outside** the `policies[]` merge, so the policy-free
+realm). The realm-match call is made **directly** (not via the policy engine — so it can
+never be skipped by a caller's include/exclude filters or deferred by the engine's
+data-availability gate) and stays **outside** the `policies[]` merge, so the policy-free
 fail-open drop can never touch realm reach. A **realm-less / anonymous** actor can never
 satisfy `own`/`ownOrNull` (only `any`), and the factor neutral-passes when no `realmMatch`
 key is present (`preEvaluate` / gate checks / realm-less resources).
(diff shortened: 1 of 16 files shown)


tada5hi:
@coderabbitai pause

PR #3292, opened 2026-07-21
feat: compile permissions to row conditions for getMany authorization

tada5hi:
## What

First slice of #3286 phase 3 — permission authorization compiled into the row query. Builds on the tri-state engine (#3290) and condition lowering (#3291).

### `@authup/access`

- `IPermissionEvaluator.compile(ctx) => PermissionCompileResult` — the query-build counterpart of `evaluate()`. Runs the SAME evaluation walk with `withConditions` and classifies the outcome:
  - `allow` — settled true for every row (no restriction),
  - `deny` — settled false for every row,
  - `conditional` — the EXACT residual as a rapiq `ICondition` over row attributes,
  - `post` — not expressible; per-row evaluation remains the sound fallback.
- Multiple names compile as a disjunction (`evaluateOneOf` semantics): any `allow` short-circuits; any non-expressible name degrades the whole result to `post` — pushing part of a disjunction would wrongly exclude rows another name would admit.

### `apps/server-core`

- `PermissionBindingPolicyEvaluator` composes the grant disjunction under `withConditions`: with no resource realm in the bag, the scope-mode reach **pends with its condition** over the row's realm column (default field `realmId`) instead of neutral-passing. `any` stays unrestricted — a policy-free `any` grant settles TRUE, so an admin compiles to `allow` with zero WHERE overhead; `none` reaches nothing. Each grant term is `and(reachCondition, junctionPolicyCondition)`, OR-composed **all-or-nothing** (one non-expressible grant makes the binding non-expressible). The non-`withConditions` paths are byte-identical in behavior.
- `KeyService.getMany` / `TrustAnchorService.getMany` consume the verdict: `conditional` → `appendQueryConditions` (authorization runs as WHERE — **pagination and totals are now exact** for realm-scoped readers); `deny` → constant-false condition (keeps the meta shape); `post` → the existing per-row `evaluate` + `total -= 1` loop stays as fallback (and the plan-039 force-select discipline keeps serving exactly that path).
- `RequestPermissionEvaluator.compile` forwards with the request-scoped identity enrichment. A non-global-scope bearer has no identity in the bag → `system.identity` pends without a condition → `post` → the per-row path denies as today (fail-safe).

### `@authup/server-test-kit`

- `FakePermissionEvaluator.compile` records calls and defaults to `post`, so existing service tests keep asserting their per-row `evaluate` expectations; override per test via `setCompileResult`.

## The bug class this retires (for the converted services)

The per-row drop loops silently corrupted pagination: a page of 20 with 15 forbidden rows returned 5 items and a wrong total, with page boundaries drifting per reader. For key/trust-anchor reads the realm reach + junction policies now filter in SQL, so a `realm_admin` gets correctly paged, correctly counted results.

## Tests

- `packages/access/test/unit/permission/compile.spec.ts` — verdict matrix (allow/deny/conditional/post), disjunction semantics (allow short-circuit, post degradation, settled-false drop), semantic predicate parity via `@rapiq/memory`.
- `test/unit/core/security/permission-binding-evaluator.spec.ts` — new `withConditions` block: own/ownOrNull reach lowering (row-semantics via `compileFilters`), `any` fast-path TRUE, `none` FALSE, reach ∧ junction-policy AND-composition, per-grant OR pairing (reach stays with its own policy), non-expressible junction policy → pending without condition, settled-true policy on own reach → reach condition.
- `test/unit/core/entities/key/service.spec.ts` — conditional pushes the condition into the repository query (replayed via `@rapiq/memory` `applyQuery`) and skips per-row evaluation; allow keeps rows/totals untouched (and still nulls `decryptionKey`); deny matches no row.
- Existing key/trust-anchor HTTP realm-isolation tests now exercise the real compile path end-to-end and pass unchanged.

Verified: access 140/140, server-test-kit 11/11, client-web-kit 150/150, server-core `check:types` clean + full suite 1683/1683 (2 failed suite files = the documented pre-existing LDAP stale-container binds), docs build green, ESLint clean.

## Remaining on #3286

- The ownership-composed drop-loop services (session / event / client / consent / user / attribute services) — these OR an ownership alternative (`sub`/`subKind`, `actorId`) into the gate, so their conversion composes `or(ownershipCondition, compiled.condition)` service-side.
- Include gating via the schema `relations.validate` hooks (rapiq ≥ beta.5).

Part of #3286.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1206,10 +1206,30 @@ mode (field = single-string `attributeName`, default `realmId`; under `withCondi
 an absent `realmMatch` resource key PENDS with the condition instead of neutral-passing
 — the resource realm IS the unknown row column for a query builder) and strict
 single-key attribute mode. Anything non-lowerable stays a per-row post-check (fail-safe:
-`toCondition` returning null/throwing just yields a condition-less pending). Phase 3
-(service integration: `compilePermissionPolicy`, replacing manual realm scoping and the
-per-row `evaluate` + `total -= 1` drop loops, include gating via `relations.validate`)
-is still open on #3286.
+`toCondition` returning null/throwing just yields a condition-less pending).
+
+**Permission compile (#3286 phase 3):** `IPermissionEvaluator.compile(ctx)` →
+`allow | deny | { verdict: 'conditional', condition } | post` is the query-build
+counterpart of `evaluate()` — the same walk with `withConditions`, classified. Multiple
+names compile as a disjunction (`evaluateOneOf` semantics): any `allow` short-circuits,
+any non-expressible name degrades the whole result to `post` (partial OR pushdown would
+wrongly exclude rows). The server-core `PermissionBindingPolicyEvaluator` composes the
+grant disjunction under `withConditions`: with no resource realm present the scope-mode
+reach PENDS with its condition over the row's realm column (default field `realmId`)
+instead of neutral-passing — `any` stays unrestricted (policy-free `any` grants settle
+TRUE → compile `allow`, the admin fast path), `none` reaches nothing — and each grant
+term is `and(reachCondition, junctionPolicyCondition)`, OR-composed all-or-nothing.
+`getMany` consumers (`KeyService`, `TrustAnchorService`) run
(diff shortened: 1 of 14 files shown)


tada5hi:
@coderabbitai full review

Issue #3278, opened 2026-07-19
refactor(client-web-kit): compose queries in the rapiq IR (defineQuery/mergeQueries) instead of forwarding build-input objects


tada5hi:
## Motivation

Since #3276 the SDK boundary already accepts the rapiq IR (`EntityQueryInput<T> = QueryBuildInput<T> | IQuery`), but client-web-kit's managers still forward and merge plain build-input objects (`QueryInput<T>`, depth-bounded aliases) through props, context callbacks and list meta. That leaves three structural gaps:

1. **Accidental merge semantics.** The collection manager combines context query + props query + pagination state with a generic object merger (smob). Array-vs-record filter shapes merge by coincidence, not by contract. rapiq's `mergeQueries` defines left-priority semantics (fields/relations/sorts keyed by name, pagination per-property), and `Filters.merge()` vs `and()`/`or()` makes replace-vs-scope an explicit choice.
2. **Displaceable scoping.** A component that injects a realm scope via object spread can have it silently overwritten by a later merge (search filters, sort changes). An injected `and(eq('realmId', x))` condition cannot be displaced — the same guarantee the server pipeline has.
3. **No compound conditions in the UI.** The record-form build input cannot express OR. A search box that should query `or(contains('name', q), contains('displayName', q))` is impossible today; a `queryFilters` context callback returning an `ICondition` (condition helpers: `eq`, `contains`, `or`, …) unlocks it.

Bonus: `IQuery` is not generic-recursive, so IR-typed props have no TS7056 depth problem — the depth-bounded `QueryInput<T, 3>` aliases introduced in #3276 (see rapiq#776) become unnecessary wherever props move to the IR.

## Design

**Keep typed authoring at the edges; compose in IR space.**

- Pages keep writing typed input, desugared immediately: `defineQuery<User>({ filters: { realmId: [...] } })` — NestedKeys checking is preserved at construction.
- Component props accept `QueryInput<T> | IQuery` during the transition and normalize via `defineQuery` at the prop boundary (duck-type `isQuery` guard until rapiq#774 ships the guard family).
- Collection/record managers store `IQuery` and compose with `mergeQueries`; injected scopes use the `Filters.and()` combinator so they survive later merges.
- `EntityCollectionManagerCreateContext.queryFilters` changes contract from `(q) => FiltersBuildInput` to `(q) => ICondition` (or accepts both during migration), enabling compound search conditions.
[...]

PR #3297, opened 2026-07-21
feat(server-core): gate include authorization at decode via relations.validate

### LINKED ISSUE #3286: RFC: data-availability-driven policy evaluation (tri-state engine, derived preEvaluate)

### LINKED ISSUE #3295: Include-authorization gating via schema relations.validate at decode time (#3286 phase 3, part 4)
## Context

Final piece of #3286 (after #3290–#3293 and the service conversions tracked in #3294): the **include-authorization gap**. Today a rapiq `include=` on a list/read endpoint joins the target relation with no check of the *target entity's* read permission — the allow-lists (`relations.allowed` in the per-entity schemas under `core/entities/*/schema.ts`) bound *what* is joinable, but not *who* may read it. The consent adapter works around this by omitting `client` from `relations.allowed` entirely (see architecture.md → OAuth2 Consent) — a blunt instrument that also blocks legitimate readers.

## Proposal (from #3286)

Run the **derived pre-gate** for the include target's read permission at decode time, via the rapiq per-parameter `validate` hooks + request-context threading (tada5hi/rapiq#806 / #807, shipped in 2.0.0-beta.5; the repo is on beta.6):

- The schemas already carry `schemaMapping: { <relation>: EntityType.X }` — the natural place to derive "the target entity type" per relation; map entity type → read permission name.
- `decodeQuery(input, { schema, ... })` (`core/query/`) gains an actor/context input threaded through rapiq's decode context, so a `relations.validate` hook can call `actor.permissionEvaluator.preEvaluate({ name: <target read permission> })` — the data-availability-derived gate from #3290, which is exactly right here: row data is unknown at decode time, so only a settled-false (actor can never read that entity type) rejects.
- Decide and pin the deny behavior: strip the include silently vs. reject the request (400/403). Rejecting is more honest; stripping matches the fail-soft flavor of the allow-lists. Whichever is chosen, document it in `.agents/architecture.md` (Query IR flow section) and cover both a permitted and a denied include in the HTTP tests.

## Explicitly out of scope

**Relation-scoped residuals for included rows** — narrowing *which* rows of the included relation are visible (the per-row counterpart of the pre-gate) needs rapiq-side support first: tada5hi/rapiq#810 (relation-scoped conditions RFC). This issue is only the decode-time gate for the include as a whole.

## Pointers

- `core/query/module.ts` — `decodeQuery`, `SchemaRegistry`, `appendQueryConditions`
- `core/entities/*/schema.ts` — per-entity schemas (`relations.allowed`, `schemaMapping`)
- `.agents/architecture.md` → *Query IR flow* and *preEvaluate is derived from data availability*
- rapiq validate-hook surface: `FiltersOptions.validate` / per-parameter schema validate hooks (core 2.0.0-beta.5 changelog, tada5hi/rapiq#807)

tada5hi:
Closes #3295. Closes #3286 — this is the final piece of the RFC; the deliberately-deferred residuals live upstream (tada5hi/rapiq#810 relation-scoped conditions, tada5hi/rapiq#815 dotted-key bypass).

## What

A rapiq `include=` on a list/read endpoint now runs the derived pre-gate (#3290) for the **target entity type's read permission** at decode time. Previously the `relations.allowed` lists bounded *what* is joinable, but not *who* may read it.

- `decodeQuery` is now **async** (`codec.decodeAsync` — the gate awaits the actor's pre-gate) and takes an `actor` option, threaded as the rapiq decode context (tada5hi/rapiq#806/#807 surface, shipped in 2.0.0-beta.5).
- Every schema's `relations.validate` is `createRelationsReadGate(schemaMapping)` (`core/query/relations.ts`): per include segment the relation maps to its target entity type and `preEvaluateOneOf` runs the type's read-permission disjunction. The gate map mirrors each target service's own `getMany` read gate; POLICY sits under the PERMISSION family; REALM and IDENTITY_PROVIDER stay ungated (both lists are anonymous surfaces). Nested paths (`include=user.realm`) gate hop by hop.
- **Deny = silent strip** (fail-soft, matching the allow-lists): the include — and every dotted filter/sort/field key riding the denied relation path — is pruned; the request succeeds with the un-joined row shape.
- **Caller classes**: a SYSTEM decode (no actor) runs unrestricted; every REQUEST decode passes the actor — `buildActorContext` supplies one for anonymous requests too (the authorization middleware sets the evaluator before the header check), so a guest's gated includes strip.

## Also fixed en route

- **Syntactic-fallback hole closed**: a schema without a `relations` key fell back to rapiq's syntactic name check, so role / scope / key / trust-anchor / role-attribute / user-attribute / user-authenticator silently accepted arbitrary includes ungated. Every schema now declares `relations` explicitly (`event` / `realm` pin `allowed: []` — their entities have no joinable relations, so a syntactic include died at the TypeORM join as a 500).
- **Latent 500 on every junction `include=` since #3278**: with no root `fields` declared, an include decoded relation-only fields and the typeorm adapter's `select()` replace dropped every root column — including the id TypeORM's DISTINCT-id pagination wrapper needs (`no such column: distinctAlias.userRole_id`). The 8 junction + 2 attribute schemas now pin complete `fields.default` lists (the consent-schema precedent). Note: these lists must enumerate every scalar column — boot validation checks key existence, not completeness.

## Known residual (upstream)

A dotted filter/sort/field key whose relation is NOT explicitly included bypasses the hook — the SQL adapters auto-join such paths without a `relations.validate` pass. Tracked as tada5hi/rapiq#815 (the removal trigger for the residual note in `.agents/architecture.md`).

## Tests

- `test/unit/core/query/module.spec.ts`: gate matrix — permitted include (+ asserted permission names), settled-false strip (query survives), per-relation partial strip, dotted-key pruning on a denied include, ungated target with a deny-all actor, actor-less (system) decode.
- `test/unit/http/controllers/entities/include-authorization.spec.ts`: end-to-end — admin gets joined `user`/`role`; a `USER_ROLE_READ`-only identity gets rows with both includes stripped; a `USER_ROLE_READ + ROLE_READ` identity keeps only `role`.
- Full server-core suite green (the two LDAP spec failures are the known stale-container issue, unrelated).

## Summary by CodeRabbit

* **New Features**
  * Added actor-aware query processing so results and included relationships respect the caller’s permissions.
  * Unauthorized relationship includes are now silently removed while base records remain available.
  * Protected related fields in filters, sorting, and projections are also removed when access is denied.
  * Added explicit default projections to improve safe and consistent response data.
* **Bug Fixes**
  * Prevented unauthorized related data from being exposed through consent and session-management queries.
* **Tests**
  * Added coverage for allowed, denied, and partially denied relationship includes.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -142,17 +142,52 @@ usable at the service level and nothing in core depends on TypeORM:
   entity metadata.
 - **Registry + codec + decode** — `core/query/` owns the `SchemaRegistry`
   (all entity schemas), the URL codec bound to it (decodes both the v2
-  expression dialect and legacy v1 bracket payloads), and
-  `decodeQuery(input, { schema, parameters? })` → `Query` (IR). `schema`
-  accepts the colocated schema object or a registered name; `parameters`
-  restricts which parameters are parsed/defaulted (e.g. `['filters']` for the
-  session bulk revoke, `['fields', 'relations']` for single reads).
+  expression dialect and legacy v1 bracket payloads), and the **async**
+  `decodeQuery(input, { schema, parameters?, actor? })` → `Promise<Query>`
+  (IR; async because the schema validate hooks await the actor's permission
+  pre-gate). `schema` accepts the colocated schema object or a registered
+  name; `parameters` restricts which parameters are parsed/defaulted (e.g.
+  `['filters']` for the session bulk revoke, `['fields', 'relations']` for
+  single reads); `actor` rides the decode context into the schema validate
+  hooks (see *Include authorization* below).
 - **Services decode once** — `getMany` calls
-  `this.repository.findMany(decodeQuery(query, { schema: roleSchema }))`;
+  `this.repository.findMany(await decodeQuery(query, { schema: roleSchema, actor }))`;
   multi-branch services (session/consent/event) decode into a local and pass
   the IR to every branch. Allow-lists, defaults and pagination bounds are all
   applied at decode time, so services can also inspect or compose the IR
   before handing it down.
+- **Include authorization — the relations read gate (#3295)** — every
+  schema's `relations.validate` hook is `createRelationsReadGate(schemaMapping)`
+  (`core/query/relations.ts`; schemas import that FILE directly, never the
+  `core/query` barrel — the barrel reaches `module.ts`, which imports every
+  schema, and the cycle would TDZ-crash). Per include segment the hook maps
+  the relation to its target entity type (the schema's own `schemaMapping`)
+  and runs the derived pre-gate (`preEvaluateOneOf`, #3290) for that type's
+  read-permission disjunction (`RELATION_TARGET_READ_GATES`, mirroring each
+  target service's own `getMany` gate; POLICY sits under the PERMISSION
+  family; REALM and IDENTITY_PROVIDER are deliberately ungated — both lists
+  are anonymous surfaces). Nested paths gate hop by hop. **Deny = silent
+  strip** (fail-soft, matching the allow-lists): the include — and every
+  dotted filter/sort/field key riding the denied relation path — is pruned,
+  the request still succeeds with the un-joined row shape. Caller classes:
+  a SYSTEM decode (no `actor` option) runs unrestricted; every REQUEST
+  decode passes the actor (`buildActorContext` supplies one for anonymous
+  requests too — an anonymous actor holds no grants, so its gated includes
+  strip). Every schema now declares `relations` explicitly (an omitted
+  allow-list falls back to rapiq's syntactic name check — that hole is
+  closed; `event`/`realm` pin `allowed: []`). **Known residual:** a dotted
+  filter/sort/field key whose relation is NOT explicitly included bypasses
+  the hook (the SQL adapters auto-join such paths) — upstream
+  tada5hi/rapiq#815 is the removal trigger.
+- **Junction/attribute schemas pin `fields.default`** — with no root fields
+  declared, an `include=` decodes the relation's default fields ONLY, and
+  the typeorm adapter's `select()` replace then drops every root column
+  (including the id TypeORM's DISTINCT-id pagination wrapper needs — every
+  junction `include=` 500'd before #3295 surfaced it). The explicit
+  `default` list keeps the root projection riding alongside relation
+  fields; it must enumerate EVERY scalar column (boot validation checks
+  key existence, not completeness — a column missing from `default`
+  silently vanishes from the API response).
 - **Server-derived scopes ride `appendQueryConditions(query, ...conds)`**
   (core/query) — an immutable AND-wrap of the filter tree
   (`IFilters.and`, the wrap-and-inject primitive; parameter nodes carry
@@ -1248,8 +1283,9 @@ term); `UserAuthenticatorService` composes `eq('userId', <actor id>)` the same w
 still sanitizes secret/codes on the compiled path; `RoleAttributeService` is a pure
 gate (key/trust-anchor shape, no ownership term).
 `FakePermissionEvaluator.compile` defaults to `post` so service tests keep their
-per-row expectations; override via `setCompileResult`. Still open on #3286:
-include gating via `relations.validate`.
+per-row expectations; override via `setCompileResult`. The final #3286 piece —
+include gating via `relations.validate` — shipped as #3295: see *Query IR flow →
+Include authorization*.
 
 ### EA loading on tree roots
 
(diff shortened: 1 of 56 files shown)


tada5hi:
@coderabbitai full review

PR #3301, opened 2026-07-21
fix: preserve API sub-path when building authorize & OAuth2 URLs

### LINKED ISSUE #3298: client-web: authorize-URL generator drops API path segment and query is non-reactive
## Summary

In `apps/client-web/pages/clients/[id]/url.vue`, the OAuth2 authorize-URL generator builds its URL with `new URL('authorize', config.public.apiUrl)`. Because the `URL` constructor replaces the **last** path segment of the base unless the base ends with a trailing slash, a `config.public.apiUrl` without a trailing slash silently drops its final segment.

## Steps to reproduce

Set the API URL without a trailing slash, e.g. `apiUrl = "https://example.com/api"`, then open a client's **URL** tab.

```js
new URL('authorize', 'https://example.com/api').href
// => "https://example.com/authorize"   ❌  ('/api' is dropped)

new URL('authorize', 'https://example.com/api/').href
// => "https://example.com/api/authorize" ✅
```

The generated authorize URL therefore points at the wrong path and the OAuth2 flow fails when the API is served under a sub-path.

## Suggested fix

Normalize the base to guarantee a trailing slash before constructing the URL:

```diff
 const generatedUrl = computed(() => {
-    const link = new URL('authorize', config.public.apiUrl);
+    const apiUrl = config.public.apiUrl as string;
+    const baseUrl = apiUrl.endsWith('/') ? apiUrl : `${apiUrl}/`;
+    const link = new URL('authorize', baseUrl);
     link.searchParams.set('client_id', props.entity.id);
     link.searchParams.set('response_type', 'code');
     ...
```

## Secondary (minor)

In the same file the client-scope query is built as a static object:

```js
const query = defineQuery<ClientScope>({
    filters: { clientId: props.entity.id },
    relations: ['scope'],
});
```

`props.entity.id` is captured once at setup, so the query is not rebuilt if the entity id changes while the component instance is reused. Wrapping it in a `computed` keeps it reactive.

## Context

Found while mirroring the `clients` pages into a downstream portal (dnpm-dip/portal#1267). Both points were flagged by an automated review there and apply equally to the upstream `client-web` implementation.

tada5hi:
## Summary

`new URL(relativePath, base)` drops the base's **final path segment** when the base has no trailing slash, so an API served under a sub-path (`https://example.com/api`) silently loses it:

```js
new URL('authorize', 'https://example.com/api').href
// => "https://example.com/authorize"    ❌  ('/api' dropped)
```

This footgun appeared at several call sites, not just the one reported in #3298:

- **`clients/[id]/url.vue`** — the OAuth2 authorize-URL generator (the reported bug).
- **`components/sidebar.vue`** — the docs link (`new URL('docs/', api.getBaseURL())`).
- **`core-http-kit` `Client`** ⭐ — resolved **every** OAuth2 endpoint the same way (`authorize`, `token`, `token/introspect`, `users/@me`), so any consumer of the typed client with a sub-path API URL had broken token/authorize/introspect/userinfo endpoints. Higher-impact than the UI bug.

## Changes

- Add a shared **`buildURL(base, path)`** helper to `@authup/kit` (`packages/kit/src/url.ts`): normalizes the base to a trailing slash and strips a leading slash off `path`, returning a mutable `URL` for `searchParams` composition. (`server-core` already had a private `resolveURL` doing this — the pattern recurs, so it now lives in `kit`.)
- Route all three call sites through `buildURL`.
- Secondary fix from #3298: wrap the client-scope `query` in `computed` so it stays reactive to a changing entity id.

`login/index.vue` was reviewed and left unchanged — it resolves against `window.location.origin` (no path to drop), which is correct.

## Tests

- `packages/kit` — 7 new `buildURL` unit cases (sub-path with/without trailing slash, root base, leading-slash strip, mutable searchParams, invalid base). All 28 kit tests pass.
- `packages/core-http-kit` — new `endpoint-resolution.spec.ts` asserting the resolved OAuth2 endpoints preserve the sub-path (black-box via `FakeClient` request recording + `authorize.buildURL()`). All 21 tests pass.

Closes #3298

## Summary by CodeRabbit

* **New Features**
  * Added consistent URL construction for API and OAuth endpoints, including base URLs with sub-paths.
  * Added a shared URL utility that preserves paths and supports query parameter updates.

* **Bug Fixes**
  * Fixed endpoint links that could lose part of the configured API base path.

* **Tests**
  * Added coverage for URL construction with trailing slashes, sub-paths, root URLs, query parameters, and invalid bases.

### DIFF
--- a/apps/client-web/components/sidebar.vue
+++ b/apps/client-web/components/sidebar.vue
@@ -6,6 +6,7 @@
   -->
 <script lang="ts">
 
+import { buildURL } from '@authup/kit';
 import { TranslatorTranslationAppKey, TranslatorTranslationNamespace } from '@authup/i18n';
 import { ITranslateT } from '@ilingo/vue';
 import {
@@ -42,7 +43,7 @@ export default defineNuxtComponent({
         });
 
         const api = injectHTTPClient();
-        const docsUrl = computed(() => new URL('docs/', api.getBaseURL()).href);
+        const docsUrl = computed(() => buildURL(api.getBaseURL(), 'docs/').href);
 
         const translate = useTranslator();
         const locale = injectTranslatorLocale();
--- a/apps/client-web/pages/clients/[id]/url.vue
+++ b/apps/client-web/pages/clients/[id]/url.vue
@@ -1,5 +1,6 @@
 <script lang="ts">
 import { defineQuery } from '@rapiq/core';
+import { buildURL } from '@authup/kit';
 import { TranslatorTranslationAppKey, TranslatorTranslationEntityKey, TranslatorTranslationNamespace } from '@authup/i18n';
 import { AClientScopes, useTranslations, useTranslationsForNamespace } from '@authup/client-web-kit';
 import type { Client, ClientScope } from '@authup/core-kit';
@@ -48,7 +49,7 @@ export default defineNuxtComponent({
         const config = useRuntimeConfig();
 
         const generatedUrl = computed(() => {
-            const link = new URL('authorize', config.public.apiUrl);
+            const link = buildURL(config.public.apiUrl as string, 'authorize');
             link.searchParams.set('client_id', props.entity.id);
             link.searchParams.set('response_type', 'code');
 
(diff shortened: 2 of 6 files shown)


tada5hi:
@coderabbitai full review

PR #3324, opened 2026-07-26
fix: complete schema field projections and re-target role client FK

tada5hi:
## What

Three fixes plus the rapiq bump, in one:

**1. Schema field projections were silently dropping columns.** rapiq v2 derives the root projection from `fields.default`/`fields.allowed`, so a column in neither list is simply absent from every collection response — no error. Four schemas had drifted:

| Schema | Missing from projection |
|---|---|
| `role` | `builtIn`, `clientId` (the reported bug: provisioned roles rendered as not built-in) |
| `user` | `status`, `statusMessage` |
| `session` | `authMethod`, `mfaAt` |
| `clientScope` | `createdAt`, `updatedAt` (also added to the core-kit domain type) |

`role.builtIn`/`clientId` are also added to `filters.allowed` for parity with permission/scope.

**2. Boot-time completeness guard.** `validateEntitySchemas` now also runs `assertSchemaFieldsCoverEntity`: every selectable column must appear in the schema's field allow-list, so this class of drift fails startup instead of vanishing from the API. `select: false` secrets are exempt automatically; the one deliberate omission (`userAuthenticator.parameters`) is an explicit `SCHEMA_FIELD_EXCLUSIONS` entry.

**3. `auth_roles.client_id` re-targeted onto `auth_clients`.** The entity had `@ManyToOne(() => RealmEntity)` on the client join column (copy-paste of the realm relation), and migration `1740991051622` baked the FK to `auth_realms` — so a client-scoped role could never reference an actual client. Same defect and fix shape as the `auth_permissions.client_id` re-target in `1766830857009`, with a narrower cleanup (only rows whose `client_id` doesn't resolve to a client are dropped). Migration `1784970000000-RoleClientForeignKey`, both dialects.

**4.** `@rapiq/*` `^2.0.0-beta.9` → `^2.0.0-beta.10`.

## Verification

- Migration round-trip on fresh MySQL + PostgreSQL: `run` (14) → `revert` × 14 → `run` (14); FK confirmed on `auth_clients` via `pg_constraint` / `information_schema` after each forward pass
- `createSchemaBuilder().log()` shows no pending `auth_roles` queries — entity and migrated schema agree
- Full monorepo build (22 projects) + full test suite green (the one `session.spec.ts` failure in the aggregate run is the known shared-admin flake; passes in isolation)
- Two new spec cases pin the completeness guard (missing selectable column rejected, `select: false` exempt)

## Summary by CodeRabbit

* **New Features**
  * Expanded default user responses with status and status message details.
  * Added client-scope timestamps to default responses.
  * Added role fields for built-in status and client association, including filtering support.
  * Added session authentication method and MFA timestamp fields for selection.

* **Bug Fixes**
  * Corrected role-to-client associations and database constraints.
  * Prevented missing selectable fields from silently disappearing in API responses.

* **Validation**
  * Added stricter startup checks to detect incomplete response field configurations.


tada5hi:
@coderabbitai pause

PR #3331, opened 2026-07-27
fix(errors): keep guard inheritance match for JSON-rehydrated errors

### LINKED ISSUE #3042: errors: duck-type guards lose inheritance match for JSON-rehydrated errors
## Context

After #3041 (PR), `@authup/errors` uses a Symbol.for-keyed `@instanceof` chain so subclass instances accumulate their full ancestor chain at construction time. Guards have a fast path (chain lookup) and a slow path (code shape check):

```ts
export function isBadRequestError(input: unknown): input is BadRequestError {
    if (hasInstanceof(input, BAD_REQUEST_ERROR_INSTANCE)) return true;   // fast
    if (!isAuthupError(input)) return false;
    return input.code === ErrorCode.BAD_REQUEST;                          // slow
}
```

This works perfectly **in-process** (fast path uses the marker chain) and **cross-realm with markers preserved** (Symbol.for is registry-backed, same symbol everywhere).

It **breaks for JSON-rehydrated errors**: when an `AuthupError` round-trips through `JSON.stringify` / `JSON.parse`, the symbol-keyed `@instanceof` array is dropped (symbols don't serialize). The rehydrated plain object has only `code`, `message`, etc. The slow path then checks `input.code === BAD_REQUEST` — but for a `RegistrationDisabledError` (code `registration_disabled`, extends `BadRequestError`), the match fails.

```ts
const err = new RegistrationDisabledError();
const rehydrated = JSON.parse(JSON.stringify(err));

isBadRequestError(err);           // true  (fast path via @instanceof chain)
isBadRequestError(rehydrated);    // false (slow path: code !== 'bad_request')
isRegistrationDisabledError(rehydrated); // true  (slow path: code === 'registration_disabled')
```

## Why this matters

Mostly hypothetical today — Authup doesn't ship error objects across process boundaries via JSON in the production path. But the gap will bite if/when:

- A frontend rehydrates an error from an HTTP JSON body and wants to handle "any kind of bad request" uniformly
- A worker/main thread or fork-mode setup IPC's errors
- A logging adapter rehydrates from structured logs and wants to categorise

For these, callers today have to either:
1. Read `code` directly and pattern-match (loses class identity entirely)
2. Map `code` to status via `httpStatusFromCode` and check the HTTP-family (loses sub-code precision)

### Option A — Serialize the chain as a string array

`AuthupError.toJSON()` emits `@instanceof` as a string list (using `symbol.description`):

```ts
override toJSON() {
    return {
        ...super.toJSON(),
        issues: this.issues,
        '@instanceof': this['@instanceof'].map(s => s.description),
        ...(this.data ?? {}),
    };
}
```

Guards then accept either:
- The native `symbol[]` chain (fast path, in-process)
- A `string[]` chain (rehydrated from JSON)

Pros: zero maintenance, fully accurate, scales with the hierarchy automatically.
Cons: adds a few bytes per error in JSON output; the `@instanceof` field becomes part of the public wire format.

### Option B — Code → ancestor codes lookup

Maintain a static table in `@authup/errors`:

```ts
export const CODE_HIERARCHY: Readonly<Record<string, readonly string[]>> = {
    [ErrorCode.REGISTRATION_DISABLED]: [ErrorCode.BAD_REQUEST],
    [ErrorCode.PASSWORD_RECOVERY_DISABLED]: [ErrorCode.BAD_REQUEST],
    [ErrorCode.HTTP_BEARER_TOKEN_MALFORMED]: [ErrorCode.IDENTITY_UNAUTHORIZED],
    // ...
};
```

Slow-path guard checks `input.code === MY_CODE || CODE_HIERARCHY[input.code]?.includes(MY_CODE)`.

Pros: minimal wire-format impact.
Cons: every new subclass must be added to the table — easy to forget; duplicates information already encoded in the class hierarchy.

### Option C — Skip and document

Slow path stays leaf-only. JSON consumers use `code` directly or via `httpStatusFromCode`. Document the limitation.

Pros: zero added complexity.
Cons: the inheritance-aware duck guards are silently incomplete for cross-process inputs.

## Recommendation

**Option A** when this becomes a real problem; **Option C** for now.

Option A is the right long-term answer — it's self-maintaining and the JSON overhead is small (4 short strings for a 3-deep chain). But there's no concrete consumer today that needs it, and Authup's pre-1.0 wire format isn't yet locked in. Worth doing as a follow-up before we ship 1.0 (locks the JSON shape), not blocking on it right now.

Tracking here so it doesn't slip through.

tada5hi:
## Summary

Implements Option A from #3042: the `Symbol.for`-keyed `@instanceof` class-marker chain now survives JSON serialization, so the duck-type guards keep their inheritance-aware matching for errors rehydrated from a JSON body.

The mechanism was upstreamed while this PR was open — tada5hi/ebec#448 landed as `@ebec/core` 1.2.0 (`BaseError.toJSON()` emits the chain natively; `matchesInstanceof` / `serializeInstanceofChain` helpers) — so the final state here is adoption, not a local implementation:

- **Deps**: `@ebec/core` → ^1.2.0 (errors), `@ebec/http` → ^4.2.0 (server-core); single deduped 1.2.0 across the tree.
- `@authup/errors` re-exports `matchesInstanceof` + `serializeInstanceofChain` from `@ebec/core` (the interim local `instanceof.ts` was removed again in the second commit).
- `AuthupError.toJSON()` re-stamps the chain **after** its `data` spread — `super.toJSON()` emits it before the spread, where a `data` key could displace it (pinned by a test). All `AuthupError` subclasses (specs' OAuth2/JWT/JWK errors, access' permission/policy errors, server-core's workflow errors) inherit this — no per-class work.
- Every duck-guard module's fast path is swept onto `matchesInstanceof`: `@authup/errors` (`isAuthupError` + all built-in guards), `@authup/specs` (OAuth2/JWT/JWK), `@authup/access` (permission/policy), server-core (registration, password-recovery, realm-cipher). For `Symbol.for` markers the string match is exactly as precise as the symbol match — the description *is* the registry key. In-process `hasInstanceof` control-flow sites (e.g. the authorize catch dispatch) are deliberately untouched.

```ts
const rehydrated = JSON.parse(JSON.stringify(AuthHeaderError.unsupportedType('X')));
isUnauthorizedError(rehydrated); // now true — previously false (leaf-only code fallback)
```

## Behavior notes

- **Wire format**: error payloads (HTTP error middleware, `serializeError`) now carry an `@instanceof` string list — the accepted Option A trade-off (pre-1.0 format, extension members are legal in OAuth2 error responses per RFC 6749 §5.2). Old-build consumers ignore the extra key; new-build consumers reading old-build JSON fall back to the previous leaf-only code match (pinned by a test).
- **No subtractive changes**: the chain fast path only ever *adds* matches. Sibling classes sharing a code (`BadRequestError` / `ValidationError`) still cross-match through the code slow path, exactly as in-process today (pinned by a test).
- **Trust**: guards remain duck-typed — a forged body could always claim any `code`; claiming a chain is the same trust class. No guard runs over client-supplied request bodies server-side.

## Testing

- New `packages/errors` test suite (20 tests): chain serialization, dual-form matching, round-trip guard matrix (subclass ancestor match, double round-trip, legacy chain-less payloads, data-key spoof resistance).
- Full monorepo suite re-run after the ebec bump: 16 test targets green, 0 nx cache hits — confirms no existing error-body assertion breaks on the new wire field (incl. `@ebec/http` errors now emitting the chain via `BaseError`).
- `tsc --noEmit` over the errors package (incl. tests) and server-core's `check:types` clean; eslint clean on all changed files.

closes #3042


tada5hi:
@coderabbitai full review

PR #3336, opened 2026-07-28
fix: bind the login-throttle window start as a UTC literal

tada5hi:
## Problem

The failed-login throttle (plan 053) never trips on PostgreSQL when the server host runs a non-UTC timezone. `EventRepositoryAdapter.countRecent` bound a JS `Date` object — the only such site in the repo — against the DB-native `created_at` timestamp:

- `created_at` is stamped by the **database server's (UTC) clock** (column `DEFAULT now()` / `CURRENT_TIMESTAMP(6)`; TypeORM emits literal `DEFAULT` for undefined create-date columns).
- The pg driver serializes a bound `Date` from **local** getters with a host offset suffix, and postgres coerces the untyped param to `timestamp without time zone` — silently discarding the offset. On a UTC+2 host the window start lands 2h in the future → count is always 0 → no 429.
- mysql2 (`timezone: 'local'`) serializes local wall-clock with no offset marker at all — same failure class.
- sqlite only worked by accident: TypeORM's sqlite driver converts a bound `Date` to a UTC string itself (`mixedDateToUtcDatetimeString`).

## Fix

Bind that same UTC wall-clock literal explicitly (`toISOString()` with `T`/`Z` stripped) — all three dialects parse it identically, and the host-timezone dependency is gone. Every other time comparison in the repo targets `varchar(28)` ISO columns with ISO-string binds and is unaffected.

## Tests

New DB-level regression spec `test/unit/app/modules/database/event-repository.spec.ts` exercising `countRecent` boundaries against the suite dialect with `created_at` left to the column DEFAULT (stamping it app-side would hide the mismatch). Verified from a CEST host: **red pre-fix / green post-fix on postgres**; green on mysql and sqlite. The existing `grant-password-throttle.spec.ts` (previously failing under `test:psql`) passes on all three dialects.

Origin: pre-existing defect surfaced by plan-073's `test:psql` run (roadmap Active Defects).

### DIFF
--- a/apps/server-core/src/app/modules/database/repositories/event/repository.ts
+++ b/apps/server-core/src/app/modules/database/repositories/event/repository.ts
@@ -93,9 +93,17 @@ export class EventRepositoryAdapter implements IEventRepository {
     }
 
     async countRecent(filter: EventCountRecentFilter): Promise<number> {
+        // created_at is stamped by the DATABASE server's (UTC) clock, while a
+        // bound Date object is serialized by the pg/mysql drivers in the
+        // HOST's local timezone — on a non-UTC host that shifts the window by
+        // the UTC offset (postgres: `timestamp without time zone` discards
+        // the offset marker entirely). Bind a UTC wall-clock literal instead;
+        // all three dialects parse it identically.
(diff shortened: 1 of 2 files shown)


tada5hi:
@coderabbitai full review

PR #3339, opened 2026-07-28
feat: configurable trust-proxy contract for client-IP derivation

### LINKED ISSUE #3230: Make getRequestIP proxy trust configurable (trustProxy is hardcoded)

**Reported in:** PR #3229, review comment by @coderabbitai (comment 3564825004)
**File:** every `getRequestIP` call site (9 across `apps/server-core/src/adapters/http/` — all grant adapters, the identity-provider controller, and the request event-context middleware `request/event-context.ts` added in #3229, which feeds actor/request attribution for audit rows)

## Description

Every call site uses `getRequestIP(event, { trustProxy: true })`, trusting `X-Forwarded-For` unconditionally. On a deployment where authup is exposed directly (no reverse proxy), a client can spoof the recorded IP. This predates PR #3229, but that PR raises the stakes: the failed-login throttle keys on `(identifier, ip)`, so a direct-deployment attacker could rotate `X-Forwarded-For` values to dodge the pair key (the default-on per-IP rate limiter shares the same trust assumption). Session/session-token `ip_address` inventory and the new audit-event `request_ip_address` column are affected the same way.

PR #3229 documents the constraint on the throttle config ("deploy behind a trusted proxy that overwrites the header"); this issue tracks making it configurable.

## Suggested fix

- New config key `trustProxy: boolean` (env `TRUST_PROXY`), default `true` (preserves current behavior; most deployments sit behind a proxy per the nginx guide).
- Thread it from `Config` through the grant/controller factories into the `getRequestIP` call sites (same optional-context pattern as `logger`).
- Optionally support a CIDR allowlist later; boolean first.

tada5hi:
Closes #3230.

## Problem

Every `getRequestIP` call site hardcoded `{ trustProxy: true }` — trust every hop of `X-Forwarded-For`. A DIRECT client (no proxy in front, or a proxy that doesn't strip the header) can therefore spoof its IP, forging: the IP half of the `(identifier, ip)` login-throttle key, audit-event `requestIpAddress` attribution, session / session-token inventory rows, the MFA verify context, and the identity-provider OAuth2 state binding.

## Design

routup already ships express-parity trust-proxy support (`TrustProxyInput`: `boolean | hops | CIDR/preset string | string[] | fn`, backed by `proxy-addr`), resolved from the app options at every request-helper call site — so this lands as **one config key threaded to one construction site**, not the per-grant context plumbing the issue sketched:

- Config `trustProxy` (env `TRUST_PROXY`), typed `boolean | number | string | string[]` (the function form stays out of serializable config). **Default `true`** per the issue (preserves behavior for the documented nginx deployment); the config reference documents pinning `1` / `loopback` / a CIDR as the hardened posture.
- `new App({ options: { trustProxy: config.trustProxy } })` — all nine hardcoded call-site options removed; every helper inherits the app contract (the access-log line now agrees with audit rows too).
- `registerRateLimitMiddleware` gains an explicit `keyGenerator` — `@routup/rate-limit`'s default hardcodes `{ trustProxy: true }` and ignores app options.
- `TRUST_PROXY=1` parses as **one hop**, not boolean true — the integer form wins over the boolean word sets (`true/false/yes/no/on/off` still parse as booleans; anything else passes through as a proxy-addr allowlist, validated loud at boot).

## Tests

- Config: default assertion, all accepted forms, rejection of negative/fractional hops, env-form parsing matrix.
- HTTP (`grant-password-throttle-trust-proxy.spec.ts`): under `trustProxy: false`, a rotating `X-Forwarded-For` neither evades the throttle (429 still trips on the socket key) nor forges audit rows; under `trustProxy: 1`, the forwarded address IS the recorded audit IP — a true wiring detector (fails if the config stops reaching the router, since routup's own default trusts nothing).
- Full server-core suite green (1752).

Docs: config reference (js/conf/env blocks) + the login-throttle caveat now points at `trustProxy` instead of hand-waving at the proxy.

## Audit follow-up (second commit)

An adversarial review found that the integer/boolean canonicalization lived only on the ENV path: a quoted `trustProxy: "1"` from a config file passed the validator as a plain string, and proxy-addr silently compiled it as single-integer IPv4 long-value notation — **an allowlist of `0.0.0.1`** instead of one trusted hop (verified against the installed proxy-addr). Fixes:

- Shared `canonicalizeTrustProxy` now runs in `normalizeConfig`, covering every config surface (env, `.conf`, file, programmatic). Integer wins over boolean words; strings lowercase (proxy-addr matches its presets case-sensitively, so `Loopback` no longer crashes boot).
- The explicit allowlist **array** form rejects mis-typed scalar entries (`['1']`, `['true']`) loud at validation instead of compiling a bogus entry.
- The `new App(...)` construction wraps proxy-addr's compile error with the offending config key name (`Invalid trustProxy (TRUST_PROXY) configuration: …`) — previously an unattributed "invalid IP address" crash.
- Docs: the access-log IP also follows the trust contract now (it was accidentally trust-nothing before — the one call site with no explicit option), noted in the security comment.

The review's main assignment — whether app-level `trustProxy` changes `getRequestHostName`/`getRequestProtocol` behavior — came back empty: **zero consumers** of those helpers exist in server-core src/dist or any `@routup/*` middleware; discovery documents, redirects, and mail links all derive from `publicUrl`. Full suite re-verified green (1756).

## Summary by CodeRabbit

* **New Features**
  * Added configurable proxy trust settings through server configuration and the `TRUST_PROXY` environment variable.
  * Added validation and normalization for supported proxy trust values.
  * Rate limiting and security auditing now consistently use the configured client IP.

* **Bug Fixes**
  * Improved protection against forwarded-IP spoofing when proxy trust is disabled or limited.

* **Documentation**
  * Documented proxy trust configuration, security considerations, and throttling behavior.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3344, opened 2026-07-28
feat: restore file config, harden the launcher and dedupe the UI bootstrap

tada5hi:
Implements phases 0-2 of the runtime-topology plan: restores server-core's file-based configuration, rebuilds the `authup` launcher as a real process supervisor, deduplicates the two UI bootstrap stacks, and reconciles the deployment documentation with what the code actually does.

## Why

Three separate mechanisms had rotted unnoticed, because nothing covered them:

- The **published `authup` CLI crashed at load** (`ReferenceError: __dirname is not defined in ES module scope`) and — past that — spawned `server-core/dist/cli.mjs`, a path that stopped existing when server-core switched to an unbundled build in April.
- **Server-core's file config was parse-and-discard.** The CLI root command read and validated `authup.conf`, then threw the result away, while the app booted env-only. An invalid file crashed startup; a valid one was ignored.
- **`npm pack` silently drops nitro's symlinked module store**, so the published `@authup/client-web` tarball answered every SSR request with a 500 (`hookable does not provide HookableCore`).

All three reproduce **only from a packed artifact** — a workspace-dist run passes straight through every one of them, which is why this PR adds a pack-and-run smoke test.

## Changes

**server-core**
- File config now reaches the running app on every CLI subcommand. `--configDirectory` / `--configFile` are threaded into a `ConfigModule` slot on `createApplication`; environment variables still win over file values, pinned by tests. Migrations and healthcheck consume the same resolved config.
- `Component` gained `stop()` and `ComponentsModule` a `teardown()`, so the cron sweeps (oauth2-cleaner, event-cleaner) can no longer outlive teardown and fire against a destroyed DataSource.
- The `reset` command is removed — it ran setup + teardown and reset nothing.

**authup launcher**
- Rewritten as a deliberately dumb supervisor: ESM-safe paths, entrypoints resolved from each package's `bin` field, `spawn` with argv arrays and `stdio: 'inherit'` (no more shell strings or output munging), environment passthrough, SIGINT/SIGTERM forwarding, and exit with the first-failing child's code.
- `PORT`/`HOST` are always set per child. Children inherit the launcher's environment, so a single ambient `PORT` — a PaaS injects one, this repo's own Dockerfile sets `PORT=3000` — previously reached both children and the second died with `EADDRINUSE`.
- `migration` and `healthcheck` no longer boot the web UI; unknown commands fail with a message rather than a stack trace.
- The private config mini-framework is gone (it carried a dead `authorizeRedirectUrl` reference and a `publicUrl` → `apiUrl` copy-paste bug); dependencies pruned from twelve to six.
- New: unit tests, a workspace smoke test, a pack-and-run smoke test, and a `tests-launcher` CI job. The launcher's `check:types` now runs in the lint job.

**UI**
- `AAuthApp` (shared auth-page shell) and `AWorkflowDisabledNotice` moved into the kit; the four server-core SSR workflow pages and client-web's auth layout are now thin callers.
- The vuecs bootstrap **values** (icon collections, install options) are shared via `client-web-kit/core/vuecs`, while each app keeps its own install sequencing — the plugin ordering is load-bearing and stays untouched.
- Dead `apps/client-web/pages/authorize.vue` deleted (a pre-hosted-login leftover that rendered nothing).
- `API_URL_DEFAULT` in `@authup/kit` replaces four divergent defaults, including a `client-web-nuxt` fallback that pointed at a port nothing listens on.

**client-web packaging**
- `postbuild.mjs` materializes nitro's symlinked store before packing and drops the now-redundant store afterwards.

**Docs**
- `.agents/architecture.md` gains a "Deployment Topology & UI Boundary" section recording why the hosted auth pages belong to server-core (WebAuthn origin binding, the `prompt=none` ladder's first-party cookies, same-path GET-HTML/POST-JSON routes, mail deep links, headless deployments) and why client-web stays an ordinary OAuth2 RP.
- Configuration guides rebuilt from source: every key enumerated, file-only keys marked, the ghost `vault` page removed, and client-web's real runtime `NUXT_*` override names documented.

## Verification

- Full monorepo build green.
- Full test suite green — server-core 166 files / 1770 tests, client-web-kit 151, launcher 28.
- Both smoke variants green. The packed variant boots the installed `authup` bin from real tarballs with client-web SSR answering 200, which is what proves the packaging fix.
- The ambient-`PORT` regression was verified in both directions: the smoke test passes with the fix and fails without it.

## Summary by CodeRabbit

* **New Features**
  * Added an `authup` launcher that starts the web interface and server together with shared configuration.
  * Added configurable ports, hosts, API URLs, cookie domains, and configuration-file selection.
  * Added graceful shutdown, signal forwarding, health checks, and migration command support.
  * Added shared authentication-page layout and workflow-disabled notices.
* **Bug Fixes**
  * Improved shutdown cleanup for scheduled background tasks.
  * Prevented ambient port settings from overriding per-application bindings.
* **Documentation**
  * Expanded deployment, configuration, launcher, and topology guidance.
* **Tests**
  * Added launcher smoke tests and expanded configuration, routing, and lifecycle coverage.


tada5hi:
@coderabbitai pause

PR #3352, opened 2026-07-29
refactor: align the schema on typeorm's derived constraint names

tada5hi:
> **Reworked.** Now one PR: the renames, the mysql uuid widening, and the revert of #3408's entity pins, which are mutually exclusive with the renames. The migration is plain SQL again; the guarded-helper approach and its 400-line data module are gone.

> **Folded (2026-08-10).** The release window's other two unreleased migrations are merged in per the release convention (one file per dialect, earliest timestamp): `SessionTokenClientId` (#3421) and `EventSessionId` (#3422) now live inside `1785871780234-AlignSchemaWithEntityMetadata`. Both new columns are created directly under typeorm's derived constraint names, which also closes a gap the ordering had left open: `EventSessionId` sorted *after* the alignment migration, so its readable `IDX_auth_events_session_id` survived the chain and stayed pinned on the entity. The populated round-trip now exempts columns the reverted migration itself adds (their values cannot survive their own column's drop, by contract) and asserts the re-run restores them. Anyone who executed the pre-fold files must drop their dev database or fix its `migrations` table by hand.

## Why

The migration chain and the entity classes are two independent descriptions of the same schema, and only the chain is exercised against mysql/postgres: the test suites build their schema with `synchronize()` on sqlite. They had drifted, and nothing was watching.

Two causes, both from migrations that were hand-authored rather than generated (`1783325495597` beta.52, `1783769340000` beta.53, `1785871780234` unreleased):

1. **Constraint names.** 20 indexes, 9 foreign keys and 3 unique constraints created under readable names (`IDX_auth_events_actor_name`) where typeorm derives a table+column hash. Every generated migration before them used the hash, so the schema carried two naming regimes.
2. **uuid column widths, mysql only.** 15 columns declared `varchar(36)`. MySQL has no uuid type and `MysqlDriver.getColumnLength` only shortens to 36 for values typeorm generates itself, so a plain `@Column({ type: 'uuid' })` derives the generic `varchar(255)` every earlier table got. Postgres has a native uuid type and is unaffected.

Neither has a runtime effect: nothing resolves a constraint by name, and `varchar(36)` holds a uuid perfectly well. The damage is that `migration generate` reconciles against the model, so any new migration arrived with 32 renames and 15 `DROP COLUMN` plus re-`ADD` pairs attached. That is data loss which reads as routine in review, and it buries the real change. It is also how two wrong-target foreign keys (`auth_permissions.client_id`, `auth_roles.client_id`) survived long enough to need finding by hand.

## One naming regime, and it is typeorm's

221 constraints exist. 37 are hand-crafted, 184 are typeorm-derived:

| | total | hand-crafted | derived |
|---|---|---|---|
| FK | 107 | 9 | 98 |
| IDX | 56 | 20 | 36 |
| UQ | 21 | 3 | 18 |
| PK | 37 | 5 | 32 |

All-hand-crafted would mean renaming 184 and then inventing and pinning a name for every future foreign key and index, forever, with nothing enforcing it: a forgotten name silently falls back to a hash and the drift reopens. So the 37 move to the derived names and the entities declare none.

**Scope is 32, not 29 or 37.** The 29 that drift, plus the 3 `UQ_auth_*` which only avoided showing as drift because their entities pinned them (`@Unique('UQ_auth_keys_name_realm_id', ...)`), and which would otherwise survive as exactly the hand-maintained exception this removes. The 5 `PK_auth_*` are left alone: the schema builder never compares primary key constraint names, so they cannot drift, and renaming them would touch tables for a name nothing reads.

**#3408 is reverted here.** It pinned those names on the entities as interim scaffolding, which bought a clean postgres `migration generate` while this sat parked. Pins and renames cannot both ship: the result is the same drift, mirrored. So the six entity files go back to bare `@Index()` / `@JoinColumn({ name })`, and the three `@Unique('UQ_...')` names come off.

## Why widen mysql rather than keep 36

70 of the 85 plain-uuid columns are already `varchar(255)`, so widening the 15 makes them consistent rather than leaving outliers. The alternative is `length: 36` on the entity, which postgres rejects at `DataSource.initialize()` (`EntityMetadataValidator`: "does not support length property") and which depends on an unfinished upstream change. A dialect-conditional column type is not expressible either, since `migration generate` builds both dialects in one process off the same entity classes.

The cost of 255 is narrower than it looks: mysql `VARCHAR` stores the actual bytes, so per-row storage and index entries are effectively unchanged. The real cost is MEMORY temp tables and sort buffers allocating the declared maximum.

`MODIFY COLUMN` widens in place. A **generated** migration would have emitted `DROP COLUMN` plus re-`ADD` instead, because `MysqlQueryRunner.changeColumn` drops and re-adds a column whose length differs, which empties it on a populated table and fails outright on a foreign key column.

## Plain SQL, not helpers

The first cut imported guarded rename/widen helpers plus a 400-line data module, which made this migration a program rather than a list of statements. Both are gone. Every other migration in the chain is a sequence of `queryRunner.query(...)` strings, and the idempotence the helpers bought is a property none of them has: the `migrations` table is what prevents a re-run.

- postgres: `ALTER INDEX ... RENAME TO` x20, `ALTER TABLE ... RENAME CONSTRAINT` x12.
- mysql: `RENAME INDEX` x23 (mysql stores a unique constraint as a unique index), foreign key drop + add x9 (no `RENAME CONSTRAINT` on mysql), `MODIFY COLUMN` x15, wrapped in `SET FOREIGN_KEY_CHECKS`.

**The widening runs first and the order is load-bearing.** MySQL commits each DDL statement regardless of the surrounding transaction, so a boot killed partway leaves this applied but unrecorded and the whole `up` re-runs. Widening is re-runnable (a `MODIFY COLUMN` onto the width a column already has succeeds); a rename is not (`RENAME INDEX` on an already-renamed index fails). Running the minutes-long phase first and the instant one last means the phase that can realistically be interrupted is the one that survives a retry. Verified by applying a prefix of `up` and re-running it whole: it succeeds in this order and fails on `IDX_auth_consents_sub` in the other.

**The checks are restored in a `finally`.** `FOREIGN_KEY_CHECKS` is per session and the connection returns to the pool when a migration aborts, so a failed statement would otherwise hand the next caller a connection that accepts orphaned rows.

**Each table is widened in one `ALTER`.** Every one of these columns crosses the utf8mb4 varchar length-prefix boundary (`varchar(36)` is 144 bytes and carries a 1-byte prefix, `varchar(255)` is 1020 and needs 2), so InnoDB rejects `ALGORITHM=INPLACE` and rebuilds the table. One statement per column copied `auth_events` and `auth_session_tokens` four times each; grouped, the two highest-volume tables are rebuilt once, 6 rebuilds instead of 15.

The checks are disabled for the duration because every constraint is re-added with the definition it already enforced, so re-validating would only buy a table scan plus a failure mode for rows some past import inserted with the checks off. It is also what lets a foreign key column be widened while its constraint is in place.

The upstream helpers this used to consume (`renameIndex`, `renameForeignKey`, `changeColumnType`, `withForeignKeyChecksDisabled`) were still filed and shipped from this work as tada5hi/typeorm-extension#1420 and #1424; `getSchemaDrift` from the same line still backs the drift gate.

## Verification

mysql 9 and postgres 18.

| Check | Result |
|---|---|
| `migration generate` | emits **no file** on either dialect |
| CLI round-trip: run, revert x17, run | clean on both |
| Drift gate against the migrated schema | 0 on both |
| **Populated** round-trip (`test:migration-latest`) | **11/11 on both** |
| sqlite `synchronize()`: create, then reopen | 0 drift and 0 drift |
| Full server-core suite | 2020/2020 |
| eslint, `check:types` | clean |

The populated run is the one that matters, since CI's round-trip runs on an empty schema and cannot surface a column rewrite losing values. With 234 `auth_events` plus sessions, session tokens, consents, trust anchors and authenticators seeded, it asserts row counts **and column values** unchanged after revert and after re-apply, that the foreign keys still reject an orphan insert and still cascade, and that the application boots and issues a token against the migrated schema.

## Also in here

- **`npm run test:schema-drift`** (`scripts/assert-schema-drift.mjs`), wired as a fourth step of `tests-migrations`. Asserts zero drift on **both** dialects, with no mysql carve-out. This is what stops the split reopening.
- **`npm run test:migration-latest`** (`scripts/verify-latest-migration.mjs`), the populated round-trip above. Not in CI, since it boots the app and drops the target database. **Fixed on the way**: it seeded against the `web` client that plan 082 removed, so it had been failing since then.
- **Conventions.** "Never hand-write a constraint name" replaces "pin the names a migration spelled out". A rule with no list cannot rot. The mysql uuid width trap is written down alongside it, including why `length: 36` is not the escape.
- **`.agents/references/typeorm.md`** updated for what is and is not consumed now.

## Notes for review

- **MariaDB is out of scope**, and this PR writes that down (`.agents/conventions.md`). There is no `migrations/mariadb/` directory and the config validator accepts only mysql / postgres / better-sqlite3. Pointing `type: 'mysql'` at a MariaDB server mostly works, so the two concrete divergences are recorded: it refuses to change a column an existing foreign key references even with the checks off, which aborts this migration, and MariaDB 11.4+ defaults to `utf8mb4_uca1400_ai_ci` where MySQL 9 uses `utf8mb4_0900_ai_ci`, so every string column reads as drift. Accommodating it was tried and reverted: it costs the retry property above and buys nothing supported.
- **Merge after the pending release.** Nothing in the current window depends on this.
- **Operational cost:** the mysql widening rebuilds 6 tables, so the boot applying it stalls on a large `auth_events` or `auth_session_tokens`. That cost is the same whenever it ships.
- Two suite failures appeared on the first full run (`session.spec.ts`, `provisioning.spec.ts`). Both pass in isolation and a re-run of the identical tree is 2020/2020: the documented parallel-suite flake, not this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration compatibility across MySQL and PostgreSQL.
  * Updated schema handling to prevent migration-related drift and preserve relationships, constraints, and data integrity.
  * Adjusted UUID storage compatibility for MySQL databases.

* **Tests**
  * Added automated checks for schema consistency, migration rollback and reapplication, populated data preservation, foreign-key behavior, application startup, and login verification.

* **Documentation**
  * Expanded migration guidance, testing instructions, database compatibility notes, and troubleshooting references.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3354, opened 2026-07-29
fix: provision client scopes into the client_scope junction

### LINKED ISSUE #3347: Per-realm `web` client provisions scopes into the dead `Client.scope` column, not `client_scope` rows
## Summary

`WebClientProvisioner` writes the per-realm `web` client's intended scopes into the **dead `auth_clients.scope` column**, but creates no `auth_client_scopes` junction rows — which is where `/authorize` actually reads scopes from. The declared intent (`global openid`) is therefore never enforced.

It is currently masked by the `global` bypass in the code-request verifier, so this is **latent, not live-breaking**.

## Detail

`buildWebClientAttributes` (`apps/server-core/src/core/entities/client/web-client.ts:53`) sets:

```ts
scope: `${ScopeName.GLOBAL} ${ScopeName.OPEN_ID}`,
```

But scope authorization at `/authorize` resolves exclusively through the junction table — `OAuth2ScopeRepository.findByClientId` (`app/modules/oauth2/repositories/scope/repository.ts`) reads `auth_client_scopes` rows and their `scope` relation. The `Client.scope` column is never read on any production path.

I verified the latter by renaming `Client.scope` in `packages/core-kit/src/domains/client/entity.ts` and compiling `apps/server-core`. All 29 resulting errors are structural assignability noise (`ClientEntity` no longer satisfying `Client`). The only real property accesses are:

| Site | Role |
|---|---|
| `core/entities/client/web-client.ts:53` | **writes** the value |
| `core/entities/client/schema.ts` | rapiq projection allow-list |
| `packages/core-http-kit/src/domains/entities/client/types.ts` | payload type |
| 4 spec files | assert the written value |

Zero production reads.

## Impact

In `OAuth2AuthorizationCodeRequestVerifier` (`core/oauth2/authorization/code-request/verifier/module.ts:97-107`):

```ts
const scopes = await this.scopeRepository.findByClientId(client.id);
const scopeNames = scopes.map((scope) => scope.name);   // [] for the web client
if (data.scope) {
    if (
        !hasOAuth2Scopes(scopeNames, data.scope) &&
        !hasOAuth2Scopes(data.scope, ScopeName.GLOBAL)
    ) {
        throw OAuth2ScopeError.insufficient();
    }
} else {
    data.scope = scopeNames.join(' ');   // '' for the web client
}
```

With `scopeNames = []`:

| Request | Outcome |
|---|---|
| `scope=global openid` | **passes** — via the `global` bypass on the second leg |
| `scope=openid` (no `global`) | **`insufficient_scope`** — `hasOAuth2Scopes([], 'openid')` is `false` (empty granted set short-circuits to `false`), and `'openid'` does not contain `global` |
| no `scope` param | code issued with `scope=''` |

Authup's own clients always send `global openid` — `buildAuthorizeURL` (`packages/client-web-kit/src/core/oauth2/authorization-request.ts:88`) defaults to it and `apps/client-web/pages/login/index.vue:100` passes it explicitly — so both take the bypass. A **non-kit RP** performing a standard OIDC `scope=openid` authorization request against a realm's built-in `web` client gets `insufficient_scope`.

## Contributing factor

`ClientProvisioningRelations` (`core/provisioning/entities/client/types.ts`) exposes `permissions` / `realmPermissions` / `globalPermissions` / `roles` / `realmRoles` / `globalRoles` — but **no `scopes`**. So there is presently no provisioning path to `auth_client_scopes` for any client, not just `web`.

## Possible resolutions

1. Have `WebClientProvisioner.ensureForRealm` upsert `auth_client_scopes` rows for `global` + `openid` (idempotent, same MERGE posture as the client attributes), and add a `scopes` relation to `ClientProvisioningRelations` so file-provisioned clients can declare them too.
2. Separately, decide the fate of the `Client.scope` column: it is write-only dead weight and a standing trap — it reads like the authorization source but is not. Either drop it (migration + core-kit/core-http-kit/schema/validator cleanup) or make it the fallback the verifier consults when the junction is empty. Dropping it seems right given `client_scope` is the modelled relation.

Whether the `global` bypass itself should stay is a separate question and deliberately out of scope here.

## Reproduction

Against a fresh instance, for any realm's built-in `web` client:

```
GET /authorize?client_id=<web-client-uuid>&response_type=code&scope=openid&redirect_uri=<registered>&state=x&code_challenge=<c>&code_challenge_method=S256
```

→ `insufficient_scope`, where the provisioned `scope` column claims `global openid`.

tada5hi:
Closes #3347 (provisioning half; the dead-column removal is deliberately out of scope, see below).

## The bug

`buildWebClientAttributes` wrote the per-realm `web` client's intended scopes into `auth_clients.scope`, but `/authorize` resolves scopes exclusively through the `auth_client_scopes` junction (`OAuth2ScopeRepository.findByClientId`). No junction rows were ever created, so the declared `global openid` was never enforced.

With an empty granted set the code-request verifier behaves like this:

| Request | Outcome |
|---|---|
| `scope=global openid` | passes, via the `global` bypass |
| `scope=openid` | `insufficient_scope` |
| no `scope` param | code issued with `scope=''` |

Authup's own clients always send `global openid`, so the gap only surfaced for a non-kit RP performing a standard OIDC `scope=openid` request against a realm's built-in `web` client.

## Changes

- **`WebClientProvisioner.ensureForRealm`** binds `global` + `openid` as `auth_client_scopes` rows, driven by a shared `WEB_CLIENT_SCOPE_NAMES` so the descriptive column cannot drift from what is actually granted. The binding runs outside the attribute dirty-check: an instance provisioned before the rows existed carries an up-to-date client with no scopes at all, and must be backfilled on the next boot. Additive like the attribute merge, so a scope an admin bound by hand survives.
- **`ClientProvisioningRelations` gains `globalScopes` / `realmScopes`.** There was no provisioning path to `auth_client_scopes` for any client, not just `web` (the contributing factor named in the issue).
- **Realm-level scopes now synchronize before the realm's clients.** Otherwise a `realmScopes` entry could not reference a scope declared in the same realm block. Scopes are leaf entities with no relations of their own, so moving them first is safe.
- **`ProvisioningEntityResolver` takes a `clientScoped` option.** `auth_scopes` has no `clientId` column, so the always-pinned client predicate would not compile against it.
- **An existing client named `web` is taken over rather than skipped.** The name is reserved (`CLIENT_RESERVED_NAMES`), so a non-built-in row predates the reservation and must not shadow the realm's login client. Note the consequence: such a client is converted to public / `builtIn` with the trusted-origin redirect patterns, so an integration still using it with a secret stops working.

## Verification

- `authorize.spec.ts` gains the issue's exact reproduction (`scope=openid` alone against the master realm's `web` client). Confirmed red with the scope binding disabled, green with it.
- Unit coverage for junction binding, backfill of an already-current client, idempotency, hand-bound scope preservation, unprovisioned-scope skip, and the takeover.
- `provisioning.spec.ts` covers both file-declared client scopes and the `web` client end to end.
- Full server-core suite green (1787 tests), plus build, `check:types` and eslint.

## Out of scope

The dead columns (`Client.scope`, `baseUrl`, `rootUrl`) are untouched. Dropping them needs a generated migration on top of #3352, and the decision applies to all three at once. `Client.scope` is now documented in code and in `.agents/architecture.md` as descriptive only, with the junction named as the authorization source.

## Summary by CodeRabbit

* **New Features**
  * Clients can now be provisioned with global and realm-specific scopes.
  * Built-in web clients automatically receive and maintain their required scope assignments.
  * Existing clients missing scope assignments are updated automatically.

* **Bug Fixes**
  * Corrected authorization for the web client when requesting OpenID scope.
  * Preserved manually assigned scopes and skipped unavailable scopes safely.

* **Documentation**
  * Added provisioning guidance for configuring client scopes.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -784,11 +784,11 @@ The provisioning system declaratively synchronizes entities (permissions, roles,
 
 ### Layers
 
-- **core/provisioning/entities/**: Provisioning entity types and validators (what can be provisioned)
+- **core/provisioning/entities/**: Provisioning entity types and validators (what can be provisioned). A client declares `permissions` / `roles` (define new client-scoped ones) plus `globalPermissions` / `realmPermissions` / `globalRoles` / `realmRoles` / `globalScopes` / `realmScopes` (assign existing). Every assign list lands in the matching junction table
 - **core/provisioning/strategy/**: Strategy types (`createOnly`, `merge`, `replace`, `absent`) and normalization
 - **core/provisioning/synchronizer/**: Business logic that applies strategies and manages relations
-  - `entity-resolver.ts`: `ProvisioningEntityResolver<T>` — resolves Permission/Role entities by name with wildcard support and scope filtering (global, realm, client)
-  - `junction-synchronizer.ts`: `ProvisioningJunctionSynchronizer<T>` — ensures junction entries (e.g. RolePermission, UserRole) exist between owner and target entities
+  - `entity-resolver.ts`: `ProvisioningEntityResolver<T>` — resolves Permission/Role/Scope entities by name with wildcard support and scope filtering (global, realm, client). The client dimension is always pinned so a client-ownable entity never resolves another owner's rows; entities without one (scope) opt out via `{ clientScoped: false }`, since the predicate would not compile against their table
+  - `junction-synchronizer.ts`: `ProvisioningJunctionSynchronizer<T>` — ensures junction entries (e.g. RolePermission, UserRole, ClientScope) exist between owner and target entities
   - `{entity}/module.ts`: Per-entity synchronizer composing resolver + junction helpers
 - **app/modules/provisioning/sources/**: Data sources that produce `RootProvisioningEntity`
   - `default/`: Built-in defaults (system policies, admin user, system client, all permissions/scopes)
@@ -827,7 +827,9 @@ dropped from the validator output) are validated via
 `ProvisionerModule` runs (1) `GraphProvisioningSynchronizer`, (2) backfill via `assignDefaultPolicy` (config-gated, deprecated).
 
 `GraphProvisioningSynchronizer` processes in order: policies → permissions → roles → scopes → realms.
-`RealmProvisioningSynchronizer` processes per realm: clients → permissions → roles → users → scopes.
+`RealmProvisioningSynchronizer` processes per realm: scopes → clients → permissions → roles → users.
+Scopes run first because they are leaf entities carrying no relations of their
+own, and a client in the same realm block may bind them via `realmScopes`.
 
 ### Per-Realm Public `web` Client
 
@@ -842,8 +844,17 @@ endpoint — the `/authorize` verifier already resolves clients via
   `authMethod: 'none'`, `tokenBindingMethod: 'none'`, `builtIn: true`, `active: true`,
   `grantTypes: 'authorization_code refresh_token'` (an enforced allowlist —
   see *Per-client grant allowlist* under the token-endpoint section),
-  `scope: 'global openid'`, `redirectUri` = one `<origin>/**` wildcard per
-  trusted app origin (matched by `isSimpleMatch`).
+  `redirectUri` = one `<origin>/**` wildcard per trusted app origin (matched
+  by `isSimpleMatch`).
+- **Scopes** (`WEB_CLIENT_SCOPE_NAMES` = `global` + `openid`) are bound as
+  `auth_client_scopes` rows. That junction is the only source `/authorize`
+  reads scopes from (`OAuth2ScopeRepository.findByClientId`). The `Client.scope` column carries
+  the same list but is descriptive only: nothing on any production path reads
+  it. Writing the column alone left the client with an empty granted set, so a
+  standard OIDC `scope=openid` request failed with `insufficient_scope` while
+  only requests carrying `global` passed via the verifier's bypass (#3347).
+  The junction rows are additive: a scope an admin bound by hand survives the
+  next boot.
 - **App origins** come from `getAppOrigins(config)` = publicUrl's origin +
   `config.trustedOrigins` merged verbatim. A `trustedOrigins` entry may carry
   an http(s) scheme (contributes exactly that origin; other protocols are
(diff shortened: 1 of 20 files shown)


tada5hi:
## Self-review notes

Two things came out of auditing the branch, plus three observations worth recording.

### Fixed: the takeover left a plaintext secret at rest

The takeover switches the client to `authMethod: none`, so a secret it carried as a confidential client can never authenticate it again (`OAuth2ClientAuthenticator` rejects a supplied secret for `none`, and `ClientAuthenticator` only accepts `secret` clients). It was still sitting in the column. Now cleared.

The first attempt at this silently did nothing: `secret` is a `select: false` column, so the loaded entity carries no value for it and typeorm's changed-column diff never emitted the null. It only surfaced because the assertion runs against a real datasource in `provisioning.spec.ts`; the repository fake merges by `Object.assign`, so a fake-only test would have passed while the column kept its value. The fix re-reads the row with the secret, confined to the takeover branch. The secret deliberately stays out of `buildWebClientAttributes`, since an undefined-vs-null compare would mark every web client dirty on every boot.

### Corrected: the "no `scope` param" row in the issue is not reachable

The impact table in #3347 lists a third case, "no `scope` param -> code issued with `scope=''`". That cannot happen: `OAuth2AuthorizationCodeRequestValidator` mounts `scope` as required (`z.string().min(3).max(512)`), and all three verifier call sites (`GET /authorize`, `POST /authorize`, the identity-provider callback) run the validator first. A scope-less request is a 400 before the verifier sees it.

That matters for reviewing this PR, because it means the verifier's `else { data.scope = scopeNames.join(' ') }` branch is dead code. Had it been reachable, this change would have widened it: a scope-less authorize against the web client would have started defaulting to `global openid` instead of `''`, and `global` is what makes the identity policy data available in `RequestPermissionEvaluator`. As it stands there is no such path, so the only behavioral delta is that in-grant scope requests stop failing.

### Observations, deliberately not addressed

- **Concurrent first boot can collide on the unique index.** `ensureScopes` is find-then-insert against `@Unique(['clientId','scopeId'])`, so two replicas booting simultaneously against an unprovisioned database can both insert and one crashes. This is the shape of the entire provisioner (the client row itself, `ProvisioningJunctionSynchronizer`, every permission and role junction), not something this PR introduces, and the window is the first boot per realm only. Guarding just this path would be arbitrary, and `isUniqueConstraintDatabaseError` lives in the adapter layer, which `core/` must not import. Worth a separate pass over the provisioner as a whole.
- **Four queries per realm per boot.** Two scope lookups and two junction lookups. Memoizing the global scope ids across realms was rejected: the same provisioner instance is long-lived behind `RealmService`, so a cached id would outlive a scope an admin deleted and recreated, and the stale id then fails as an FK violation at insert time. The queries are cheap and correct.
- **`ensureScopes` duplicates what `ClientProvisioningSynchronizer` now does for `globalScopes`.** The web client is deliberately not routed through the client synchronizer, because `RealmService.save()` provisions it at runtime and would otherwise have to construct the permission and role repositories and synchronizers too. The duplication is the existing split, kept minimal.

Failure handling is unchanged and self-healing: `RealmService.save()` already catches and logs a provisioning failure, and because the scope binding runs outside the attribute dirty-check, the next startup pass reconciles anything a runtime failure left behind.

PR #3358, opened 2026-07-29
feat: hand server-rendered data to the client instead of fetching it twice

### LINKED ISSUE #2773: Feature: UI Async Data
use `useAsnycData` composable for api requests.

https://nuxt.com/docs/4.x/guide/recipes/custom-usefetch#custom-usefetchuseasyncdata

tada5hi:
Closes #2773.

The UI fetched everything twice. Detail pages awaited the record fetch directly in `setup()`, which runs during SSR and again on hydration, and entity collections fired their load from a detached microtask, so the server render started a request whose result was discarded when the render flushed and the browser then fetched the same rows again.

## What changed

**A hydration seam in the kit.** `install({ hydrationStore })` takes a get/set/delete bucket the host backs with its hydration payload (`core/hydration/`). Without one the kit performs no server-side load at all, because the response could not reach the client anyway. Nuxt backs it with `nuxtApp.payload.data`, the same bucket `useAsyncData` transports through; the embedded SSR app backs it with a new `HydrationPayload.hydration` field, which works because the payload is serialized after `renderToString`.

**Collections** load inside `onServerPrefetch`, so their rows reach the markup, and record the result under the identity of the request they would send (entity type plus serialized query). Both sides derive that key the same way, and a miss simply loads as before. A load that failed is deliberately not recorded: an adopted snapshot suppresses the client's own load, so handing over an empty one would strand the page on it.

**Detail pages** fetch through `useAsyncData` with `deep: true` (Nuxt 4 returns a `shallowRef` by default, and the pages update their entity in place via `extendObject`).

**Translations and permission verdicts.** SSR-rendering real content exposed a second problem: `useTranslation` and `usePermissionCheck` resolve asynchronously, so the markup said `Name` where the first client render said `authupField.name`, and a control the markup showed as enabled started from a fail-closed `false`. Vue reported five hydration mismatches on `/users`. Both now record their resolved value through the same seam and seed the first client render. Permission keys carry the actor, so an account switch cannot adopt the previous one's verdict, and checks carrying a `PolicyData` bag are not keyed at all and keep their fail-closed default.

The translation half is a workaround for ilingo having no synchronous read path even over an in-memory store, which is what forces `@ilingo/vue` into `computedAsync`. Filed upstream as tada5hi/ilingo#988; once that lands the payload entries become unnecessary.

## Verification

Driven in a headless browser over CDP, `/users` as an authenticated admin:

| | client `GET /users` | rows in SSR HTML | hydration warnings |
|---|---|---|---|
| before | 1 | 0 | 0 |
| after | 0 | 1 | 0 |

The SSR `/authorize` page reports no hydration warnings either. Payload cost on `/users` is 25 entries (22 translations, 2 permission verdicts, 1 collection snapshot), roughly 865 bytes on top of the collection snapshot itself.

Suites: kit 177/177, server-core 1795/1795 (including a new case asserting the `/authorize` identity-provider list is server-rendered and present in the payload), `vue-tsc` clean for client-web.

## Notes

- The embedded SSR app now also passes `isServer`, which it never did: without it the authentication hook scheduled a token-refresh timer on every server render.
- Collection snapshots are consumed on read, so a later client-side visit fetches fresh rows. Translation and permission entries are not, since the same key is shared by every component asking the same question.
- Other fire-and-forget SSR fetches in the kit (the authorize page's MFA and consent probes) are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added server-rendered data handoff so initial pages display collection results immediately and avoid duplicate client requests.
  * Improved first-render consistency for translations and permission checks during hydration.
  * Expanded SSR support across client, event, identity provider, key, permission, policy, realm, role, scope, trust anchor, and user detail pages.

* **Bug Fixes**
  * Improved failed detail-page loading behavior by returning users to the relevant listing page.


tada5hi:
@coderabbitai pause

PR #3363, opened 2026-07-30
fix: enforce the permission guard on entity index detail links

### LINKED ISSUE #3071: Replace :disabled on NuxtLink with proper guards in entity index pages

**Reported in:** [PR #3069 review comment](https://github.com/authup/authup/pull/3069#discussion_r3279789823) by @coderabbitai
**Files:**
- `apps/client-web/pages/robots/index/index.vue#111-117`
- `apps/client-web/pages/roles/index/index.vue#142-148`
- `apps/client-web/pages/scopes/index/index.vue#126-132`
- `apps/client-web/pages/users/index/index.vue#113-119`

## Description

The entity index pages use `<NuxtLink :disabled="!hasEditPermission">` to guard the per-row "edit" button. `NuxtLink` does not honour a `disabled` prop — the attribute is rendered onto the underlying anchor but does **not** block navigation, and Bootstrap's `.disabled` visual class is not applied either. Users without edit permission see an active-looking button, click it, and only get rejected at the edit page by its own permission check.

This is pre-existing — the same pattern lives in `master` and was untouched by the vuecs-1.x migration PR (#3069). The migration only swapped `<BTable>` for `<ATable>`. Tracking separately to keep PR scope tight.

## Suggested fix

Replace the four `NuxtLink :disabled` instances with one of:

1. Conditionally render — `<NuxtLink v-if="hasEditPermission" ...>` and a disabled-looking `<span>` otherwise.
2. Add `aria-disabled`, `tabindex=-1`, and `@click.prevent` when not permitted, plus a `.disabled` CSS class for the visual cue.

Approach 1 is cleaner; approach 2 keeps the markup more uniform.

tada5hi:
Closes #3071.

## Problem

The pattern moved on since the issue was filed (`<NuxtLink :disabled>` became
`<VCButton :as="NuxtLink" :disabled>`), but the defect is the same and a bit
worse than reported:

- `VCButton` declares `disabled` as its **own prop**, so it never reaches the
  `as` target. A non-native target receives only `aria-disabled="true"` and
  keeps its own click behaviour, so the link still navigates.
- There was **no visual cue either**: theme-tailwind styles the disabled state
  through Tailwind's `disabled:` variant, which matches the `:disabled`
  pseudo-class, i.e. form elements. An `<a>` never matches it.

So an actor without the update permission saw a fully enabled-looking button,
clicked it, and was only rejected by the detail page.

## Fix

Across all 11 entity index pages (`clients`, `roles`, `realms`, `permissions`,
`users`, `policies`, `trust-anchors`, `keys`, `events`, `scopes`,
`identity-providers`):

- `resolveComponent('NuxtLink')` is replaced by `<VCLink>` from `@vuecs/link`,
  which resolves NuxtLink / RouterLink / a plain anchor at render time.
- The route target is **withheld** when the actor may not edit:
  ``:to="hasEditPermission ? `/roles/${row.id}` : undefined"``. `VCLink` then
  renders an href-less `<a>` whose click is `preventDefault`ed and which cannot
  be tab-focused.
- `clientWebKitTheme()` appends `aria-disabled:cursor-not-allowed
  aria-disabled:opacity-60` to the button root, so a disabled button-link looks
  disabled again. Both UI apps register the kit theme.

## Also fixed

`AEntityDelete` carried the same defect on its `link` / `dropDownItem` render
paths: `disabled` was handed to an element with no native disabled semantics and
was never checked in the click handler, so a disabled destructive action still
fired. No call site uses those element types today, but it is public kit API.
Guarded, with a regression spec that fails without the guard.

## Upstream

Filed as tada5hi/vuecs#1699 (VCButton should guard activation for a non-native
`as` target; theme-tailwind should mirror the disabled cue onto
`aria-disabled`). Both workarounds here reference it as their removal trigger.

## Verification

- Generated CSS in the `apps/client-web` build contains
  `aria-disabled\:opacity-60[aria-disabled=true]{opacity:.6}` and the cursor
  rule.
- A vue-router mount of both states: permitted renders `href="/roles/abc"` and
  navigates on click; disabled renders `aria-disabled="true"`, no `href`, and
  the route does not change.
- `npm run build -w apps/client-web` (includes typecheck) passes; kit build and
  its 181 tests pass; eslint clean on every changed file.
- Not exercised in a live browser against a running backend.

## Summary by CodeRabbit

* **New Features**
  * Standardized table action links across entity pages for consistent navigation and permission handling.
  * Disabled or unauthorized actions no longer expose clickable navigation targets.
  * Improved disabled-state styling for button-styled links.

* **Bug Fixes**
  * Prevented delete actions from executing when busy or disabled.
  * Added coverage to verify enabled and disabled delete behavior.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3385, opened 2026-08-03
feat: theme the served consoles from a mounted directory

### LINKED ISSUE #3376: server-core: assert the auth console render-contract version at boot
## Context

Since #3375, the hosted auth workflow UI ships as `@authup/client-auth-console`. server-core resolves the package at runtime, loads its built `dist/server/server.js`, and calls the exported `render()`. The boundary is typed by the package's `src/contract.ts` (`HydrationPayload`, `RenderContext`, `RenderResult`, `RenderFunction`), which server-core imports type-only. Substituting a package that fulfills the contract is the supported way to replace the hosted login/consent UI (the Keycloak login-theme analog).

Types only protect this boundary at compile time. At runtime, server-core calls whatever function the resolved bundle exports. Today that is safe because server-core and the auth console live in one monorepo with linked versions, so the two sides cannot drift.

## Problem

Once third-party auth console packages exist, a server-core upgrade can change the render contract (a new required payload field, a changed return shape) while an operator's custom package still implements the old one. That mismatch would surface as a confusing per-request error on `/authorize` (or silently wrong pages), not as an actionable operator message.

## Suggested fix

- The package exports a runtime constant alongside `render()` in the server bundle, e.g. `export const contractVersion = 1`, bumped whenever `src/contract.ts` changes incompatibly.
- server-core asserts the version when it first loads the bundle (or at boot) and fails loud with an actionable message naming the required and found versions.
- A bundle without the export counts as version 1 during a transition window, then fails closed.

## Trigger

Deliberately deferred from the extraction (plan 083 decision): while the in-repo package is the only consumer, the assert would compare a constant against itself. Implement it when the first external/custom UI package becomes a real use case, or before documenting the substitution seam publicly.

tada5hi:
Operators can rebrand the auth console and the account console by mounting a directory. No image build, no rebuild of authup.

```
/etc/authup/theme/
  theme.json     tokens, title, favicon, logo, stylesheet
  assets/        the only directory served over HTTP, at /theme
  fragments/     opt-in head markup
```

```bash
docker run -e THEME_DIRECTORY_PATH=/etc/authup/theme \
           -v /srv/authup/theme:/etc/authup/theme:ro authup/authup
```

Four independently reviewable commits.

## What is reachable, and what is not

Both consoles are compiled Vue bundles, so there is no request-time template engine. Colours, surfaces, spacing, radii, fonts, logo, favicon and title are reachable; markup, UI copy and icons are not. Every colour and geometric proportion already resolves through a live `var()` chain, so a token map reaches essentially all of it.

Keycloak splits the same way. Its `.ftl` half needs a per-request template compile and cannot transfer; its asset/data half does. Keycloak hit this exact wall with its own React account and admin consoles and answered with published packages plus a scaffolder, which is what commit 4 turns into a first-class option here.

## Notable decisions

**Tokens go into their own `@layer`, never an inline `style` attribute.** A layer name absent from the bundle's `@layer` statement is appended last, so it beats `@layer authup` and `@layer base` without `!important`, while `.dark` still wins by source order. An inline attribute would also win, but permanently: the colour-mode switcher flips that class client-side with no reload, which would freeze the brand at whatever the cookie said at render time.

**Injection happens at `</head>`, not at a new template marker,** so theming also applies to console packages built before this existed.

**Token names are validated by grammar, not against a closed list.** A list would have to live in server-core, which cannot depend on `client-web-kit-theme`, so nothing could bind the two and it would rot in both directions.

**`/theme` is hand-written rather than `@routup/assets`.** That plugin's lookup walks path suffixes, so `/theme/anything/logo.svg` would serve `/logo.svg`; it probes `.html`/`index.html`; and it performs no realpath containment check. Symlinks must be followed since a Kubernetes ConfigMap volume is a symlink farm, so containment is realpath-per-request against the mount root instead. Content types are pinned from a map the extension allowlist is derived from, and SVG is served sandboxed because it is active content.

**Failure posture is deliberately asymmetric.** An invalid manifest fails the boot with the file path and every issue. A manifest that becomes invalid after boot keeps the last good value and logs. Provisioning seeds authorization data; a theme is decoration, and a broken logo must never take down an identity provider's login page.

**The default is not under `writableDirectoryPath`.** That directory is process-writable, and pairing it with content injected into the login page would turn any write primitive landing there into persistent branding control on the IdP origin.

## Trust boundary

The theme directory is operator trust, equal to `authup.conf`. Arbitrary CSS on `/authorize` can overlay or relabel the consent buttons, and `frame-ancestors 'none'` stops framing but not same-document overlay. So this is filesystem-only, and any future per-realm rung may carry data only (a theme name, or a token map through the same validator), never a stylesheet and never markup. Head fragments are off by default, head-only, and passed through verbatim: a partial sanitizer would be worse than none because it invites treating them as untrusted-safe.

## Scope note

Commit 3 is the one part that does not apply to already-published console packages. The kit theme CSS ships raw and is inlined by each console's Tailwind build, so the logo tokens take effect once both consoles are rebuilt. Commits 1, 2 and 4 work against the currently published packages.

## Verification

- server-core: 177 files, 1949 tests, 0 failures
- client-web-kit: 28 files, 194 tests
- 118 new tests covering the token grammar matrix, cascade ordering, `$'` non-expansion, traversal and `%2e%2e`, symlink containment, the suffix-walk and index-probe cases, the fragment-off path, and the contract asserts
- Verified the logo tokens reach both consoles' built CSS and that the real auth console bundle exports `CONTRACT_VERSION`
- `check:types` and eslint clean

Closes #3376.

## Summary by CodeRabbit

* **New Features**
  * Added configurable branding for authentication and account consoles, including logos, colors, titles, favicons, stylesheets, and optional head content.
  * Added live theme updates and secure theme asset serving.
  * Added support for custom authentication and account console packages with startup validation.
* **Documentation**
  * Added theming deployment guidance and configuration examples.
  * Updated Docker volume paths in deployment examples.
* **Bug Fixes**
  * Improved themed page caching and asset security, including traversal and symlink protection.


tada5hi:
@coderabbitiai pause

tada5hi:
@coderabbitai pause

PR #3397, opened 2026-08-04
fix!: redirect-pattern matching, plus fixes from the beta.58 release audit

tada5hi:
Fixes found while reviewing `v1.0.0-beta.58..master`. The first one should
land before the release.

## Auth-code exfiltration via a host wildcard

`isSimpleMatch` treats `/` as its only boundary, but a URL authority also ends
at `?`, `#` and `\`. A `*` in the host of a registered pattern absorbs one of
those, so the pattern's remaining host literal lands in the query or fragment
of a foreign origin:

```
pattern https://*.example.com/**          before   after
  https://evil.test?.example.com/cb       match  → no
  https://evil.test#.example.com/cb       match  → no
  https://evil.test\.example.com/cb       match  → no
  https://user@evil.test#.example.com/cb  match  → no
  https://evil.test/.example.com/cb       no       no    (control: `/` blocks)
```

Running the delivery code from `AuthorizeController.confirm` on those inputs
puts the code in the query string on `evil.test`:

```
https://evil.test/?state=x&code=SECRET_CODE#.example.com/cb
https://evil.test/.example.com/cb?state=x&code=SECRET_CODE
```

The per-realm console clients are public, `builtIn` (auto-consent, no screen)
and `global` scoped, and the attacker chooses the PKCE challenge, so one click
by a signed-in admin yields a full-permission token. `isValidPostLogoutRedirect`
has the same shape (it parses the URL for a protocol check, then matches the
raw string), giving a server-issued open redirect carrying `state`.

Reachable via config: `new URL('https://*.example.com').origin` returns the
wildcard verbatim, so `expandToOrigins` → `trustedOrigins` →
`buildSystemClientAttributes` writes `https://*.example.com/**` into
`redirectUri` **and** `postLogoutRedirectUri` in every realm.

Introduced by #3396, which is what made host wildcards work at all (all four
variants matched nothing before) and which also documented them as safe.

**The fix is the pattern already in this repo.** `resolveAccountConsoleRef`
canonicalizes through `URL` before matching, which is why the account console
is the one unaffected consumer. That is now `isSimpleURLMatch` in
`@authup/kit`, and all three consumers use it, so the string that gets
authorized is the string the browser navigates to. Case, the default port,
userinfo and dot segments resolve in the same step, which also closes
path-scoped patterns being walked out of with `..`.

Both sides are normalized, not just the candidate. Normalizing one side makes
the comparison asymmetric, and a stored `https://app.example.com:443/**` or
`https://APP.example.com/**` would stop matching, which on upgrade reads as
"login broke". Wildcards survive the step untouched, and two pre-existing
mismatches resolve as a side effect.

Verified with three independent implementations (the shipped two-pointer, a
recursive reference, and a regex compiler) agreeing over **1,863,225 pairs**,
plus a matrix of real URL shapes. A regression sweep over every value/pattern
pair a wildcard-free pattern accepts — including non-canonical pattern hosts —
confirms nothing legitimate stops matching, and it fails if the normalization
is removed.

> **Breaking:** `**` is no longer accepted inside the host of a pattern or a
> `TRUSTED_ORIGINS` entry. It matches the rest of the value outright, so
> `https://**.example.com/**` read as "any subdomain" but accepted every
> origin. A single `*` is unchanged. Stored patterns are not rewritten; new
> writes are rejected and an offending `TRUSTED_ORIGINS` value fails the boot.

## Two pinia copies in the published SSR bundle

Both consoles bundle the kit from source, so `pinia` resolved from
`packages/client-web-kit/node_modules` for kit modules and from the app for
app modules. `dist/server/server.js` carried two complete pinia modules, and
`Symbol('pinia')` is per copy.

It worked only through a module global: installing the store calls the factory
with an explicit instance, and pinia's `useStore` calls `setActivePinia()` on
its own copy. The auth console builds a fresh app and pinia **per SSR
request**, so under concurrent renders that global points at whichever request
installed last and a kit component calling `injectStore()` can read another
request's session.

Fixed with `resolve.dedupe` rather than a root override: the root slot is held
at `@vue/devtools-api` 7.x by vitepress, so an override would drag docs tooling
onto 8.x to fix a bundling problem. Verified on the rebuilt bundle
(`Symbol("pinia")` 2 → 1).

## Wildcard provisioning could delete an explicitly declared entity

A child declared by both a wildcard entry and an explicit realm block
inherited the *wildcard's* strategy when the explicit child declared none.
`absent` is in that vocabulary, so the documented sweep recipe deleted the row
the explicit block declares, on every boot. The inheritance is correct for the
composite source; the wildcard expansion inverts the precedence, so it opts
out.

## Theme token grammar

`url(` was blocked while `image-set()`, `image()`, `src()` and `cross-fade()`
were not, and all take a bare string URL into a token substituted straight
into an image property. `{` was allowed and silently swallowed the `.dark`
rule. Both fixed, and the claim that this grammar is reusable for an untrusted
per-realm source is dropped: a denylist over CSS functions leaks by
construction, which is what happened here.

## Tests that did not assert what they claimed

- The symlink containment test asserted a 404 for a file an earlier test had
  already deleted. It passed with the containment check removed. Now
  self-contained, with a control proving the target is readable through the
  mount. Verified by disabling the check: it fails, serving the leaked file.
- The packed smoke run never requested a console page, so the `locateUpSync`
  walk that only the published layout exercises was untested, and a tarball
  with no bundle went green. It now asserts both consoles serve.

## Smaller items

`entrypoint.sh` exited **0** on an unknown service (so a container that started
nothing succeeded — reachable for anything still passing `client/web`); the
`/settings` stub read its path map through `Object.prototype`; the contract
version was a hand-copied literal with no link to the package's own (now a
type-level link that fails the build on drift); the account nav back-link
modifier was declared before the base rule that overrode it.

`upgrading.md` gained the three breaking changes it was missing, and the
theming quick start printed an `/authorize` URL the verifier rejects (it still
answers 200, so the reader saw an error card with no indication why).

## Verification

`server-core` 1995 tests, `client-web-kit` 201, `kit` 48, `core-kit` 19,
`authup` 30 — all passing. Type check and lint clean. Both console apps build.

## Summary by CodeRabbit

- **Security & Reliability**
  - Strengthened redirect URL validation and blocked unsafe wildcard patterns.
  - Improved theme asset validation to prevent unsafe CSS and file access.
  - Unknown startup services now fail clearly with supported options.

- **Bug Fixes**
  - Improved canonical URL matching for OAuth and logout redirects.
  - Ensured newly available theme assets are detected without restarting.
  - Preserved explicit provisioning settings over wildcard defaults.
  - Corrected account navigation styling.

- **Documentation**
  - Added upgrade guidance, provisioning behavior details, and improved theming smoke-test instructions.

### DIFF
--- a/apps/authup/test/smoke/run.mjs
+++ b/apps/authup/test/smoke/run.mjs
@@ -121,6 +121,36 @@ async function probe(url) {
     }
 }
 
+/**
+ * Assert that server-core can actually resolve and serve a console package.
+ *
+ * The `locateUpSync` walk that finds `@authup/client-auth-console` and
+ * `@authup/client-account-console` exists specifically for the published
+ * install layout, which only the packed scenario reproduces. Nothing else
+ * exercises it: a resolution regression, or a tarball shipped without its
+ * bundle (`npm pack` does not run `prepublishOnly`), degrades silently
+ * because boot still succeeds and only the console routes break.
+ */
+async function assertConsoleServed(name, url, marker) {
+    let response;
+    try {
+        response = await fetch(url, { redirect: 'follow' });
+    } catch (e) {
+        throw fail(`${name}: ${url} could not be reached (${e.message}).`);
+    }
+
+    if (response.status !== 200) {
+        throw fail(`${name}: ${url} answered ${response.status}, expected 200. The console package is probably unresolved or unbuilt.`);
+    }
+
+    const body = await response.text();
+    if (!body.includes(marker)) {
+        throw fail(`${name}: ${url} answered 200 but the body does not contain ${marker}, so the console shell was not rendered.`);
+    }
+
+    log(`${name}: ${url} served the console shell.`);
+}
+
 async function waitUntilReady(name, url, child) {
     const startedAt = Date.now();
 
@@ -208,6 +238,24 @@ async function executeScenario(name, cliExec, cliArgs, cwd) {
     try {
         await waitUntilReady(`${name}/server-core`, SERVER_URL, child);
         await waitUntilReady(`${name}/client-admin-console`, WEB_URL, child);
+
+        // Both consoles are RUNTIME dependencies of server-core, resolved out
+        // of node_modules and served from their built dist. Probing the root
+        // URLs alone leaves that resolution untested.
+        await assertConsoleServed(
+            `${name}/client-auth-console`,
+            `${SERVER_URL}/logout`,
+            'window.__AUTHUP__',
+        );
+        // window.__AUTHUP__ rather than the shell markup: it only appears if
+        // the `<!--account-config-->` marker was found and replaced, which is
+        // that console's entire runtime contract. Without it the SPA silently
+        // degrades to deriving its API URL from the origin.
+        await assertConsoleServed(
+            `${name}/client-account-console`,
+            `${SERVER_URL}/account`,
+            'window.__AUTHUP__',
+        );
 
         log(`${name}: sending SIGTERM.`);
         child.kill('SIGTERM');
--- a/apps/authup/test/smoke/run.mjs
+++ b/apps/authup/test/smoke/run.mjs
@@ -131,16 +131,22 @@ async function probe(url) {
  * bundle (`npm pack` does not run `prepublishOnly`), degrades silently
  * because boot still succeeds and only the console routes break.
  */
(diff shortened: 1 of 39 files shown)


tada5hi:
@coderabbitai pause

PR #3400, opened 2026-08-05
feat!: attribute issued tokens to the client they were issued for

tada5hi:
> **Scope reduced.** This PR originally also carried a client-side gate (a `client_name` token claim, a kit store ref, and an account-console check that re-minted a foreign session through a silent `prompt=none` flow). That was detection layered on top of a *sharing* problem: every surface on the IdP origin uses one set of cookies, so the account console could hold another application's tokens at all. Namespacing the store cookies per client removes the sharing outright, so the gate is unnecessary. That work moves to a follow-up; this PR keeps only the server-side attribution, which is needed either way.

## Problem

`auth_sessions.client_id` was the only per-application attribution, and `resolveSession` overwrote it on every authorize that reused a session:

```ts
existing.clientId = authorizationCode.client_id || null;
```

One browser session serves several applications (the hosted auth pages and the account console share the IdP origin by design), so the column was last-writer-wins and accurate for none of them. Per-app attribution, one of the two stated reasons #3371 split the per-realm system clients, did not hold in this topology.

## What the reference implementations do

Authentik's `AuthenticatedSession` is one per browser and carries **no** application or provider field; the link sits on `BaseGrantModel`, inherited by `AuthorizationCode`, `AccessToken` and `RefreshToken`. Keycloak pairs one `UserSessionModel` per browser login with per-client `AuthenticatedClientSessionModel` children.

Neither creates sibling sessions per application. Both keep the browser session client-agnostic and derive per-application facts one level down. authup already had that two-level structure in `auth_session_tokens`; it simply had no client column. Recorded in `.agents/references/authentik.md`.

## Changes

- **`auth_session_tokens.client_id`** — nullable FK to `auth_clients`, `ON DELETE CASCADE`, mirroring `auth_sessions.client_id`. Both issuers stamp it from the same client already on the token payload, so no lookups are added. Null where the minting path has no client (an MFA-login completion rides a client-less session) and on pre-existing rows.
- **`auth_sessions.client_id` is now write-once** — set only while still null, so it means "the client that first authorized on this session".
- Migration for both dialects.

## Breaking change

`auth_sessions.client_id` now names the client that **first** authorized on a session rather than the most recent one, so `GET`/`DELETE /sessions?filter[clientId]` selects on that basis. The two differ only where several applications share one session. Revocation that should reach every session an application *served* must filter `auth_session_tokens.client_id` instead — that is #3401, stacked on this branch.

## Unblocks

This column resolves the open "audience resolution" decision for OIDC back-channel logout, whose correct audience is every RP that rode the ending session. Notifying `session.client_id` alone is correct only for single-RP sessions; approximating from consent rows under-notifies, because consent is skipped for `builtIn` clients, so neither console would ever be told.

## Verification

- Build 24/24 projects; `check:types` (test-inclusive) and ESLint clean.
- server-core 180/180 files, 2003 tests. core-kit and client-web-kit green.
- Migration round-trip on **MySQL and PostgreSQL** against fresh databases: run → revert ×15 → run, leaving only the `migrations` table at full revert.
- Entity-vs-schema drift check (`createSchemaBuilder().log()`) on both dialects: `client_id` produces no column statement on either, so the migration matches TypeORM's metadata. Note this contradicts a review suggestion to use `varchar(36)` on MySQL — `auth_sessions.client_id` is `varchar(255)` there (`1766830857009-Default.ts`), and `varchar(36)` would introduce drift.
- New `session-token-attribution.spec.ts` pins the core case: two applications on one browser session leave `session.clientId` at the first while each application's token rows carry their own.


tada5hi:
@coderabbitai pause

tada5hi:

PR #3401, opened 2026-08-05
feat(server-core): revoke sessions by the application that used them

tada5hi:
> **Stacked on #3400.** Base is `feat/account-console-session-binding`, because this builds on the `auth_session_tokens.client_id` column added there. Review or merge #3400 first; GitHub will retarget this to `master` automatically once it lands.

## Problem

Sessions can only be found by `auth_sessions.client_id`, and #3400 makes that column write-once, so it names the client that **first** authorized on the row. One browser session serves several applications, so an operator revoking application X after a compromise misses every session X served but did not create.

Concretely, with the admin console opened first and the account console second, `DELETE /sessions?filter[clientId]=<account-console>` matches nothing. Before #3400 the column was last-writer-wins, so it matched — differently partial, but partial either way. Neither answers "every session this application served".

## Change

`GET` and `DELETE /sessions` accept `?usedClientId=<id>`, a single id or a comma list. A session matches when **either**

- its own `client_id` names one of them (the client that first authorized), **or**
- an `EXISTS` over `auth_session_tokens.client_id` finds a token issued for one of them.

Both reaches are needed. The column alone misses every later application; the `EXISTS` alone misses a session whose token rows have since been swept by the cleanup component.

## Why a request parameter and not a rapiq filter

Inferring this from a user-supplied condition tree means deciding what a negated or OR-nested `clientId` means. On a bulk delete a wrong answer is a mass deletion, so the safe move is a parameter with exactly one meaning. It composes as a mandatory AND-ed bracket next to the existing `owner` constraint, so no client-supplied filter can widen it.

Its presence selects the admin bulk-revoke path on its own. Without that it would fall through to self-service and silently revoke the caller's own devices instead of the named application's sessions. That means it carries the same `SESSION_DELETE` gate and per-session realm match as any target filter — filter breadth still cannot escalate, so a `realm_admin` reaches only its own realm's sessions.

`filter[clientId]` is untouched and still matches the session column.

## Verification

- server-core 180/180 files, full suite green. `check:types` clean, ESLint clean.
- New test in `session-token-attribution.spec.ts` pins the actual capability gap: with two applications on one browser session, `filter[clientId]=<second app>` does **not** return the session, `usedClientId=<second app>` does, and the revoke removes it.

## Follow-up left open

The account console's Applications page revokes *consent*, which is prompt-level: it stops the next silent issue and leaves live tokens working. It could now offer a real "sign out of this application everywhere" by calling this endpoint. Not done here, and note the page lists consent rows, which `authorization/module.ts:57` skips for `builtIn` clients, so neither console appears there.

### DIFF
--- a/apps/server-core/src/adapters/http/controllers/entities/session/module.ts
+++ b/apps/server-core/src/adapters/http/controllers/entities/session/module.ts
@@ -27,6 +27,28 @@ import {
 import { ForceLoggedInMiddleware } from '../../../middleware/index.ts';
 import { buildActorContext, useRequestSessionId } from '../../../request/index.ts';
 
+/**
+ * Read the `usedClientId` request parameter: the "every session that served
+ * application X" target, accepting a single id or a comma list.
+ *
+ * It is deliberately not a rapiq filter. `filter[clientId]` matches the
+ * session's own column, which names only the client that FIRST authorized on
+ * the row, and inferring intent from a user-supplied condition tree (what
+ * does a negated or OR-nested value mean?) is not something a bulk delete
+ * should be doing.
+ */
+function useRequestUsedClientIds(event: IAppEvent) : string[] | undefined {
+    const raw = useRequestQuery(event)?.usedClientId;
+
+    const values = (Array.isArray(raw) ? raw : [raw])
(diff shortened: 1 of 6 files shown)


tada5hi:

PR #3402, opened 2026-08-05
feat!: namespace store cookies per client

tada5hi:
> **Stacked on #3400.** Base is `feat/account-console-session-binding`. GitHub retargets this to `master` once that lands.

## Problem

Cookies are per **origin**, not per application, and the hosted auth pages share their origin with everything else server-core serves. So every surface there used one set of tokens under bare names (`access_token`, `refresh_token`, …).

Two consequences, in both directions:

- The account console **adopts** whatever login came before it, so the `account-console` client's `accessPolicyId` is never evaluated for a visitor arriving with a lingering session.
- Its own tokens are **readable** by the next surface on the origin, including the admin console or any trusted-origin app sharing the domain.

Locally it is broader still: cookies ignore the port, so an app on `localhost:3000` shares the jar with the IdP on `:3001`.

An earlier revision of #3400 answered this by *detecting* a foreign session (a `client_name` claim, a store ref, a gate that re-minted through a silent `prompt=none` flow). That was a check layered on top of a sharing problem. This removes the sharing, so the question cannot arise, and it closes the reverse leak the detection approach never addressed.

## Change

The store takes a `cookiePrefix` and writes `<prefix>.<name>`. The prefix alone distinguishes two tiers:

- **Bare names are the IdP's own SSO session**, owned by the hosted auth pages. They are not a client, they are the IdP surface, and their session is what the `prompt=none` / `select_account` ladder reads. `apps/client-auth-console` is therefore unchanged.
- **A prefixed set belongs to one application.** Both consoles set it to their OAuth2 client name; any RP embedding the kit should too.

Threaded through `@authup/client-web-kit`'s install options and `@authup/client-web-nuxt`'s module options.

## Two decisions worth stating

**`.` and not `:`.** RFC 6265 defines a cookie name as an RFC 2616 `token`, whose separator set includes `:`. Browsers accept it; proxies, WAFs and cookie libraries do not uniformly.

**Client name, not client id.** The name is what an application knows *before* it has a session; the id it cannot know until it has one, which is exactly when it needs to read a cookie. Names are unique per realm, so two realms' same-named clients still collide on one origin and switching realms replaces the session — which is what the single cookie set already did, so this preserves the behaviour rather than regressing it.

## Load-bearing details

- **Clearing shadowed copies follows the prefix and leaves the bare tier alone.** Clearing it would sign the visitor out of the IdP as a side effect of loading an application. Pinned by a test that fails if the implementation reverts to the unprefixed name list (verified by reintroducing the bug).
- **The admin console gets a prefix too**, despite its own origin, because the default dev setup shares a host with server-core.
- **The account console reads one value out of the bare tier: the realm.** A visitor arriving from another application now has no tokens here and would otherwise see a realm chooser while plainly signed in on the IdP. The tokens beside it are deliberately not read.

## Breaking change

Cookies written by an application configured with a `cookiePrefix` use `<prefix>.<name>`. Existing bare cookies are **not** migrated — they cannot be attributed to an application, since they were written by the hosted login and possibly overwritten by a console — so each prefixed application signs out once on upgrade. A consumer reading `CookieName` values off the document directly must add the same prefix. Inside this repo `CookieName` has exactly one consumer (`core/store/install.ts`), but it is exported from `@authup/core-http-kit`'s public surface, so **hub and portal should be checked before release notes.**

## Verification

- Build 24/24 projects; ESLint clean.
- client-web-kit 29 files / 206 tests, account console 3 files / 23 tests, server-core 180/180 unaffected.
- New tests: hydration from the namespaced names; the bare tier is **not** adopted; writes never touch the bare tier and this app's tokens are not readable under bare names; the pinned expire-date-before-token hydration order survives the prefix; shadow-clearing touches only this namespace. Plus a spec for the realm reader covering malformed, non-string, empty and non-object payloads, since any same-origin script can write that cookie.

### DIFF
--- a/apps/client-account-console/src/main.ts
+++ b/apps/client-account-console/src/main.ts
@@ -14,6 +14,7 @@ import {
     loadAuthorizationRequest,
     syncTranslatorLocaleFromManager,
 } from '@authup/client-web-kit';
+import { CLIENT_ACCOUNT_CONSOLE_NAME } from '@authup/core-kit';
 import { matchLocale } from '@authup/i18n';
 import { omitRecord } from '@authup/kit';
 import { OAuth2ErrorCode } from '@authup/specs';
@@ -199,6 +200,11 @@ const localeHandles = installLocale(app, {
 install(app, {
     baseURL: config.apiUrl,
     pinia,
+    // This console is served from the IdP origin, so without a namespace it
+    // shares one cookie set with the hosted auth pages: it would adopt
+    // whatever login came before it, and its own tokens would be readable by
+    // the next surface on the origin (plan 087).
+    cookiePrefix: CLIENT_ACCOUNT_CONSOLE_NAME,
     translatorLocale: matchLocale(localeHandles.resolved.value),
 });
 
--- a/apps/client-account-console/src/pages/index.vue
+++ b/apps/client-account-console/src/pages/index.vue
@@ -44,6 +44,7 @@ import {
 import { useRoute, useRouter } from 'vue-router';
 import { injectAccountConsoleConfig } from '../di';
 import { saveAccountConsoleRef, useAccountConsoleRef } from '../ref';
+import { readSsoRealmId } from '../sso-session';
 import { useAccountToasts } from './utils';
 
 export default defineComponent({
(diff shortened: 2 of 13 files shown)


tada5hi:

PR #3404, opened 2026-08-05
feat!: per-application session token attribution, inventory API and revocation

tada5hi:
Makes per-application session state real: attributed on the server, queryable over the API, and acted on by the account console. Four commits, reviewable in order.

## 1. Attribute issued tokens to their client (`feat!`)

`auth_sessions.client_id` was the only per-application attribution, and `resolveSession` overwrote it on every authorize that reused a session. One browser session serves several applications (the hosted auth pages and the account console share the IdP origin by design), so the column was last-writer-wins and accurate for none of them. Per-app attribution, one of the two stated reasons #3371 split the per-realm system clients, did not hold in this topology.

- `auth_session_tokens` gains a nullable `client_id` (FK to `auth_clients`, `ON DELETE CASCADE`). Both issuers stamp it from the client already on the token payload, so no lookups are added. Null where the minting path has no client (an MFA-login completion rides a client-less session) and on pre-existing rows.
- `auth_sessions.client_id` becomes write-once: "the client that first authorized on this session".
- Migration for both dialects.

This is the shape both reference implementations use. Authentik's `AuthenticatedSession` is one per browser and carries no provider field, with the link on `BaseGrantModel`; Keycloak pairs one `UserSessionModel` with per-client `AuthenticatedClientSessionModel` children. Neither creates sibling sessions per application. Recorded in `.agents/references/authentik.md`.

## 2. Expose the token inventory as a resource

`auth_session_tokens` was the one table with no query surface, so questions about it had to be answered by bolting parameters onto `/sessions`. An earlier revision did exactly that with `?usedClientId=`: a session-shaped endpoint answering a token-shaped question, implemented as a repository option carrying raw cross-table SQL. It also could not express the thing that matters, revoking one application while the session keeps serving the others. It has been replaced.

`/session-tokens` is read and revoke only, since rows are created solely by the grants.

| Query | Answers |
|---|---|
| `filter[sessionId]` | which applications rode this session (the back-channel logout audience) |
| `filter[clientId]` | every session an application served, which `sessions.client_id` cannot: it names only the first authorizer |
| both | revoke one application on one session, leaving the session alive |

**`DELETE` means revoke, not remove.** The row is stamped and the jti is blocklisted with a TTL pinned to the token's real expiry. Dropping the row would block refresh (the row is the authority for refresh validity since plan 016) while leaving an already-issued access token verifying against a stateless JWKS adapter until its own `exp`.

**Authorization resolves through the session**, since the rows carry neither realm nor subject. The repository joins it unconditionally and pins the columns the gate reads, so a `fields` projection cannot neutralize it (the plan-039 discipline, through a relation instead of a column). An actor without `SESSION_READ` is scoped to its own sessions' tokens by an AND-injected condition no client filter can displace. A bulk revoke must carry a target filter and is rejected otherwise: unlike `DELETE /sessions` there is no self-service fallback to degrade to, so an unscoped call has to fail loudly rather than widen.

## 3. Typed client

`client.sessionToken.{getMany,getOne,delete,deleteMany}`. Named after the `EntityTypeMap` key, so it joins the derived entity-API registry automatically; the cast-free assignment in `pickEntityAPI` is the compile-time proof that it serves the `SessionToken` record type.

`deleteMany` takes its query as a **required** argument, unlike the session counterpart: the server rejects an unscoped bulk revoke, and an optional parameter would invite writing the one call shape that always fails.

## 4. The account console actually signs out of a revoked application

"Revoke access" removed consent rows only. Consent is prompt-level, so it stops the next silent issue and leaves already-issued tokens working — the application the user just revoked kept working. It now revokes that application's tokens too, and because the rows hang off the session rather than being it, the other applications on the same session stay signed in.

The call is scoped to the user's **own sessions explicitly, by id**. The server force-scopes a caller that lacks `SESSION_DELETE`, but an administrator holds it, so relying on that would have an admin revoke the application for every user from their own account page.

## Breaking change

`auth_sessions.client_id` is no longer written for **user**-subject sessions.

It carries a nullable `user_id` **and** a nullable `client_id`, and `SessionManager.create` populates exactly one of them from `sub` according to `subKind` — the same trick `auth_consents` uses. They are typed foreign keys for a polymorphic subject, so `ON DELETE CASCADE` can drop a subject's own sessions.

Three writers were instead putting the *application* into `client_id` on user sessions (the password grant, the MFA-login ticket, and the authorize grant). That placed an unrelated id behind a cascade meaning "this client owns this row", so **deleting an application deleted a user's session** and, through `auth_session_tokens.session_id`, every other application's tokens riding it. Deleting the account console would have signed users out of the admin console.

The column now has exactly one meaning: the client-subject FK. Consequences:

- `GET`/`DELETE /sessions?filter[clientId]=` now selects sessions whose **subject** is that client, symmetric with `filter[userId]`. It no longer returns a user's sessions because some application authorized on them.
- "Which application served this session" is `/session-tokens?filter[clientId]=`, where a session serving several applications can say so.

## Unblocks

Resolves the open "audience resolution" decision for OIDC back-channel logout, whose correct audience is every RP that rode the ending session. Notifying `session.client_id` alone is correct only for single-RP sessions; approximating from consent rows under-notifies, since consent is skipped for `builtIn` clients, so neither console would ever be told.

## Verification

- Build 24/24; ESLint and `check:types` (test-inclusive) clean.
- server-core 181 files / 2008 tests; core-http-kit 24; account console 14.
- Migration round-trip on MySQL **and** PostgreSQL against fresh databases: run → revert ×15 → run, leaving only the `migrations` table at full revert.
- Entity-vs-schema drift check (`createSchemaBuilder().log()`) on both dialects: `client_id` produces no column statement, so the migration matches TypeORM's metadata. This contradicts an earlier review suggestion to use `varchar(36)` on MySQL — `auth_sessions.client_id` is `varchar(255)` there (`1766830857009-Default.ts`), and `varchar(36)` would introduce drift.
- `session-token-attribution.spec.ts` pins two applications on one session leaving `session.clientId` at the first while each application's rows carry their own. `session-token.spec.ts` pins the resource: the audience query, finding every session an application served, a per-application revoke that leaves the session and the other application intact, rejection of an unscoped bulk revoke, and single-token revoke.

**Not verified by execution:** `apps/client-account-console/src/pages/applications.vue` has no spec, and I did not run the app to watch a revoke sign the user out. The endpoint it calls is covered end to end (including that the revoked application's refresh is rejected while the untouched one still works), and the change is ~26 lines, but the UI path itself was verified by reading, not by running.

One caveat on the suite: the `provisioning.spec.ts` wildcard-realm case flakes intermittently under parallel workers against the shared sqlite file. It passes in isolation and on re-run, and this branch does not touch provisioning.

## Summary by CodeRabbit

* **New Features**
  * Added session-token APIs for listing, viewing, and revoking individual or multiple tokens.
  * Added filtering, pagination, ownership scoping, and application/session-based access controls.
  * Issued access and refresh tokens now retain application attribution for clearer activity tracking.
* **Bug Fixes**
  * Revoking an application’s consent now signs the user out of that application without affecting others.
  * Reused sessions retain their original authorization context while newly issued tokens are attributed correctly.
* **Documentation**
  * Clarified session and account-console attribution behavior.


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3418, opened 2026-08-09
test(server-core): stop parallel workers from sharing one test database

### LINKED ISSUE #3405: server-core test suite flakes under parallel workers: every spec file shares one sqlite database
## Symptom

The `apps/server-core` suite fails intermittently on a **different** spec each run, while a full re-run is usually green. Observed across a single afternoon, all on branches touching none of these areas:

- `test/unit/modules/provisioning.spec.ts > should apply a wildcard realm entry to every realm` (most frequent)
- `test/unit/http/controllers/entities/session.spec.ts > revokes every session of the actor except the current one`
- `test/unit/http/controllers/entities/key.spec.ts > should eagerly provision sig + enc keys at realm creation`
- `test/unit/http/controllers/workflows/token/grant-refresh-token.spec.ts > should grant refresh for a PUBLIC bound token without client auth`

Every one of them passes in isolation. The provisioning case fails as `SqliteError: FOREIGN KEY constraint failed` inside `ClientProvisioningSynchronizer.synchronize` → `clientRepository.save`.

## Cause

Every test application points at the **same** sqlite file, with no per-worker suffix:

`apps/server-core/test/app/database.ts:29-31`

```ts
config.db = {
    type: 'better-sqlite3',
    database: path.join(PACKAGE_PATH, 'writable', 'test.sql'),
};
```

`apps/server-core/test/vitest.config.ts` sets no `fileParallelism` or `pool` option, so vitest runs spec **files** in parallel workers by default. Concurrent files therefore share one database file and one dataset, including the provisioned master realm, the admin user and the system clients.

That matches the symptom exactly: the failing spec varies run to run (a deterministic bug would fail the same test every time), and isolation always passes. `createTestDatabaseModuleForSetup` additionally does `fs.rmSync(options.database)` on setup, so one worker can delete the database another is mid-transaction on.

## Why it is worth fixing

It trains everyone to re-run and move on, which is exactly how a real regression gets waved through. During the work on #3404 this cost several full-suite re-runs purely to establish that a failure was not caused by the change under review.

## Options

1. **Per-worker database file** — suffix the path with `VITEST_WORKER_ID` (or `VITEST_POOL_ID`). Keeps parallelism, and sqlite files are cheap. Needs `test/setup.ts`'s provisioning to run per worker rather than once.
2. **`fileParallelism: false`** — one-line fix, guaranteed correct, but serialises the whole suite. Worth measuring the wall-clock cost before rejecting.
3. **Per-file in-memory sqlite** (`:memory:`) — full isolation and fastest, but each file pays a schema `synchronize()` plus provisioning, and any spec relying on cross-file state would surface.

I would try (1) first and fall back to (2) if the provisioning cost per worker turns out to dominate.

## Notes

- The MySQL and PostgreSQL CI runs are unaffected: each gets its own container and database.
- Unrelated to the LDAP suite failures, which come from a stale docker container.

tada5hi:
## Problem

Closes #3405.

Every spec file pointed at the same `writable/test.sql`, while vitest runs spec files in parallel workers by default. Concurrent files corrupted each other's rows (provisioned realms, the admin user, system clients), so the suite failed intermittently on a different spec each run and always passed in isolation.

The issue assumed the MySQL/Postgres runs were unaffected. They are not: `test:mysql` with 13 local workers reproduced the issue's headline failure (`SystemClientProvisioner.ensureForRealm` hitting the `auth_clients.realm_id` FK because a concurrent spec deleted the realm) against the one shared `app` database. CI has simply been running with fewer cores.

## Fix

- **sqlite (issue option 1):** the global setup provisions `writable/test.sql` as a template only. `createTestDatabaseModuleForSuite` gives every worker its own copy (`writable/test-<poolId>.sql`, keyed by `VITEST_POOL_ID`), created on first use and swept by the next global setup. Files that reuse a pool slot share its copy sequentially, which matches the previous single-file semantics minus the cross-worker races. A missing template (running vitest without the global setup) now fails with an actionable message instead of half-working against a stale file.
- **mysql/postgres (issue option 2, scoped):** a server database has no cheap per-worker copy, so `test/vitest.config.ts` turns `fileParallelism` off when `DB_TYPE` names one. The default sqlite run keeps full parallelism.
- Both test database modules now discriminate the sqlite branch on `options.type` instead of `typeof options.database` (a mysql/postgres database name is a string too). That makes the previously unreachable `dropDatabase` branch live, so `test:mysql` / `test:psql` start from a dropped database instead of a stale one (the dirty local `app` databases that have confused several past sessions).

## Verification

- sqlite: 3 consecutive full-suite runs green (181 files, 2023 tests, ~22s wall-clock, unchanged).
- mysql, parallel, before the serialization: reproduced the wildcard-realm FK failure.
- mysql, serial (local compose): full suite green in 184s.
- postgres (local compose): scoped run green, exercising the drop + provision path on that dialect.
- `npm run check:types -w apps/server-core` green.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for isolating SQLite test databases across parallel workers.

* **Tests**
  * Improved SQLite test isolation by creating separate databases for each worker.
  * Preserved appropriate cleanup and setup behavior for different database types.
  * Enabled parallel SQLite test execution while preventing unsafe parallelism for shared MySQL and PostgreSQL databases.

- CI on this PR: all checks green. The serialized `tests-server-core` jobs took 7m04s (mysql) and 6m41s (postgres) against a ~4m baseline on master, both including the ~2.5m checkout/install/build preamble: the test step roughly doubled, on 4-core runners that only ever ran 3-4 workers.

### DIFF
--- a/.agents/testing.md
+++ b/.agents/testing.md
@@ -13,6 +13,18 @@
   so its own `import 'reflect-metadata'` does not reach test-file
   contexts. Any spec importing `src/core` transitively loads x509 (via the
   client-certificate module), so this must stay — do not remove it.
+- **Per-worker sqlite isolation (server-core, issue #3405)**: spec files run
+  in parallel vitest workers, so a shared database file made the suite flake
+  on a different spec each run. The global setup provisions a template
+  database at `writable/test.sql`; `createTestDatabaseModuleForSuite` gives
+  every worker its own copy (`writable/test-<poolId>.sql`, keyed by
+  `VITEST_POOL_ID`), created on first use in that worker and swept by the
+  next global setup. Files that reuse a pool slot share its copy
+  sequentially, so a spec must never rely on state written by another spec
+  file. The MySQL/Postgres runs have no per-worker copy (one shared server
+  database), so `test/vitest.config.ts` turns `fileParallelism` off for
+  them instead; expect those runs to take several times the sqlite
+  wall-clock.
 
 ## Running Tests
 
--- a/apps/server-core/test/app/database.ts
+++ b/apps/server-core/test/app/database.ts
@@ -14,7 +14,20 @@ import { ConfigInjectionKey, DatabaseModule } from '../../src';
 import type { IContainer } from 'eldin';
 import { PACKAGE_PATH } from '../../src/path.ts';
 
-async function resolveDataSourceOptions(container: IContainer) {
+const DATABASE_DIRECTORY_PATH = path.join(PACKAGE_PATH, 'writable');
+const TEMPLATE_DATABASE_PATH = path.join(DATABASE_DIRECTORY_PATH, 'test.sql');
+const WORKER_DATABASE_REGEX = /^test-\d+\.sql$/;
+
+// Spec files run in parallel worker processes, so every worker gets its own
+// copy of the provisioned template database. VITEST_POOL_ID is stable per
+// pool slot: two concurrently-running files never share it, while files that
+// reuse a slot sequentially share the copy (the pre-existing single-file
+// semantics, minus the cross-worker races).
+function buildWorkerDatabasePath(): string {
+    return path.join(DATABASE_DIRECTORY_PATH, `test-${process.env.VITEST_POOL_ID || '0'}.sql`);
+}
+
+async function resolveDataSourceOptions(container: IContainer, database: string) {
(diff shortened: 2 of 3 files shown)


tada5hi:
@coderabbitai pause

PR #3430, opened 2026-08-12
feat: title-row action, breadcrumbs and record sub titles for the admin console

### LINKED ISSUE #3427: client-admin-console: replace the overview+add rail with a route-aware title-row action
## Summary

Two UI changes made in [FLAME Hub](https://github.com/PrivateAIM/hub), which mirrors this admin console closely. Filing them here since the same pattern exists in `client-admin-console` and the reusable half arguably belongs in `client-web-kit`.

## 1. Replace the "overview + add" rail with a title-row action

All 12 collection pages in `apps/client-admin-console/pages/*/index.vue` render a two-item vertical pill rail beside the list:

```vue
<div class="content-wrapper">
    <div class="content-sidebar flex-col">
        <VCNavItems :data="items" variant="pills" orientation="vertical" />
    </div>
    <div class="content-container">
        <NuxtPage />
    </div>
</div>
```

where `items` is always `overview` + `add`. That spends a fixed 200px column and a navigation affordance on what is really one primary action, on every page, at every viewport.

### Proposal

Drop the rail to a bare `<NuxtPage />` and move the action into the page's existing `<h1>` row:

```
┌─ /clients ──────────────────────────────┐
│ 👻 Client Management         [ + Add ]  │
│                                         │
│  Clients             [search......]     │
│  name    active   auth   built-in       │
└─────────────────────────────────────────┘
```

The component is route-aware, so a section needs exactly one of them regardless of which of its routes is active:

| current route | renders |
|---|---|
| `=== overviewUrl` | `+ Add` (primary) |
| `=== addUrl` | `← Back` (outline) |
| anything else | nothing |

The `← Back` case matters: the add form is a route of its own, so removing the rail without it leaves a dead end. The "nothing" case matters when a section has more than one list route — Hub has `/projects` and `/projects/in`, where the second is an approval queue and an "add" there would be the wrong verb for the wrong audience.

The disabled state should carry the create permission (`usePermissionCheck({ name: PermissionName.CLIENT_CREATE })` and friends). `VCButton` blocks activation of a disabled link target, so it cannot navigate into a page the actor would be rejected from — today's rail is ungated and links into a page that rejects you.

Hub's implementation is a single ~110-line SFC, `FContentAction`, with props `overviewUrl` / `addUrl` / `addDisabled`. The authup equivalent would sit in `client-web-kit` as `AContentAction`. Worth noting it needs `vue-router` (it reads `useRoute().path`), which `client-web-kit` does not currently declare.

## 2. Nest paired list routes under one sidebar group

Less directly applicable — it came out of Hub's domain nav, which has Outgoing/Incoming pairs the admin console does not. Recording the two findings from it, both of which apply to any group-bearing nav:

- **A `@vuecs/navigation` group cannot be a link.** `renderLeaf` runs only when `!hasChildren`, so a parent with both `url` and `children` renders a trigger and no link. Filed separately as tada5hi/vuecs#1714, along with a proposed `expanded: true` for always-open groups.
- **The nav-filtering reducer needs two guards** that Hub's did not have, and this console's `Navigation` class shares the shape. First, a parent whose children are *all* permission-filtered away must be dropped — `VCNavItem` renders a childless item as a leaf, so a url-less group becomes an inert dead entry. Second, the reducer must not write the filtered array back into `item.children`: those items are the module-level nav constant, so one resolve prunes the source permanently and a later resolve with wider permissions can never restore it. In Hub this was reachable on a realm switch; a filtered-out child stayed gone for the life of the page.

## 3. Do not paint nav active state from `router-link-active`

Also generic. If a theme styles `.router-link-active` / `.nuxt-link-active`, note those are vue-router's **prefix** classes: standing on `/projects/in` lights up a `/projects` item too, so a group with paired child routes paints both rows active.

`@vuecs/navigation` already computes the right thing — `findBestItemMatches` elects exactly one winner per nav and `VCNavItem` feeds it to `<VCLink :active>` as a plain `.active` class, which still covers the prefix case (at `/clients/<id>` the `/clients` item is the top scorer). Styling `.active` and the `*-exact-active` classes only, and dropping the prefix ones, fixes it.

tada5hi:
Closes #3427, and ports the breadcrumb half of PrivateAIM/hub#1825.

## Navigation

**The `overview + add` rail is gone from all twelve collection pages.** It spent a fixed 200px column and a navigation affordance on what is really one primary action, at every viewport. A page now renders a breadcrumb, then a title row carrying the heading and an `<AContentAction>`, then its `<NuxtPage>` at full width.

`AContentAction` lives in `client-web-kit` and is route-aware, so a section needs exactly one regardless of which route is active:

| route | renders |
|---|---|
| `overviewUrl` | `+ Add` |
| `addUrl` | `← Back` |
| anything else | nothing |

The last case is load-bearing: a section that later grows a second list route does not inherit the button. `/events` and `/sessions` have no add route and simply lose the rail.

The button is permission-gated. The rail it replaces was **ungated** and linked into pages that reject the actor.

## Breadcrumbs

On all twelve collection and all twelve detail pages. Detail trails are built from the **resolved record**, not from url segments, so they cannot go stale, and the same helper supplies both the crumb and the `<h1>` so the two cannot disagree.

Sections are declared once in `LayoutSections` (route, icon, label), which the sidebar now reads too. `buildSectionURLs` owns the `/add` suffix for its two consumers, because `AContentAction` renders nothing on a route matching neither of its props: a section whose url moved would otherwise lose its last entry point to the create form, with no error.

## Sub titles

The sub title moved out of the `<h1>` onto its own line and now says something. Sections take one of twelve new description keys (authored in en, de, fr and es). Records walk a ladder: their own `description`, else the `name` the heading gave up to a display name, else the `id`. The entity type never appears, because the breadcrumb and the sidebar already state it, which is why the provisioned `master` realm used to read "Realm" under a heading saying "master".

## Search

`ASearch` now matches **name and display name together**. #3429 made `displayName` filterable and index-leading on eight entities, but no page ever sent it: all 24 call sites pass no `queryFilters` hook, so the server-side half was unreachable from the UI. The fix is one default in the collection manager rather than 24 page edits, driven by a map mirroring each schema's `filters.allowed`.

## Bugs fixed along the way

- **`reduceItem` mutated the module-level nav constant.** One resolve pruned the source permanently, so no later resolve with wider permissions could restore it. Reachable on a realm switch, where a filtered-out child stayed gone for the life of the page.
- **`/trust-anchors/:id` answered 500** with "No ToastManager available". It was the only detail page calling `useToast()` below its `await`, where `inject()` no longer resolves. Pre-existing on master.
- **The policies add route had no `definePageMeta` at all**, so it required neither a login nor a permission while its nine siblings each declare one.
- **Breadcrumbs rendered one crumb per line.** `@vuecs/theme-tailwind` styles the breadcrumb's colour but not its layout, which lives in a `@vuecs/navigation` stylesheet no authup app imports.

## Verification

`client-web-kit` 209 tests, `i18n` 77, `server-core` query specs 66, `nuxi typecheck` clean, eslint clean.

Two guards were negative-controlled rather than trusted:

- `content-action.spec.ts` asserts a disabled action leaves `router.currentRoute` on the overview route, not merely that the attribute is present, which would pass either way.
- `console-search-surface.spec.ts` fails when `displayName` is removed from a schema, while `indexed-invariant.spec.ts` stays green at 17 passed. That sibling derives its search shape from the allow-list, so it quietly stops exercising the compound shape instead of failing.

Behaviour was checked in a browser against a real server: every collection page, the add and back round trip, detail and tab routes, the three sub title rungs, dark mode, and a display-name-only search returning its row.

## Not verified

Narrow viewports below roughly 1400px, and the disabled action under a genuinely restricted actor (the permission names were checked, not the runtime behaviour). `apps/client-admin-console` still has no test harness, so the composables and pages carry no committed test.

## Summary by CodeRabbit

* **New Features**
  * Added breadcrumb navigation across administration pages for clearer orientation.
  * Replaced collection-page sidebars with streamlined titles, descriptions, and contextual Add/Back actions.
  * Added permission-aware controls that disable or hide actions when access is unavailable.
  * Improved detail-page headings with record names, descriptions, and helpful fallbacks.
  * Expanded entity search to include display names for supported entity types.
* **Bug Fixes**
  * Prevented invalid or empty navigation groups from appearing after permission filtering.
  * Improved search behavior and preserved navigation state consistency.
* **Localization**
  * Added translated breadcrumbs and entity descriptions in English, German, Spanish, and French.


tada5hi:
@coderabbitai pause

PR #3437, opened 2026-08-13
fix: answer invalid_grant when a session is revoked mid-rotation

### LINKED ISSUE #3435: Refresh-token rotation can 500 with a foreign-key violation when the session is revoked mid-flight
## Observed

In a running deployment (server-core 1.0.0-beta.59, postgres), a refresh-token replay revoked a session and, 88ms later, a second request died with an unhandled driver error:

```
warn : OAuth2 refresh token replay detected; revoking session {"jti":"1c64c6ba…","sessionId":"e4c8cc9a…"}
info : audit: oauth2.refreshReplayDetected {…}
error: insert or update on table "auth_session_tokens" violates foreign key constraint "FK_auth_session_tokens_session_id"
    at SessionTokenRepositoryAdapter.create (…/repositories/session-token/repository.mjs:26:3)
    at OAuth2RefreshTokenIssuer.issue (…/core/oauth2/token/issuer/refresh/module.mjs:27:99)
    at HTTPOAuth2RefreshTokenGrant.runWith (…/core/oauth2/grant-types/refresh-token.mjs:58:47)
    at TokenController.createToken
```

The client receives `500 internal_error` instead of an OAuth2 error.

## Cause

Within a single request this cannot happen: `revokeFamily` is immediately followed by `throw OAuth2GrantError.invalid(...)`, so the revoking request never reaches the issuer. The two log lines are therefore two concurrent requests over the same token family.

`OAuth2RefreshTokenGrant.runWith` looks the session up and then issues:

```ts
const session = await this.sessionManager.findOneById(payload.session_id);
if (!session) { throw JWTError.payloadPropertyInvalid('session_id'); }
await this.sessionManager.verify(session);
await this.sessionManager.refresh(session);

const [refreshToken, refreshTokenPayload] = await this.refreshTokenIssuer.issue({ … });
```

`revokeFamily` calls `sessionManager.revoke(sessionId)`, which deletes the `auth_sessions` row. If that delete lands between request A's `findOneById` and the row insert inside `issue()`, the insert references a session that no longer exists and the FK rejects it.

Concretely:

1. A and B both present the same refresh token.
2. A wins `markRefreshConsumed` and proceeds.
3. B loses it, finds no grace window (default `tokenRefreshGracePeriod` is 0), calls `revokeFamily` → the session row is deleted.
4. A reaches `refreshTokenIssuer.issue()` and its insert violates the FK.

The same window is open to any other concurrent session delete: an explicit logout, `DELETE /sessions/:id`, an admin force-logout, or the session sweeper.

## Impact

Correctness and observability rather than security. The revoke itself is right, and A gets no tokens. But:

- the status is wrong: a client that should see `invalid_grant` and re-authenticate gets a 500, which most OAuth2 clients treat as retryable
- A's presented token was already consumed and cache-blocklisted before the failure, so there is nothing to retry with
- it logs as an internal error, which hides a benign-and-expected race among real faults

## Possible fixes

1. Translate the constraint failure into `OAuth2GrantError.invalid(...)`. `adapters/database/errors/driver.ts` already has this shape for uniqueness (`isUniqueConstraintDatabaseError`, used by the consent insert race); a sibling `isForeignKeyConstraintDatabaseError` (`ER_NO_REFERENCED_ROW_2` / `23503` / `SQLITE_CONSTRAINT_FOREIGNKEY`) would let the grant catch and re-raise cleanly.
2. Re-read the session immediately before issuing. Narrows the window without closing it, so it is a complement to (1) rather than an alternative.
3. Run consume-through-issue in one transaction. Closest to correct, biggest change, and it does not help when the delete is committed by an unrelated request.

(1) alone fixes the observable symptom and is the smallest change.

## Reproduction sketch

Fire two concurrent `grant_type=refresh_token` requests with the same token and `tokenRefreshGracePeriod=0`, with a delay injected between `findOneById` and `issue()` in the winner. The loser revokes the session; the winner's insert then fails the FK.

tada5hi:
Closes #3435.

## Problem

A refresh-token rotation resolves the session and then issues, holding nothing in between:

```ts
const session = await this.sessionManager.findOneById(payload.session_id);
// ...
const [refreshToken, refreshTokenPayload] = await this.refreshTokenIssuer.issue({ ... });
```

A concurrent request can delete the `auth_sessions` row in that gap: the replay reaction on this same token family (two requests presenting one token, the loser calling `revokeFamily` while the winner is still issuing), an explicit logout, `DELETE /sessions/:id`, an admin force-logout, or the sweeper. The inventory insert is then rejected by `FK_auth_session_tokens_session_id` and the driver error escaped as an unhandled internal error.

Refusing the request is correct. The status was not: the client saw `500 internal_error`, which most OAuth2 clients treat as retryable, while its presented token was already consumed and cache-blocklisted, so there was nothing left to retry with. It also logged as a fault rather than the benign-and-expected race it is.

## Fix

Item (1) from the issue. Since `core/**` deliberately never imports `adapters/database/**`, the driver knowledge stays in the adapter and the translation happens in two hops:

- `adapters/database/errors/driver.ts` gains `isForeignKeyConstraintDatabaseError`, the sibling of the existing `isUniqueConstraintDatabaseError` that the consent-insert race already uses.
- `SessionTokenRepositoryAdapter.create` catches a foreign-key rejection and raises `SessionTokenSessionMissingError` (`core/oauth2/session-token/error.ts`), declared on the `ISessionTokenRepository.create` port. `session_id` is the only foreign key an insert there can violate, so the condition is unambiguous.
- `OAuth2RefreshTokenGrant.runWith` wraps both issuer calls and re-raises that one error as `invalid_grant`. Everything else is rethrown untouched, so a signing or cache fault still surfaces as a fault.

Items (2) and (3) from the issue are deliberately skipped. A pre-issue re-read closes nothing, since the delete can still land one statement later, and once (1) is in place the client-facing outcome is identical either way.

The new error extends `ValidationError` and reuses `ErrorCode.ENTITY_RELATION_INVALID`, so it adds nothing to the public error vocabulary and needs no new i18n copy. Its guard is marker-only on purpose: that code is shared with typeorm-extension's join-column lookup failures through `sanitizeError`, so a `code` fallback would claim unrelated errors.

Both issuers write inventory rows, so the error can also reach the password / authorize / identity / client-credentials / MFA-completion paths. Those create their session in the same request, making the race near-impossible, but if it ever happens they now settle at a 400 instead of a 500. They keep the error's own code rather than an OAuth2 grant error, since `invalid_grant` describes no grant artifact they hold.

## Verification

All three new tests were confirmed failing first, by stashing only the two behavioral source files. The 17 pre-existing grant tests stayed green throughout.

- `test/unit/core/oauth2/grant-types/refresh-token.spec.ts` — the missing-session error becomes `invalid_grant` from either issuer, with no second family revoke and no audit event or metric; an unrelated issuer failure is rethrown unchanged.
- `test/unit/app/modules/oauth2/session-token-repository.spec.ts` — new spec over an isolated in-memory database, deleting the session and triggering the real foreign-key violation.
- `test/unit/adapters/database/driver-error.spec.ts` — the code matrix.

Full server-core suite green (192 files, 2141 tests). Build, `check:types` and eslint clean.

The driver codes are the part a green suite cannot vouch for on its own, so they were checked against the running containers rather than assumed: postgres reports `23503`, mysql reports `ER_NO_REFERENCED_ROW_2` on insert and `ER_ROW_IS_REFERENCED_2` on delete, and sqlite's `SQLITE_CONSTRAINT_FOREIGNKEY` is exercised for real by the adapter spec.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved refresh-token handling when a session is deleted during token issuance.
  * Returns a standard OAuth `invalid_grant` response instead of exposing a database error.
  * Preserves existing behavior for unrelated failures and prevents replay side effects.

* **Tests**
  * Added coverage for database constraint detection, deleted-session token creation, and refresh/access-token issuance failures.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -3002,12 +3002,12 @@ old jti`) then AT (`refreshTokenId = new RT jti`). On consume-failure →
 `sessionId` is present (M2M client-credentials writes only an access row —
 it mints no RT).
 
-**Nothing holds the session between resolving it and issuing.** A concurrent
-request may delete the `auth_sessions` row after `findOneById` and before an
-issuer writes its inventory row: the replay reaction on this same family (two
-requests presenting one token, the loser calling `revokeFamily` while the winner
-is still issuing), an explicit logout, `DELETE /sessions/:id`, an admin
-force-logout, or the sweeper. The insert then violates
+**Nothing holds a token row's parents between resolving them and issuing.** A
+concurrent request may delete the `auth_sessions` row after `findOneById` and
+before an issuer writes its inventory row: the replay reaction on this same
+family (two requests presenting one token, the loser calling `revokeFamily`
+while the winner is still issuing), an explicit logout, `DELETE /sessions/:id`,
+an admin force-logout, or the sweeper. The insert then violates
 `FK_auth_session_tokens_session_id`. Refusing is correct, but the refusal has to
 be an OAuth2 answer: the presented token is already consumed and
 cache-blocklisted, so the client has nothing to retry with, and a 500 reads as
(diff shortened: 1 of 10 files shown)


tada5hi:
@coderabbitai pause

PR #3447, opened 2026-08-13
fix(server-core): declare a sort allow-list for the client-scope schema

### LINKED ISSUE #3441: clientScopeSchema declares no sorts allow-list, so any ?sort= key reaches ORDER BY
`clientScopeSchema` (`apps/server-core/src/core/entities/client-scope/schema.ts`) is the only one of the 26 registered schemas with no `sorts` block. rapiq falls back to a syntactic name check when a sort allow-list is undeclared, so an arbitrary root sort key survives decode and is handed to the TypeORM adapter as an `ORDER BY`.

Verified empirically against the real codec, with `clientSchema` as the control:

```
client-scope  {"sort":"name"}                => sorts: [{"name":"name","operator":"ASC"}]
client-scope  {"sort":"totallyBogusColumn"}  => sorts: [{"name":"totallyBogusColumn","operator":"ASC"}]
client (ctl)  {"sort":"name"}                => sorts: []
client (ctl)  {"sort":"totallyBogusColumn"}  => sorts: []
```

So `GET /client-scopes?sort=totallyBogusColumn` reaches `ORDER BY clientScope.totallyBogusColumn`, the driver rejects it, and `sanitizeError` maps it to `INTERNAL_ERROR` — a 500 on an unauthenticated-shaped input, where every sibling endpoint fails soft and returns unsorted rows.

Pre-existing (the schema never had a `sort` block), but #3425 swept `sorts: { allowed, indexed: true }` onto all 25 siblings and this one was missed, so it is now the sole outlier.

### Fix

Add the block with the matching `indexes` entries, following the sibling junction schemas:

```ts
sorts: { allowed: ['id', 'createdAt', 'updatedAt'], indexed: true },
```

`client-scope`'s `indexes` currently declares no `createdAt` / `updatedAt` sequences, so those need adding too (or the allow-list trimmed to what is backed) to satisfy the indexed invariant — `test/unit/core/query/indexed-invariant.spec.ts` will say which.

### Coverage

Worth a case asserting an unknown sort key is *stripped* rather than executed, for every registered schema — the invariant spec walks them already and is the natural home. A per-endpoint test would leave the next schema that forgets the block equally unguarded.

### LINKED ISSUE #3443: Sessions subject-kind filter keeps the previous page offset, so filtering from page 2+ shows an empty table
`apps/client-admin-console/pages/sessions/index/index.vue:101-107`:

```ts
const applySubjectKind = (load: (input: IQuery) => Promise<unknown>) => {
    if (subjectKind.value === 'all') {
        return load(defineQuery<Session>({ filters: {} }));
    }
    return load(defineQuery<Session>({ filters: { subKind: subjectKind.value } }));
};
```

The load carries no `pagination`, and the collection manager merges the current `meta` pagination when a load omits it, so the offset of the previously viewed page is retained. Paging to page 3 and then switching the subject-kind filter requests rows past the end of the freshly filtered set and renders an empty table with a non-zero total — the user has to page back to see anything.

`ASearch` gets this right by resetting the offset on each search, which is the shape to copy.

### Fix

```ts
return load(defineQuery<Session>({
    filters: { subKind: subjectKind.value },
    pagination: { offset: 0 },
}));
```

Both branches need it, including the `all` reset.

tada5hi:
Closes #3441

The sessions-page half of this PR (#3443) landed separately via #3462, so that commit was dropped in a rebase onto master. What remains is the client-scope fix plus the kit spec that pins the offset-reset contract at the collection manager (the place a regression would reappear; #3462 shipped without one).

## `#3441` — `clientScopeSchema` declared no sort allow-list

It was the sole outlier among the 26 registered schemas (missed by the `#3425` sweep). Without the block rapiq falls back to a syntactic name check, so `GET /client-scopes?sort=totallyBogusColumn` reached `ORDER BY clientScope.totallyBogusColumn` and surfaced as a **500**, where every sibling endpoint fails soft and returns unsorted rows.

**The allow-list is wide on purpose.** `default`, `clientId`, `scopeId`, `clientRealmId` and `scopeRealmId` all sort correctly *today* through the missing-allow-list fallback and are all already indexed, so the narrow junction-sibling triple would have demoted them from working-and-sorted to silently unsorted. `createdAt` / `updatedAt` were the only listed keys with no backing index, which is what the migration adds.

### Coverage

The new case is per schema, not per endpoint: a per-endpoint test would leave the next schema that forgets the block equally unguarded. It decodes `{ sort: 'totallyBogusColumn' }` through the real codec for all 26 schemas and asserts the key is stripped.

Verified differentially: with the schema change reverted the suite reports exactly one failure (`should strip an unknown sort key for clientScope`); with it, the cases pass.

### Kit spec: the offset reset a filter load has to carry

`entity-collection.spec.ts` drives the manager to page 3, then hands it the shape a filter control uses (an assembled Query carrying `filters` plus `pagination: { offset: 0 }`) and asserts the request returns to the first page with the retained page size intact. Dropping the reset from the load input makes the request carry offset 20, the #3443 bug.

### Verification (after the rebase)

- `apps/server-core` full suite: **196 files / 2214 tests passed**; `indexed-invariant.spec.ts` + `console-search-surface.spec.ts`: 70 passed; kit `entity-collection.spec.ts`: 16 passed
- postgres round trip on a scratch DB: `migration run` / `revert` / `run` clean, `test:schema-drift` matches; `1786631686318-ClientScopeSortIndexes` is the newest migration (master's `1786436332251-QueryIndexes` shipped in beta.60/61, so a new file is correct)
- generated DDL touches `auth_client_scopes` only, and neither column is a foreign key, so the MySQL implicit-FK-index `down()` trap does not apply
- `check:types` clean, eslint clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved client-scope sorting performance for creation and update dates.
  * Prevented unsupported sort fields from causing database errors or server failures.
  * Fixed pagination so resetting the offset returns to the first results while preserving the selected page size.
* **Tests**
  * Added regression coverage for invalid sort fields across schemas and pagination resets.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -362,6 +362,20 @@ usable at the service level and nothing in core depends on TypeORM:
   ran. It now carries a *should have sort allow-lists to check* guard;
   keep that guard, and prefer it whenever a spec derives its subject
   from an upstream shape.
+  **A missing `sorts` block is not fail-soft, it is a 500.** With no
+  sort allow-list rapiq falls back to a syntactic name check, so an
+  arbitrary root key survives decode and reaches `ORDER BY` on a column
+  that does not exist; the driver rejects it and `sanitizeError` maps
+  that to `INTERNAL_ERROR`, where every declaring sibling answers with
+  unsorted rows. `clientScope` was the sole schema without the block
+  (#3441, missed by the #3425 sweep). The property is now pinned per
+  schema rather than per endpoint — *should strip an unknown sort key
+  for %s* decodes a bogus key through the real codec for all 26 — so
+  the next schema that forgets the block fails the suite instead of one
+  endpoint. Note the fix direction: the narrow sibling list would have
+  demoted `default`/`clientId`/`scopeId` and the two owner-realm keys
+  from working-and-sorted to silently unsorted, so client-scope
+  allow-lists every column it already indexes.
   Structures whose leading column is not queryable (session
   `ipAddress`/`userAgent`, `user.email`) stay undeclared on purpose:
   the declaration describes the query surface, not the whole table.
--- a/apps/server-core/src/adapters/database/domains/client-scope/entity.ts
+++ b/apps/server-core/src/adapters/database/domains/client-scope/entity.ts
@@ -42,9 +42,11 @@ export class ClientScopeEntity implements ClientScope {
 
     // ------------------------------------------------------------------
 
+    @Index()
     @CreateDateColumn({ name: 'created_at', transformer: dateToISOStringTransformer })
     createdAt: string;
 
+    @Index()
     @UpdateDateColumn({ name: 'updated_at', transformer: dateToISOStringTransformer })
     updatedAt: string;
 
--- a/apps/server-core/src/adapters/database/migrations/mysql/1786631686318-ClientScopeSortIndexes.ts
+++ b/apps/server-core/src/adapters/database/migrations/mysql/1786631686318-ClientScopeSortIndexes.ts
@@ -1,0 +1,31 @@
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * Index `auth_client_scopes.created_at` and `updated_at`.
+ *
+ * `clientScopeSchema` was the only registered schema with no `sorts`
+ * allow-list, so rapiq fell back to a syntactic name check and handed an
+ * arbitrary `?sort=` key to `ORDER BY` (issue #3441). Declaring the
+ * allow-list requires every listed key to lead a real entity index; the
+ * two timestamp columns were the only ones missing theirs.
+ */
+export class ClientScopeSortIndexes1786631686318 implements MigrationInterface {
+    name = 'ClientScopeSortIndexes1786631686318';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            CREATE INDEX \`IDX_52f59535945c8cdbd62439bba5\` ON \`auth_client_scopes\` (\`created_at\`)
+        `);
+        await queryRunner.query(`
+            CREATE INDEX \`IDX_04264aebc8b6625b1da88e23df\` ON \`auth_client_scopes\` (\`updated_at\`)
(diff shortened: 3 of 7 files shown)


tada5hi:
## Audit outcome

Reviewed adversarially against the runtime rather than the diff. **Verdict: sound, no defects.**

The reviewer executed the real `@rapiq/core` + `@rapiq/codec-url` from `node_modules` rather than reasoning about them:

- **#3443** — reproduced the composition chain against this page's actual base query with state pagination `{limit:10, offset:20}`. The old input shape yields `{"limit":10,"offset":20}` (the bug), the new one `{"limit":10,"offset":0}`. Omitting `limit` does retain the page size (`Pagination.merge` is per-property), and `offset: 0` survives because `0` is not nullish.
- **#3441** — against master's schema shape, `?sort=totallyBogusColumn` decodes to `{"name":"totallyBogusColumn","operator":"ASC"}` and reaches `qb.orderBy()` on a nonexistent column; on this branch it decodes to `[]`. Every allow-listed key leads a declared index backed by a real entity structure. The migration's index names were confirmed to be genuine typeorm hashes (`sha1('auth_client_scopes_created_at').slice(0,26)`), so the drift gate stays green.

### Acted on since

- **Added the missing test for #3443** (`entity-collection.spec.ts`). The audit noted the fix shipped unpinned and the manager is where a regression would reappear. Verified by dropping the reset from the load input: the request then carries offset 20, the reported bug.
- **Recorded in the schema itself** why the client-scope sort allow-list is wider than its junction siblings'.

### Noted, not changed

- A multi-key sort on client-scope with no matching composite prefix now drops whole-parameter. That is the documented fail-soft narrowing every sibling schema already carries.
- The subject-kind select drops a change made while a load is in flight (`load` early-returns on `busy`). Pre-existing and identical on master; `ASearch` debounces, this select does not.
- `ClientScopeService.getMany` has no per-row realm gate, so a `CLIENT_READ` holder sees other realms' rows. Pre-existing on master and unrelated to this change, but it is what makes "sorting exposes nothing new" true here.

PR #3448, opened 2026-08-13
fix(server-core): derive the federated identity from id_token and userinfo claims

### LINKED ISSUE #3434: Federated users are provisioned with the remote subject UUID as their name
## Observed

Logging in through an external OAuth2/OIDC identity provider creates a local user whose `name` is the remote subject identifier. Against another authup instance that identifier is a UUID, so the provisioned user looks like:

```
name:  54e8b720-cc1f-453c-978c-7d31bab52949
email: 54e8b720-cc1f-453c-978c-7d31bab52949@example.com   (buildUserFakeEmail placeholder)
```

The same value is copied into `auth_identity_provider_accounts.provider_user_name`. Every federated user in the admin console is therefore identified by an opaque UUID, and the placeholder email is neither routable nor recognisable.

Observed end-to-end with authup federating to authup (server-core 1.0.0-beta.59).

## Cause

`IdentityProviderOAuth2Authenticator.buildIdentityWithTokenGrantResponse` (`apps/server-core/src/core/identity/provider/authentication/protocols/oauth2/module.ts`) derives the identity from the **access token** alone:

```ts
const payload = extractTokenPayload(input.access_token);

return {
    id: payload.sub!,
    attributeCandidates: {
        name: [payload.sub],
        email: [payload.email],
    },
    data: payload,
    provider: this.provider,
};
```

Two richer sources are available at that point and both are ignored:

- **`userInfoUrl`** is already configured onto the hapic client as `userinfoEndpoint`, and `@hapic/oauth2` ships a `UserInfoAPI`, but nothing ever calls it. Setting the field on a provider today has no effect on the login path.
- **`id_token`**, present on the grant response for any `openid`-scoped OIDC provider, carries `preferred_username`, `name` and `email` by convention. `TokenGrantResponse.id_token` is discarded.

An authup access token carries `sub`, `kind`, `realm_name` and friends — no username, no email. So for authup-to-authup federation the only candidate offered is a UUID, and `email` is always absent.

## Impact

Cosmetic rather than a security problem, but it makes federated users hard to administer: the console lists UUIDs, and the fake email means notification and password-recovery paths address a non-existent mailbox. It also makes `userInfoUrl` misleading, since it is configurable in the admin UI and silently inert on login.

## Possible directions

1. Prefer the `id_token` claims when the grant response carries one, falling back to the access token. Cheapest, no extra request, covers every OIDC provider.
2. Call the userinfo endpoint when `userInfoUrl` is set, and merge its claims into `attributeCandidates`. Makes the existing field do what it says.
3. Widen `attributeCandidates` to the usual claim ladder (`preferred_username`, `nickname`, `name`, `email`, `sub`) so `pickCandidate` picks the best available rather than only `sub`.

(1) and (3) together would fix the common case with no additional round trip; (2) is the general answer for plain OAuth2 providers that issue no id_token.

Note the attribute mappers (`IIdentityProviderMapper`) do not close this gap on their own: they map over `identity.data`, which is the access-token payload, so there is nothing richer to map from until one of the above lands.

tada5hi:
Closes #3434.

## The defect

An external identity was built from the **access token alone**, which is opaque by contract (OIDC Core §2). Authup's own access token carries `sub`, `kind` and `realm_name` and no username, so federating authup to authup provisioned every local user under the remote subject UUID plus a `<uuid>@example.com` placeholder.

## Why the issue's direction (1) alone would not have fixed it

Reading the `id_token` is necessary but **not sufficient**. Authup's `OAuth2OpenIDClaimsBuilder` maps *both* `preferred_username` and `nickname` onto the **nullable** `displayName`, and puts the real username in `name`. The previous ladder was `preferred_username, nickname, sub`, so against an authup id_token it still fell through to `sub`.

The ladder is now `preferred_username, nickname, name, sub`. `preferred_username` stays first because Keycloak and Authentik put the username there while their `name` is a display name. A candidate that fails `isUserNameValid` shifts to the next one, so a bad first candidate degrades rather than failing the login.

## userinfo

Called only when the provider declares a `userInfoUrl`; failures are logged and swallowed. The guard is load-bearing — the hapic client is constructed with no `baseURL`, so the `/userinfo` default would be a relative fetch URL that throws.

This is the only fix for a plain `protocol: oauth2` provider whose remote client has no `openid` scope bound (no `openid`, no id_token), which is a plausible reading of the reported config, and it is what stops the admin-editable `userInfoUrl` being silently inert.

## What deliberately did not change

- **`identity.id` stays the ACCESS token `sub`.** It is `provider_user_id` and keys `findOneByProviderIdentity`, so sourcing it from a richer claim set would orphan every existing `auth_identity_provider_accounts` row and silently provision duplicate users on the next login. Pinned by a dedicated case.
- **`identity.data` stays the access-token payload.** `IIdentityProviderMapper` pattern-matches over it, so merging richer claims in would change every existing mapper's input. Separable, and a product call rather than a bug fix.
- **Forward-only.** The account manager's UPDATE branch never rewrites `user.name` (`nameLocked` at creation), so users already provisioned under a UUID keep it.
- The five presets (github, facebook, instagram, paypal, google) override the builder themselves and are untouched. The OIDC subclass override became identical to the base and is deleted.

## Coverage

`buildIdentityWithTokenGrantResponse` had **zero** coverage — the existing authenticator spec only tested `buildRedirectURL` and scope merging. Added 8 unit cases (subject fallback, the authup `name` shape, Keycloak ordering, the account-key invariant, an undecodable JWE id_token, the userInfoUrl guard, userinfo precedence, userinfo failure degrading) plus 2 end-to-end cases where the fake IdP now returns an authup-shaped access token with an id_token beside it.

Verified differentially: with the source fix reverted, **8 of the 23 fail**; with it, all 23 pass. (The first attempt at this check was itself wrong — the appended block had landed in the wrong working copy and the unit tests were never running. The differential is what caught that.)

- `apps/server-core` full suite: **194 files / 2170 tests passed**
- eslint clean, `apps/server-core` build green

## Summary by CodeRabbit

- **New Features**
  - Federated sign-in now builds user identities from access-token claims, ID-token claims, and optional userinfo.
  - Usernames and email addresses use the most complete available claims, with clear fallback behavior.
  - Existing account identifiers remain stable and are based on the access-token subject.

- **Bug Fixes**
  - Invalid, encrypted, or unavailable identity tokens no longer block authentication.
  - Userinfo service failures are handled gracefully while allowing sign-in to continue.

- **Documentation**
  - Added documentation describing federated identity claim handling and account stability.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -2654,6 +2654,50 @@ Domain type `Consent` (core-kit) + `EntityType.CONSENT`, TypeORM entity +
   `isUniqueConstraintDatabaseError` (`adapters/database/errors/driver.ts`, the
   reusable driver-error-code unwrapper covering mysql/postgres/sqlite).
 
+## Federated Identity Claims
+
+`IdentityProviderOAuth2Authenticator.buildIdentityWithTokenGrantResponse`
+derives the external identity from three sources, richest last:
+**access token, then `id_token`, then userinfo**. The access token is only
+the floor because it is opaque by contract (OIDC Core §2) and authup's own
+carries `sub`, `kind` and `realm_name` and no username at all, so reading
+it alone provisioned every federated user under the remote subject UUID
+plus a `<uuid>@example.com` placeholder (#3434).
+
+- **The name ladder is `preferred_username, nickname, name, sub`.**
+  Keycloak and Authentik put the username in `preferred_username` while
+  their `name` is a display name; authup's own claims builder maps BOTH
+  `preferred_username` and `nickname` onto the **nullable** `displayName`
+  and puts the real username in `name`, so a ladder without `name` still
+  falls through to `sub` for authup-to-authup federation. Candidates are
+  consumed reactively — `validateAttributes` only substitutes one when the
+  validator raises an issue for that exact path — so a first candidate
+  failing `isUserNameValid` degrades to the next rather than failing the
+  login, and adding candidates for OPTIONAL keys (`firstName`/`lastName`)
+  would be dead code.
+- **`identity.id` stays the ACCESS token `sub`, always.** It becomes
+  `provider_user_id` and keys `findOneByProviderIdentity`, so sourcing it
+  from a richer claim set would orphan every existing
+  `auth_identity_provider_accounts` row and silently provision duplicate
+  users on the next login. Pinned by *should key the account on the access
+  token subject, never the id_token*.
+- **userinfo is called only when the provider declares `userInfoUrl`**, and
+  its failure is logged and swallowed. The guard is load-bearing: the hapic
+  client is constructed with no `baseURL`, so the `/userinfo` default would
+  be a relative fetch URL that throws. It is the only fix for a plain
+  `protocol: oauth2` provider whose remote client has no `openid` scope
+  bound (no `openid` scope, no id_token), and it is what stops the
+  admin-editable `userInfoUrl` field being silently inert.
+- **`identity.data` is deliberately still the access-token payload alone.**
+  Mappers (`IIdentityProviderMapper`) pattern-match over it, so merging the
+  richer claims in would change every existing mapper's input.
+- The `IdentityProviderOpenIDAuthenticator` override is gone: the base
+  ladder now covers what it did. The five presets (github, facebook,
+  instagram, paypal, google) still override the builder and are untouched.
+- **Forward-only.** The account manager's UPDATE branch never rewrites
+  `user.name` (it is `nameLocked` at creation), so users already
+  provisioned under a UUID keep it.
+
 ## Identity-Provider Account Linking (plan 091)
 
 `auth_identity_provider_accounts` (external identity → `userId`; unique
--- a/apps/server-core/src/adapters/http/controllers/entities/identity-provider/module.ts
+++ b/apps/server-core/src/adapters/http/controllers/entities/identity-provider/module.ts
@@ -680,6 +680,7 @@ export class IdentityProviderController {
         return createIdentityProviderOAuth2Authenticator({
             accountManager: this.accountManager,
             provider,
+            logger: this.logger,
             options: {
                 baseURL: this.options.baseURL,
                 clientId: options.clientId,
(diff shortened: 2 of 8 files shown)


tada5hi:
@coderabbitai pause

tada5hi:
## Audit outcome: the fix did not work, and now does

The adversarial review of this branch returned **has-defects**, and it was right. Three items, all fixed in `e4ab21b`.

### The fix was silently disabled for the users it targets

`extractTokenPayload` decoded with `atob`, which **rejects the base64url alphabet** (`-`/`_`) and yields latin1 where a JWT payload is UTF-8. I verified it independently on node 24:

```
ascii only           | base64url chars: no  | decoded OK
emoji displayName    | base64url chars: yes | THROWS: Invalid character
cyrillic             | base64url chars: yes | THROWS: Invalid character
cjk                  | base64url chars: yes | THROWS: Invalid character
```

An authup id_token carries the display name under both `preferred_username` and `nickname`, so a user whose display name is not ASCII produced exactly that payload, the id_token decode threw, the bare `catch {}` swallowed it with no log, the ladder fell through to `sub`, and the user was provisioned under the remote UUID again. The reported defect, now invisible.

Fixed in `@authup/server-kit` rather than worked around here, which also repairs the pre-existing **hard** failure on the same decode of the access token. The new decode is pinned by a differential comparison against the platform's own base64url decode (96/96 samples identical), and the catch now logs.

### Two more from the same review

- **No timeout on userinfo.** hapic passes no `signal` to `fetch`, so an endpoint that accepts the connection and never answers would have held the login for undici's 300s headers timeout, contradicting this branch's own "enrichment, never a login blocker" comment. Bounded at 5s.
- **The userinfo `sub` was not checked.** OIDC Core 5.3.2 requires a response whose `sub` does not match the token's to be discarded; this branch merged it unconditionally, and **the integration test enshrined the violation** (the fake IdP answered `external-user-2` for a token saying `external-user-1`, and the test asserted the mismatched name was used). A mis-routed response would have named and, worse, **emailed** the local user after somebody else, and `auth_users.email` has no unique constraint, so that third party could then drive password recovery on the federated account. The check is in, the fake IdP now answers with a matching subject, and the mismatch is pinned as its own case.

### Verified clean

`identity.id` is the access-token `sub` on every path (pinned); the five presets override the builder and never reach `resolveClaims`; `identity.data` is unchanged for mappers; the deleted OIDC override changes nothing beyond the ladder; the logger reaches the single construction site.

Suite: 194 files / 2172 tests. Differential: reverting the decode fix fails the non-ASCII case and nothing else.

PR #3449, opened 2026-08-13
fix(server-core): make one external identity belong to one local user

### LINKED ISSUE #3442: auth_identity_provider_accounts has no unique constraint on (provider_id, provider_user_id)
`IdentityProviderAccountManager.link()` (`apps/server-core/src/core/identity/provider/account/module.ts:99-119`) enforces "one external identity belongs to one local user" with a plain `findOneByProviderIdentity` followed by `repository.save` — no transaction, no row lock, and nothing backing it in the schema.

The entity declares:

```ts
@Index(['providerId', 'userId'], { unique: true })   // one link per provider per user
@Index(['providerUserId', 'providerId'])             // NOT unique
```

So the invariant the `IdentityProviderAccountAlreadyLinkedError` guard exists to protect is application-only. Two concurrent completions for the same external identity toward two different local users can both read "not linked" and both insert, after which the same upstream subject resolves to two accounts and `findOneByProviderIdentity` returns whichever the database happens to order first.

The federated login path has the same shape via `save()`.

### Fix

Add the unique index and generate the migration:

```ts
@Index(['providerUserId', 'providerId'], { unique: true })
```

Per the repo convention, DDL comes from `migration generate` rather than hand-authored — and note the MySQL implicit-FK-index trap documented in `.agents/conventions.md` if the generated `down()` drops an index a foreign key is relying on.

Existing deployments may already hold duplicate rows, so the migration needs a pre-check (or a documented cleanup) before the constraint can be created — a `SELECT provider_id, provider_user_id, COUNT(*) … HAVING COUNT(*) > 1` guard in the same file, following the precedent in `1779267068441`.

`link()` can keep its explicit check for the friendly `already_linked` message; the constraint is the backstop, and the duplicate-key catch already exists in `ConsentRepositoryAdapter` (`isUniqueConstraintDatabaseError`) as a model.

tada5hi:
Closes #3442.

`IdentityProviderAccountManager.link()` enforced "one external identity belongs to one local user" with a `findOneByProviderIdentity` followed by a `save`: no transaction, no row lock, and nothing behind it in the schema. Two concurrent completions for the same upstream subject could both observe "not linked" and both insert, after which `findOneByProviderIdentity` returns whichever row the database happens to order first. The federated login path has the same shape through `save()`.

## The column order is kept, deliberately

The index is flipped **in place** to unique as `(provider_user_id, provider_id)`, not reordered to match the issue title. Reordering would break three things at once: the rapiq schema declares `['providerUserId', 'providerId']` and would lose its backing leftmost prefix (boot-time `assertSchemaMatchesEntity` failure), `providerUserId` would stop anchoring as a filter key, and the derived index name would change so the migration became a rename rather than a uniqueness flip.

## Pre-existing duplicates

The migration checks first and aborts with the number of affected groups. It would abort either way — `CREATE UNIQUE INDEX` fails on duplicates — so the check only decides whether the operator reads a readable sentence or a hash-named driver error. Follows the `1779267068441` precedent. `docs/src/guide/deployment/upgrading.md` carries the note and the query to find them.

That file's top section was still headed *Next release (after v1.0.0-beta.59)* although those entries shipped in beta.60 (verified with `git tag --contains`), so it is corrected to `## v1.0.0-beta.60` and a new *Next release* section opened above it.

## What is not in this change

No adapter-level translation of the driver error into `IdentityProviderAccountAlreadyLinkedError`. `save()` also serves the federated login, where a lost race is not "already linked"; the honest answer there is to re-read and return the winner's row, which is a larger change. Without translation the race loser gets `?linkError=link_failed` — correct, just less specific.

## Verification

- `test:migration-latest` (populated round-trip: drift, revert, re-run, FK cascade, reboot): **12/12 on postgres, 12/12 on mysql**
- **The generated postgres `down()` was corrected.** It recreated the index as `(provider_id, provider_user_id)`, but the released `1786436332251-QueryIndexes` created it as `(provider_user_id, provider_id)`, so as emitted a revert would have left a *different* index than existed before. Verified empirically: after `migration revert`, `pg_indexes` reports `USING btree (provider_user_id, provider_id)`, byte-identical to the released chain. This is the documented exception to committing generated DDL untouched — the generated statement was wrong, and the correction is pinned by observation rather than by reading.
- MySQL `down()` needs no FK-constraint wrapper: the index is being replaced under the same name rather than dropped, and the round-trip confirms it.
- New case *should refuse a second account for the same external identity* in `account.spec.ts`. It points the duplicate row at a user holding **no** account for that provider — the obvious version of this test reuses an already-linked user, which violates the pre-existing `(providerId, userId)` index instead and passes with or without the constraint under test. Caught by running the differential: the first version passed against a reverted entity.
- Differential: with the entity change reverted the new case fails; with it, 9 pass.
- `apps/server-core` full suite: **194 files / 2161 tests passed**; eslint clean.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -2656,8 +2656,18 @@ Domain type `Consent` (core-kit) + `EntityType.CONSENT`, TypeORM entity +
 
 ## Identity-Provider Account Linking (plan 091)
 
-`auth_identity_provider_accounts` (external identity → `userId`; unique
-`(providerId, userId)` — one link per provider per user) is surfaced as a
+`auth_identity_provider_accounts` (external identity → `userId`) carries TWO
+unique indexes: `(providerId, userId)` — one link per provider per user — and
+`(providerUserId, providerId)` — one external identity belongs to one local
+user (#3442). The second was application-only until then: `link()` and
+`save()` both read then write with no transaction and no row lock, so two
+concurrent completions for the same upstream subject could both insert, after
+which `findOneByProviderIdentity` returned whichever row the database ordered
+first. The column ORDER of that index is load-bearing — `providerUserId`
+leads the sequence the rapiq schema declares, so reversing it would fail
+`assertSchemaMatchesEntity` at boot and turn the migration into a rename.
+`link()` keeps its explicit check for the friendly `already_linked` message;
+the constraint is the backstop. The table is surfaced as a
 read+delete entity API and an explicit self-service link flow. No migration:
 the table predates the feature (federated login has always written it, and
 still auto-creates a NEW user for an unknown external identity — explicit
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -2668,10 +2668,12 @@ leads the sequence the rapiq schema declares, so reversing it would fail
 `assertSchemaMatchesEntity` at boot and turn the migration into a rename.
 `link()` keeps its explicit check for the friendly `already_linked` message;
 the constraint is the backstop. The table is surfaced as a
-read+delete entity API and an explicit self-service link flow. No migration:
-the table predates the feature (federated login has always written it, and
-still auto-creates a NEW user for an unknown external identity — explicit
-linking is the only way to bind an external identity to an EXISTING user).
+read+delete entity API and an explicit self-service link flow. The LINKING
+FEATURE itself needed no migration (the table predates it: federated login
+has always written it, and still auto-creates a NEW user for an unknown
+external identity, so explicit linking is the only way to bind an external
+identity to an EXISTING user); the uniqueness flip above does ship one,
+`1786633352004-IdentityProviderAccountUniqueness` on both dialects.
 
 - **Unified repository port** — `IIdentityProviderAccountRepository`
   (`core/entities/identity-provider-account/types.ts`) carries the entity
--- a/apps/server-core/src/adapters/database/domains/identity-provider-account/entity.ts
+++ b/apps/server-core/src/adapters/database/domains/identity-provider-account/entity.ts
@@ -29,7 +29,10 @@ import { IdentityProviderEntity } from '../identity-provider/index.ts';
 
 @Entity({ name: 'auth_identity_provider_accounts' })
 @Index(['providerId', 'userId'], { unique: true })
-@Index(['providerUserId', 'providerId'])
+// one external identity belongs to one local user. The column order is
+// load-bearing: `providerUserId` leads the sequence the rapiq schema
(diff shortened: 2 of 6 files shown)


tada5hi:
## Audit outcome

**Verdict: sound-with-nits.** The constraint is semantically correct and no existing flow can produce a duplicate. The reviewer traced all four writers: federated OAuth2/OIDC login (`save()` takes the UPDATE branch on a repeat login), LDAP (same DN under two providers yields two rows with different `provider_id`, which the index permits), provisioning (writes no account rows at all), and `link()` (short-circuits with `already_linked`). Cross-realm duplication is impossible because `auth_identity_providers.realm_id` is NOT NULL and `link()` rejects a realm mismatch.

Also confirmed independently: `assertSchemaMatchesEntity` still passes (a `unique: true` `@Index` stays in `metadata.indices`, and typeorm's name hash excludes uniqueness), the pre-check cannot false-negative on either dialect (both columns NOT NULL, grouping collation equals index collation), and the new test genuinely isolates the new constraint.

### Fixed since (`58492a6`)

**MySQL DDL is not transactional**, so `migrationsTransactionMode: 'all'` cannot roll a partial run back. `DROP INDEX` followed by `CREATE UNIQUE INDEX` leaves the table with **no index of that name** if the process dies between them (pod eviction during the build, connection reset), typeorm never records the migration, and every later boot then fails at the DROP with `ER_CANT_DROP_FIELD_OR_KEY` until someone hand-recreates a hash-named index.

Now one atomic `ALTER TABLE … DROP INDEX …, ADD UNIQUE INDEX …`. Postgres keeps the generated statements; its DDL is transactional. Re-verified: mysql populated round-trip 12/12, and `information_schema` reports `NON_UNIQUE=0` over `(provider_user_id, provider_id)`.

Also re-scoped the architecture note's "No migration" clause, which sat under the new uniqueness paragraph and read as a claim about it.

### Deliberately not changed

The losing side of a genuinely concurrent first federated login now gets a 409 rather than silently writing a second account row. Resolving it onto the winner (catch the unique violation, re-read via `findOneByProviderIdentity`) is the better end state, but it is a change to the login write path rather than to this constraint, and the current outcome is self-healing on retry and strictly better than the silent duplicate it replaces. Worth its own issue.

tada5hi:
Superseded by #3463, which folds in this PR's migration (atomic MySQL swap, corrected postgres down(), pre-check), the entity comment and the upgrading.md note, and adds the write-path conflict handling plus the populated-round-trip probe. Closing to keep one PR per issue.

PR #3450, opened 2026-08-13
feat(server-core): complete an identity-provider link on an authenticated request

### LINKED ISSUE #3439: Move the identity-provider account link behind a bearer-authenticated confirmation
## Context

The account-linking callback (plan 091) performs a durable credential binding from an **unauthenticated** request. The browser round-trip cannot carry a bearer, so the linking user's id is carried in the OAuth2 state blob (`link: { userId, providerId }`) and the callback trusts it.

The only thing distinguishing "the browser that started this link" from any other browser is the state's ip / user-agent binding. That binding is weak by construction:

- `OAuth2AuthorizationStateManager.verify` (`core/oauth2/authorization/state/module.ts:35-47`) skips each check when the **stored** value is falsy.
- Both stored values come from the request that minted the state (`saveAuthorizationState`: `getRequestIP(event) ?? ''`, `getRequestHeader(event, 'user-agent') ?? undefined`).
- The shipped default is `trustProxy: true`, which per `.agents/architecture.md` returns the "left-most, client-supplied entry" of `X-Forwarded-For`.

State unguessability does not help here. The attacker is not guessing the victim's state; they supply their own and want the victim to use it.

**Attack:** attacker mints a link state carrying their own `userId`, sends the upstream authorize URL to a victim, and the victim's callback binds the **victim's** external identity to the **attacker's** account. The victim's next "Sign in with \<provider\>" then resolves through `IdentityProviderAccountManager.save()` and lands in the attacker's account.

A stopgap landed in #3440: the callback refuses a state with an absent ip or user-agent binding. That closes the trivial bypass (omit the UA, blank the IP) but leaves the values guessable — behind corporate NAT with a common browser string, ip + user-agent is not much of a bar.

## Proposal

Stop completing the link in the callback. Split it:

1. `authorize-in` resolves the external identity as today, then stashes it under a short-lived one-time handle instead of writing the account row.
2. It redirects to the account console with that handle.
3. The account console `POST`s a confirmation **with its bearer**, and the server links the stashed identity to the *authenticated* user.

The credential-binding write then happens on an authenticated request, the state blob no longer needs to carry a `userId`, and the ip/user-agent heuristic stops being load-bearing for this path. This also matches the reasoning that produced the current design: the userId is in the state precisely because the callback has no bearer, and this removes that need rather than hardening around it.

The stopgap guard in `completeLink` can be dropped when this lands.

### Alternative considered

A `__Host-` cookie nonce set at `link-request` and required at the callback is the standard state-cookie pattern and is genuinely unforgeable cross-site. It is rejected as the primary fix because a standalone-hosted account console (a supported topology) calls `link-request` cross-origin, so the cookie would be third-party there.

## Notes

- Login-path (`authorize-out`) exposure to the same weakness is pre-existing and lower impact: session fixation, nothing persists. Worth revisiting separately.
- Not a takeover of the victim's authup account; the victim is fooled into using an account the attacker controls (federated account pre-hijacking).
- Regression coverage exists: *refuses to link on a state that carries no browser binding* in `test/unit/http/controllers/entities/identity-provider/link.spec.ts` fails when the guard is removed (verified: the link succeeds and `linked` is set).

tada5hi:
Closes #3439.

## The problem

The account-linking callback performed a **durable credential binding from an unauthenticated request**. The browser round-trip cannot carry a bearer, so the linking user's id rode the OAuth2 state blob and the callback trusted it. The only thing separating "the browser that started this link" from any other was the state's ip / user-agent binding, and both values are chosen by whoever minted the state (`verify` skips a check whose *stored* value is falsy; under the shipped `trustProxy: true` the ip is the client-supplied left-most `X-Forwarded-For` entry).

The attack is the reverse of the obvious one: an attacker mints a link request for their **own** account, gets a victim to follow it, and the victim's callback binds the **victim's** external identity to the **attacker's** account. The victim's next federated login then resolves into it. State unguessability defends nothing, because the attacker supplies the state rather than guessing it.

## The split

1. `link-request` is unchanged.
2. `authorize-in` resolves the external identity, **stashes** a four-scalar projection under a one-time handle, and redirects to `.../connected-accounts?linkHandle=…&provider=…`. It writes nothing.
3. The account console auto-POSTs the handle to the new `POST /identity-providers/:id/link-confirm` **with its bearer**, and the row is written for the *authenticated* user.

The #3440 stopgap (refusing a state with no browser binding) is deleted along with the hole it patched.

## Two decisions worth reviewing

**The stash keeps `userId`, and the confirm requires `stash.userId === authenticated user`.** The issue sketched dropping it, since the state "no longer needs to carry" it. That would make a *leaked handle redeemable by anyone* — and the handle rides a URL query parameter, so it reaches browser history, referers and proxy logs. That is the same pre-hijacking in a new shape. Keeping it is both stronger (two-sided: who started it *and* who is finishing it) and a smaller diff. The handle is additionally single-use and bound to its provider, so on its own it authorizes nothing.

**A dedicated 5-minute cache store, not a second `OAuth2AuthorizationStateManager` blob.** Reuse looks smaller but inherits the state's 30-minute TTL for a value redeemed one page load later, and re-applies the ip check to an XHR from the SPA, so a network flip between callback and confirm would break the link.

The stash holds four scalars rather than the `IdentityProviderIdentity`: that object carries the provider entity including the EA-loaded `clientSecret` plus the raw external token payload, neither of which belongs in a cache. They are also exactly what `link()` reads.

## Behavior changes

- **Return params**: `linked=<id>` → `linkHandle=<handle>&provider=<id>`. Only the bundled account console consumes them, but it is a documented URL contract for a self-hosted one, so it is in the upgrade notes.
- **`already_linked` surfaces on the confirm response** instead of a redirect marker, so the SPA shows the server's own localized error rather than a string it has to map.
- **The `identityProviderLinked` event moves to the confirm**, where the request context can supply actor and session attribution the unauthenticated callback could not.
- **The callback also rejects a subject-less identity** now. It used to reach `link()`, which refused it; without the guard an empty `providerUserId` would be stashed.
- **Deployment**: callback and confirm must reach the same cache, so a multi-replica deployment on the in-process memory cache needs sticky routing across one more hop. Redis removes it. Documented.

## Verification

- `link.spec.ts` rewritten to the two-step flow, now 13 cases: the callback writes **no row** before confirmation, the confirm links, idempotency, already-linked, **a handle minted for another user is refused**, no bearer → 401, replay → 400, wrong provider → 400.
- **Mutation-tested the guard that matters**: deleting `link.userId !== identity.id` from the confirm makes exactly *refuses a handle minted for another user* fail, and nothing else.
- `apps/server-core` full suite: **194 files / 2163 tests passed**; eslint clean; server-core, core-http-kit and account-console builds green.

Not addressed here (pre-existing, called out in the issue): the login path's `authorize-out` state has the same weak binding, and the callback's return target is hard-coded to `publicUrl`, so a standalone-hosted account console never receives the handle.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -2771,8 +2771,9 @@ linking is the only way to bind an external identity to an EXISTING user).
   id/name — never tokens).
 - **UI** — account console page `/connected-accounts` (realm providers ×
   own linked rows; Connect = `createLinkRequest` navigation, Disconnect =
-  confirm + delete; reads and strips the `linked`/`linkError` return
-  params) + admin console tab `users/[id]/identity-provider-accounts`
+  confirm + delete; auto-POSTs the `linkHandle` return param to the confirm
+  endpoint, then strips it and `linkError` from the URL)
+  + admin console tab `users/[id]/identity-provider-accounts`
   (kit collection `AIdentityProviderAccounts`, gated on
   `IDENTITY_PROVIDER_ACCOUNT_READ` like the sessions tab).
 - **Tests** — service matrix + guardrail
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -2698,44 +2698,74 @@ linking is the only way to bind an external identity to an EXISTING user).
   has no password (`IdentityProviderAccountUnlinkBlockedError`,
   `ErrorCode.IDENTITY_PROVIDER_ACCOUNT_UNLINK_BLOCKED`, 400): the row may
   be the only way into the account. An admin sets a password first.
-- **Link flow (server-completed)** — server auth is header-only, so the
-  browser round-trip cannot carry a bearer; the link intent rides the
-  one-time, IP+UA-bound `stateManager` blob instead:
-  `POST /identity-providers/:id/link-request` (ForceLoggedIn, user
-  identities only; provider must be OAuth2/OIDC, enabled, and in the
-  actor's realm) saves a state with the new optional
-  `link: { userId }` field (`OAuth2AuthorizationState`; cache blob, no
-  migration) and returns `{ url }` = the external authorize URL
-  (`IdentityProviderLinkRequestResponse` in core-http-kit;
-  `client.identityProvider.createLinkRequest(id)`). The `authorize-in`
-  callback branches on `state.link`: `authenticator.resolveIdentity(code)`
-  (token exchange + identity build, extracted from `authenticate()` —
-  which is now `resolveIdentity` + `accountManager.save`, login path
-  unchanged) then `accountManager.link(identity, userId)`: rejects an
-  identity linked to a DIFFERENT user
-  (`IdentityProviderAccountAlreadyLinkedError`), refreshes idempotently
-  for the same user, verifies user existence + provider/user realm match,
-  and NEVER runs mappers, mutates the user, mints a code or creates a
-  session. The redirect back is fixed and server-derived:
-  `<publicUrl>/account/connected-accounts?linked=<providerId>` or
-  `?linkError=already_linked|link_failed`. **Takeover posture — reason about
-  BOTH directions.** Linking an attacker's external identity to a victim
-  requires a state blob carrying the victim's userId, which only the
-  victim's bearer can mint. The reverse direction is the one that bites:
-  the attacker mints a state carrying their OWN userId and gets the victim
-  to follow it, so the victim's callback binds the VICTIM's external
-  identity to the ATTACKER's account, and the victim's next federated login
-  resolves through that row into it. State unguessability does not defend
-  this — the attacker is not guessing a state, they are supplying one — so
-  the ONLY control is the state's ip/user-agent binding, and both values
-  are chosen by whoever minted the state (`verify` skips a check whose
-  STORED value is falsy, and under the shipped `trustProxy: true` the ip is
-  the client-supplied left-most `X-Forwarded-For` entry). `completeLink`
-  therefore refuses a state carrying no binding at all. That is a stopgap
-  for the absent-binding hole only, not for a guessed one; issue #3439
-  moves the write behind a bearer-authenticated confirmation, after which
-  the guard and the state-carried userId both go away.
-- **Events** — `identityProviderLinked` (recorded in the callback branch)
+- **Link flow (two steps, the write is bearer-authenticated)** — server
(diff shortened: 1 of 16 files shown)


tada5hi:
## Audit outcome

**Verdict: sound-with-nits.** The security claim holds. The reviewer traced the attack end to end and confirmed the closure: the attacker's state carries their own `userId` (never client input, `linkRequest` reads it from the bearer), the victim's callback now writes nothing, and the confirm rejects on `link.userId !== identity.id` with the handle already burned by `consume`. The attacker never sees the handle either, since the redirect lands in the victim's browser and the console page ships `referrer-policy: no-referrer` with `frame-ancestors 'none'`.

Independently confirmed: the handle is 21 chars over a 36-char alphabet via nanoid's CSPRNG (~108 bits); `buildCacheKey` prevents a crafted handle from reaching another prefix's namespace; every re-validation (provider resolve, protocol, `enabled`, ForceLoggedIn, user type) runs **before** `consume`, so a disabled provider does not burn the handle; `accountManager.link()`'s DB-loaded realm check is stronger than repeating `linkRequest`'s; the double body read is safe (routup caches under `BodySymbol`); DI resolves lazily and cannot fail at boot; and the auto-POST cannot loop, since `router.replace` to the same route record does not remount.

### Fixed since (`5443410`)

- **`consume` was `get` + `drop`**, two round-trips on redis, where `ICache.pop` is the atomic read-and-drop the OAuth2 authorization-code repository already uses for the same semantic. Two simultaneous redemptions of one handle could both read the payload before either drop landed and both reach `link()`, where the `(providerId, userId)` unique index would reject one with a raw driver error instead of the intended 400. The comment now describes what the code does.
- **The link event carried no session attribution**, contradicting this branch's own architecture note that moving the write onto an authenticated request buys exactly that. It was hand-built with ip and user-agent only, while the sibling unlink event fills session, path and method from the request context. Now it reads the same context.
- The architecture doc's UI bullet still described the removed `linked` return param.

### Known residuals

- A session that lapses during the external round-trip loses the handle silently (the return lands unauthenticated, the realm grid renders, and the login kick's `redirect_uri` carries no query string). Requires the refresh token to expire mid-flow.
- authup's own HTTP access log records `?linkHandle=…` on the console GET. Harmless given the two redemption checks, but worth knowing.
- `isIdentityProviderAccountAlreadyLinkedError` lost its last call site with this change and is now exported but unused.

PR #3464, opened 2026-08-16
fix(server-core): harden the federated login callback

### LINKED ISSUE #3455: redirect_uri userinfo is dropped when matching but kept when redirecting
Found during the review of #3453. Pre-existing, and shared with the interactive path rather than specific to the federated one.

`isSimpleURLMatch` compares `${url.origin}${url.pathname}${url.search}${url.hash}` (`packages/kit/src/is-simple-url-match.ts`), which **drops userinfo**, while both redirect sites build the `Location` from the raw string:

- `AuthorizeController.confirm` (`adapters/http/controllers/workflows/authorize/module.ts`)
- the federated callback (`adapters/http/controllers/entities/identity-provider/module.ts`)

So `https://u:p@app.example.com/cb` matches the registered pattern `https://app.example.com/**` and the browser is then sent to `https://u:p@app.example.com/cb?code=...`.

This is **not** origin confusion: `https://app.example.com@evil.com/` canonicalises to `evil.com` and is correctly rejected. The issue is that an attacker-controlled credential blob rides the `Location` header, the RP's access logs, and the URL bar of anything that renders it.

Smallest fix: reject a `redirect_uri` carrying `url.username || url.password` in `OAuth2AuthorizationCodeRequestVerifier`, so the value that is matched is the value that is navigated. Worth also considering whether `ClientValidator` should refuse to register such a pattern in the first place.

### LINKED ISSUE #3456: Authorization state consumption is read-then-delete rather than an atomic pop
Found during the review of #3453. Pre-existing.

`OAuth2AuthorizationStateManager.verify` reads the state and then removes it as two steps (`core/oauth2/authorization/state/module.ts`: `findOneById` then `remove`), so two concurrent callbacks carrying the same state both read the payload before either deletes it.

Impact is bounded today because the **provider's** authorization code is single-use, so the second `authenticate` fails at the provider's token endpoint. That makes this a hardening item rather than a live hole, but the bound comes from the provider's behaviour rather than from ours.

The authorization-code repository already exposes the right primitive: `popOneById` (`core/oauth2/authorization/code/verifier/module.ts` uses it). Adding the same to `IOAuth2AuthorizeStateRepository` and using it in `verify` makes consumption atomic on our side.

Worth doing together with the account-link path, which reads the same state and where a duplicate completion is more than a nuisance (see #3439).

### LINKED ISSUE #3457: Federated callback mints an unbound code any confidential client can redeem
Found during the review of #3453. Pre-existing: the branch behaves identically before and after that PR.

The federated callback ends with:

```ts
// Only reachable when the state carried no code request at all
const url = new URL(this.options.baseURL);
url.searchParams.set('code', authorizationCode.id);
return sendRedirect(event, url.href);
```

Reachable when `authorize-out` is called with no `codeRequest`, which the exported kit component `ALoginForm` produces whenever it is rendered without one (`buildIdentityProviderURL` omits the parameter), and which `apps/server-core/test/unit/http/controllers/entities/identity-provider/login.spec.ts` exercises directly.

**Correction to the original description of this issue, from a review comment by @coderabbitai on #3453.** It called the code "unredeemable". That is only true for a public client. The code is minted with no `client_id`, no `redirect_uri` and no `code_challenge`, and `OAuth2AuthorizationCodeVerifier` skips each of those checks when the corresponding field is absent:

- `if (entity.client_id)` guards the client binding, so it is not checked
- `if (entity.redirect_uri)` guards the redirect binding, so it is not checked
- `if (options.clientIsPublic && !entity.code_challenge)` throws for a public client, but a **confidential** client passes

So **any confidential client that can authenticate at `/token` can redeem this code** and receive tokens for that user, without being the client the login was started for. Reaching it still requires obtaining the code, which is delivered to the IdP's own origin (the server root) and appears in the victim's address bar, so this is a defence-in-depth weakness rather than a direct path. It is worth closing regardless, because a code bound to nothing is exactly what the binding checks exist to prevent.

## Suggested fix

Refuse the completion when the state carries no code request, rather than minting a code from the `{ response_type: 'code' }` fallback. Nothing that works today depends on it: the code is sent to the API root, where no client is waiting for it.

The fix must also update `login.spec.ts`, whose `authorizeOut()` helper deliberately starts a login with no code request and asserts a 302. That premise changes with the fix, which is why this was not folded into #3453.

### LINKED ISSUE #3458: Federated callback refusals render as raw JSON instead of a page
Follow-up to #3453.

Before that PR, `GET /identity-providers/:id/authorize-in` always answered with a redirect. It can now refuse, and several of the refusals are new fail-closed checks the PR added deliberately:

- the client was deactivated, its grant allowlist changed, or its scopes were revoked while the user was at the provider (`401 invalid_client` / `400`)
- the client's `redirect_uri` patterns no longer match (`400 invalid_grant`)
- the provider was disabled (`400`)
- the user is inactive (`400`)

Refusing is right. Refusing with a raw JSON error body is not: this is a top-level browser navigation at the end of a login, so the person sees a JSON blob rather than a page telling them the login could not be completed.

The mechanism to do better already exists. The access-policy denial on the same path bounces to the hosted authorize page with `error=access_denied`, and `AuthorizeController.serve` maps that recognized query param onto a neutral hydration-payload error which the page renders as a denial card. The work is to decide which refusals map onto which neutral markers, and to extend that mapping in `serve()` (today it recognises `access_denied` only).

Care needed on two points:

- the marker must stay a small closed set of recognised values mapped to neutral server-side text, never attacker-controlled text echoed into the page
- a refusal must not become a redirect carrying a code or anything resembling success

### LINKED ISSUE #3459: Custom-scheme redirect_uri fails after the provider code was spent
Follow-up to #3453.

`isSimpleURLMatch` deliberately supports custom-scheme redirect URIs for native apps (`packages/kit/src/is-simple-url-match.ts` has an explicit branch for `myapp://cb`), so such a `redirect_uri` passes the code-request verifier and sets `redirectUriVerified`.

The federated callback then hands it to routup's `sendRedirect`, which allows http/https only, so the request fails **after** `authenticator.authenticate()` has already spent the provider's single-use authorization code and (on a first login) provisioned a user.

Both the old and the new behaviour are broken for this case: before #3453 the browser landed on the hosted authorize page with a code it could not redeem. The regression is only that it now fails louder and later.

Two ways out, depending on whether native clients are a supported federated case:

1. **Not supported:** reject a non-http(s) `redirect_uri` at `authorize-out`, before the user is sent to the provider. Fails early, spends nothing, and gives the RP an actionable error at the point of misconfiguration.
2. **Supported:** the callback needs an interstitial page that performs the custom-scheme navigation client-side, since a `Location` header cannot.

Option 1 is the smaller change and is worth doing regardless, because it also stops the same misconfiguration silently wasting a provider round trip.

tada5hi:
Closes #3455
Closes #3456
Closes #3457
Closes #3458
Closes #3459

Five follow-ups from the review of #3453, one commit each (in order below), all on the federated `authorize-out` / `authorize-in` leg.

## #3456 atomic state pop
`IOAuth2AuthorizeStateRepository` gains `popOneById` (cache `pop`) and `OAuth2AuthorizationStateManager.verify` pops instead of find-then-remove, so two concurrent callbacks carrying one state cannot both read it (login and link path alike; `findOneById` / `remove` had no other callers and are gone). The state repository also moves to its own `CacheOAuth2Prefix.AUTHORIZATION_STATE` (it was keyed under the CODE prefix, so either repository could pop the other's entries; the commit body describes what that allowed).

## #3455 redirect_uri userinfo
`isSimpleURLMatch` canonicalizes and drops userinfo while every redirect site navigates the raw value. A `redirect_uri` carrying `username`/`password` is now refused in the code-request verifier (before the pattern match), in `isValidPostLogoutRedirect` (the end-session `post_logout_redirect_uri`, the third site the review found), and `ClientValidator` refuses registering a `redirectUri` / `postLogoutRedirectUri` pattern carrying userinfo (the matcher ignores a pattern's userinfo, so such a pattern would silently accept the bare origin while reading as if it required credentials).

## #3457 no code request, no federated login
`authorize-out` requires a `codeRequest` (and now refuses a disabled provider) before any provider round trip; `authorize-in` refuses a login state without one (backstop for pre-deploy states). The fallback that minted an unbound code and redirected it to the server root is deleted, `verified` is non-optional. Kit `LoginForm` renders the identity-provider list only when a `codeRequest` prop is present. `login.spec.ts` starts its logins with a real code request.

## #3458 refusals bounce to the hosted page
Once the state is verified, refusals redirect to the hosted `/authorize` page instead of answering JSON: code-request re-verification failures (client deactivated / grant allowlist / scopes / `redirect_uri` mismatch) bounce with NO marker, since `serve()` re-runs the same verifier and renders the same refusal; a disabled provider bounces with `error=login_required` (new `OAuth2LoginRequiredError.providerUnavailable()`, neutral text); an inactive user with `error=access_denied`; the access-policy denial is unchanged. `serve()`'s marker set stays closed (an unknown `error` value is ignored, pinned) and no refusal redirect ever carries a code. Server failures (provider token endpoint down, invalid state) keep JSON: they are not refusals.

## #3459 custom-scheme redirect_uri: interstitial page
routup's `sendRedirect` is http(s)-only, so a native app's verified `myapp://cb` target used to fail after the provider code was spent. The callback now renders an interstitial page (new auth-console route `/identity-providers/:id/authorize-in`, matching the browser URL so hydration works) whose payload carries the verified target; the page attempts `window.location.assign` on mount and always renders an "Open application" button (browsers may require a gesture for external protocols), after `history.replaceState` to the hosted authorize URL so a reload lands on the login page rather than a consumed-state JSON error. Hardening that came out of review: a shared `isSafeRedirectURLScheme` (`@authup/kit`, denylist `javascript:` `data:` `vbscript:` `blob:` `file:` `about:`) is enforced in `ClientValidator`, the code-request verifier and, fail closed, the interstitial branch itself (an unhardened version would have navigated a `javascript:` target on the IdP origin); every served console page now sets `Cache-Control: no-store`; the auth console `CONTRACT_VERSION` is bumped to 2 (the route is part of the render contract a substituted package must provide); the router's dead code-exchange guard branch is removed (no producer sends `?code=` to a hosted page since #3453). i18n keys ×4 locales.

Docs: `.agents/architecture.md` (Federated Login Completion rewritten: state pop, no-code-request rule, refusal matrix + closed marker set, interstitial, scheme policy; Auth Workflow UI headers/routes; Console substitution v2), `.agents/structure.md`, `.agents/testing.md` (seven rendered pages).

## Verification
- `npm run test -w apps/server-core`: 196 files / 2199 tests passed (sqlite); kit 213, core-kit 24, kit 64, i18n 77, specs 15 passed.
- `check:types` clean; `npm run build -w apps/client-auth-console` (vue-tsc + client/ssr) success; eslint clean; `git diff --check` clean.
- Two adversarial review passes (security / correctness / UI / conventions lenses); every confirmed finding folded into the commit it belongs to.

## Summary by CodeRabbit

* **New Features**
  * Added a federated-login callback interstitial that helps users return to their application, including support for custom-scheme redirects.
  * Added localized messages and an “Open application” option in English, German, Spanish, and French.

* **Bug Fixes**
  * Improved handling of refused or unavailable federated logins.
  * Blocked unsafe redirect URLs and userinfo-containing URLs.
  * Prevented sensitive authentication pages from being cached.
  * Strengthened authorization-state handling to prevent reuse.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3466, opened 2026-08-17
fix(server-core): batch the event retention sweep

tada5hi:
## Problem

`EventRepositoryAdapter.deleteExpired` issued a single unbounded statement:

```ts
await this.repository.delete({ expiring: true, expiresAt: LessThan(now) });
```

`components/event-cleaner` re-runs it every minute, and every replica runs its own cron. In steady state that removes a trickle. The pathological cases are the first sweep after someone lowers `eventLogRetentionDays`, and the day a full retention window first matures: both can match millions of rows in one long transaction, issued concurrently by N pods.

## Change

Remove rows in bounded batches (`EVENT_RETENTION_SWEEP_BATCH_SIZE`, 1000): select ids, delete by id, loop until drained.

Two details worth review:

- **Why an id select and not `DELETE ... LIMIT`.** typeorm's `DeleteQueryBuilder.limit()` is MySQL-only; postgres has no `DELETE ... LIMIT`. The id select is the portable form across all three supported drivers.
- **Why the loop can stop early.** A batch that removes nothing means another replica's sweep already owns those rows. Stopping avoids re-selecting them; the sweep is idempotent and the next tick takes the remainder. Termination otherwise comes from the short final batch.

Cross-replica coordination is deliberately unchanged: this bounds each statement, it does not serialize sweeps between processes.

## Not included

`SessionTokenRepositoryAdapter.deleteExpired` (the oauth2 cleaner's half) carries the identical unbatched shape. Left out to keep this reviewable; happy to fold it in or file it separately.

## Verification

- New specs in `test/unit/app/modules/database/event-repository.spec.ts`, written test-first. The batching spec failed as `expected "delete" to be called 3 times, but got 1 times` against the old code, which is the defect stated as an assertion.
  - drains every expired row across several batches
  - bounds each statement to the batch size
  - keeps rows that have not expired and rows that never expire
- Full server-core suite: 197 files, 2 247 tests, all passing.
- `npm run build -w apps/server-core` and eslint clean.

The sweep's end-to-end behaviour was also confirmed by hand against a populated sqlite database before and after: 2 205 rows with 1 896 aged out swept to 0 residual expired rows.

## Summary by CodeRabbit

* **New Features**
  * Expired security events are now removed in controlled batches, improving cleanup reliability and reducing the impact of large retention sweeps.
  * Cleanup can continue safely across concurrent service instances without repeatedly processing the same records.

* **Bug Fixes**
  * Prevented retention cleanup from stopping prematurely when database operations report incomplete affected-row counts.
  * Ensured active and non-expiring events remain available.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -4287,7 +4287,16 @@ hub lacks: a **closed taxonomy** (`EventName`/`EventScope` enums in
   mirror) deletes `expiring = true AND expiresAt < now` (hub's cleaner shape);
   scheduled only when
   `eventLogEnabled && eventLogRetentionDays > 0`. Per-action retention later is
-  per-action stamping — no schema change.
+  per-action stamping — no schema change. The delete is **batched**
+  (`EVENT_RETENTION_SWEEP_BATCH_SIZE`, 1000): steady state removes a trickle,
+  but the first sweep after lowering `eventLogRetentionDays` (or the day a
+  full retention window first matures) can match millions of rows, and a
+  single unbounded `DELETE` would be one long transaction issued concurrently
+  by every replica. Batching selects ids then deletes by id, because
+  `DELETE ... LIMIT` is MySQL-only; the loop drains, and a batch that removes
+  nothing means another replica's sweep owns those rows, so it stops and the
+  next tick takes the remainder. `SessionTokenRepositoryAdapter.deleteExpired`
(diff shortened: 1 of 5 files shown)


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3474, opened 2026-08-19
fix(server-core): writable directory defaults

tada5hi:
Two related changes to `writableDirectoryPath`, plus the doc corrections they turned up.

## `writableDirectoryPath` now resolves against `rootPath`

`normalizeConfig` resolves every path key against `rootPath` after the `...parsed` spread, so consumers receive an absolute path and none of them has to know what the process cwd was. `themeDirectoryPath`, `authConsolePath` and `accountConsolePath` were in that block; `writableDirectoryPath` was not — it was computed *before* the spread off `process.cwd()` and never resolved afterwards. So a configured relative value stayed relative, and setting `rootPath` moved the three console/theme paths while leaving the writable directory behind.

Only a deployment that sets `rootPath` away from the working directory **and** gives `writableDirectoryPath` a relative value sees a change. Absolute values and the default are unaffected.

Four tests pin the behaviour (default, configured root, relative-resolves, absolute-wins).

## Docker: writable directory moved to `/var/lib/authup`

The image wrote to `/usr/src/app/writable`, inside the application install directory. `/var/lib/<app>` is the filesystem hierarchy standard location for mutable application state, and a cleaner mount point than a path nested in the install tree.

**Only the image default moves.** The code default stays `<rootPath>/writable`, which is what keeps an unprivileged `npx @authup/authup start` and a bare-metal run working — every absolute system path (`/etc/authup`, `/opt/authup`, `/var/lib/authup` alike) needs root or a pre-chowned directory, so making one the built-in default would turn a working quickstart into EACCES on the first production log write.

### Breaking, and silently so

A deployment mounting a volume at the old path keeps starting fine: the container writes its production log files into the container layer instead of the volume, and finds no provisioning files, so file-based provisioning stops being applied with nothing in the log to say why. `upgrading.md` carries the mount diff and the `WRITABLE_DIRECTORY_PATH=/usr/src/app/writable` escape hatch.

The mount target is updated at all 10 places it appears (README, `docker.md`, `docker-compose.md` ×4, `provisioning.md`, and the two VitePress components).

**The helm chart lives in `authup/helm` and needs the same mount path change there.**

## Doc corrections

Both `configuration-server-core.md` and the `Config` type claimed the writable directory holds the SQLite database. It never has — the file comes from `db.database` / `DB_DATABASE`, which typeorm-extension resolves against the process cwd (`resolveSQLiteDatabasePath`), so pointing `WRITABLE_DIRECTORY_PATH` at a volume does **not** move the database onto it. What actually lands there is the production log files, plus `<dir>/provisioning` on the read side. `.agents/conventions.md` records both facts so the next path key joins the resolution block and nobody re-derives the database misconception.

## Not included

Defaulting the SQLite `database` to `<writable>/db.sql`. With `db` unset and no `DB_TYPE`, `buildDataSourceOptions` throws outright today, so that default would be purely additive and would finally put the file that actually needs a volume onto the volume — but it is a separate call from this one.

## Verification

- `npm run build -w apps/server-core` — clean
- `npx eslint --fix` on the changed TypeScript — clean
- full server-core suite — **197 files / 2256 tests passed**

## Summary by CodeRabbit

- **New Features**
  - Docker deployments now use `/var/lib/authup` for persistent writable files, including logs and provisioning data.
  - Relative writable-directory paths resolve consistently from the configured application root.

- **Bug Fixes**
  - Absolute writable-directory paths remain unchanged, while default and relative paths are normalized reliably.

- **Documentation**
  - Updated Docker, provisioning, configuration, and upgrade guidance, including volume-mount migration instructions.
  - Clarified that SQLite database storage is configured separately.

### DIFF
--- a/.agents/conventions.md
+++ b/.agents/conventions.md
@@ -205,6 +205,22 @@ lists `vue` needs this before merge; its own CI already shows the failure.
 - Config keys in `app/modules/config/types.ts` match the service option names
 - Environment variable names use `SCREAMING_SNAKE_CASE` with `_ENABLED` suffix: `REGISTRATION_ENABLED`, `PASSWORD_RECOVERY_ENABLED`, `EMAIL_VERIFICATION_ENABLED`
 - Config file keys (`.conf`) use `camelCase` matching the TypeScript property name
+- **Every path key is resolved against `rootPath` after the `...parsed` spread**
+  in `normalizeConfig`, so consumers receive an absolute path and none of them
+  has to know what the process cwd was: `writableDirectoryPath`,
+  `themeDirectoryPath`, `authConsolePath`, `accountConsolePath`. A new path key
+  joins that block; computing it *before* the spread reads a default `rootPath`
+  rather than the configured one (the `writableDirectoryPath` bug, fixed after
+  v1.0.0-beta.62 — it silently ignored `rootPath` and stayed relative).
+- **`writableDirectoryPath` does not hold the database.** It holds the
+  production log files and is where `<dir>/provisioning` is read from; that is
+  the whole of it. The sqlite file comes from `db.database` / `DB_DATABASE`,
+  which typeorm-extension resolves against the process cwd
+  (`resolveSQLiteDatabasePath`), so pointing `WRITABLE_DIRECTORY_PATH` at a
+  volume does NOT move the database onto it. The Docker image defaults the key
+  to `/var/lib/authup`; everything else defaults to `<rootPath>/writable`,
+  which is what keeps an unprivileged `npx` start working (any absolute system
+  path needs root or a pre-chowned directory).
 
 ## Upstream (Own) Libraries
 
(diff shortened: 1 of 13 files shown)


tada5hi:
Downstream consumers of the image, checked and filed:

- authup/helm#9 — provisioning ConfigMap mounts at `<old>/provisioning`, no `WRITABLE_DIRECTORY_PATH` set. Pinned `appVersion: 1.0.0-beta.62`, so it breaks on the appVersion bump.
- PrivateAIM/helm#175 — same provisioning-mount break, but `image.tag: latest` + `pullPolicy: Always`, so it lands as soon as the image is published rather than on a chart change.
- dnpm-dip/deployment#14 — named volume at the old path; MySQL-backed with no provisioning mount, so the only effect is that logs stop persisting across container recreation.

The common fix suggested in all three is to set `WRITABLE_DIRECTORY_PATH` explicitly to whatever path is mounted, which makes them immune to the image default in either direction. PrivateAIM's chart already does exactly that for the theme directory (`THEME_DIRECTORY_PATH` from `theme.mountPath`), which is why its theme mount is unaffected while provisioning is not.

PR #3475, opened 2026-08-19
feat(server-core): complete a federated login through the hosted authorize ladder

### LINKED ISSUE #3454: Federated login bypasses the MFA, consent and prompt gates the interactive authorize applies
Split out of #3453, which fixed the delivery of the federated authorization code but deliberately did not change this.

The external-IdP callback (`IdentityProviderController.authorizeIn`) issues its code by calling `codeIssuer.issue` directly, where the interactive path routes through `OAuth2Authorization.authorizeInner`. So a federated login skips:

- the **MFA backstop** (`authorization/module.ts`, `if (challenge.required && !session?.mfaAt) throw`). A user holding a confirmed TOTP/WebAuthn factor completes an RP authorization on the external factor alone, and `enrollmentRequired` never routes them to inline enrollment.
- **`prompt=login` / `max_age`** freshness. Arguably satisfied in spirit (the user just authenticated at the provider, and the code's `auth_time` is now), but it is not evaluated.
- **consent** for non-`builtIn` clients, so no `auth_consents` rows are written. A later `prompt=none` request therefore answers `consent_required` indefinitely for that user/client pair.
- **`acr_values` step-up**. The value rides the code (`code/issuer/module.ts`) but nothing reads it at redemption, so an RP asking for `urn:authup:mfa` gets a token that truthfully reports no MFA instead of the `stepUpRequired` the interactive path raises.

The MFA and consent exclusions are documented (`.agents/architecture.md` -> *Intentional enforcement boundaries*, `.agents/references/authentik.md`): the upstream provider is the trusted authentication. This issue is to decide whether that still holds now that the leg actually completes.

**The obvious fix is a trap.** Delegating to `OAuth2Authorization.authorize()` would restore all of the above in one move, but its MFA backstop reads `session?.mfaAt` and this leg has no session, so every user with a confirmed factor would be permanently unable to log in through federation. Any real fix needs to answer "where does the second factor get challenged for a federated login" first - most likely a step-up on the hosted page before the code is issued.

Related exposure: `GET /identity-providers/:id/authorize-out` is unauthenticated and takes an attacker-supplied `codeRequest`, so a one-click victim flow issues a code to any uri inside the target client's registered patterns with an attacker-chosen `code_challenge`, with no interactive step in front of it. That is the usual redirect-pattern hygiene requirement rather than a new hole, but it is sharper here than on the interactive path, which renders consent/MFA first.

### LINKED ISSUE #3478: deriveAmrAcr reports acr urn:authup:pwd for an externally authenticated session
`deriveAmrAcr` (`apps/server-core/src/core/oauth2/authorization/helpers.ts`) maps a session's `authMethod` onto `amr`, then derives `acr` from `mfaAt` alone:

```ts
acr: session.mfaAt ? OAuth2AuthenticationContextClass.MFA : OAuth2AuthenticationContextClass.PASSWORD
```

So a session established by an external identity provider reports `amr: ['ext']` together with `acr: urn:authup:pwd`. The `amr` is right and the `acr` says the subject authenticated with a password, which is not what happened.

It was a cosmetic mismatch while every session eventually carried a local factor. It matters more now that a federated login deliberately carries none (#3454), because `ext` sessions are precisely the ones that will report it.

## Options

- A third value, e.g. `urn:authup:ext`, for a session authenticated elsewhere with no local factor.
- Or omit `acr` entirely for that case and let the absence speak, which is also OIDC-legal since `acr` is voluntary.

Either way the value is on every token kind, so this is a wire-visible change and wants a note in the release.

## Related

An application sending `acr_values=urn:authup:mfa` against a federated user who holds NO local factor gets a token with the achieved `acr` and no error, since the step-up branch only fires when the user actually holds a factor. That is deliberate per OIDC Core section 5.5.1.1, but it is worth stating in the same place, so an application author knows the request can be silently unsatisfiable.

tada5hi:
Closes #3454, closes #3478. Implements stage 1 of plan 094.

## What changes

The external-IdP callback (`GET /identity-providers/:id/authorize-in`) no longer mints the relying party's authorization code. It establishes the authup session (`auth_method: ext`, `mfa_at: null`, lifetime of the pending login) and returns the browser to the hosted `/authorize` page. The page completes the login, and from there the flow is the ladder every interactive login takes, so a federated login now passes the gates a password login passes:

- `prompt=login` / `max_age` freshness and `select_account`
- the `acr_values=urn:authup:mfa` step-up
- consent, with its persisted `auth_consents` rows, so a later `prompt=none` for that user/client pair is covered
- the client's grant allowlist, scopes and `redirect_uri` patterns, re-verified at completion time rather than at the start

One `auth_sessions` row per federated login: the `/token` exchange reuses the one the callback created, and the id_token's `sid` names it. Claims read `amr: ['ext']`, plus `otp` and `acr: urn:authup:mfa` once a factor was verified.

## The browser binding is a cookie, at both hops

Server auth is header-only and the hosted page's session lives in the browser, so the callback has to hand the browser something it can turn into a bearer. That is `authup_federated_login`: HttpOnly, `SameSite=Lax`, `Secure` under an https `publicUrl`, path-scoped to `<basePath>/identity-providers` so it never rides an ordinary API request.

- **`authorize-out`** mints a nonce into that cookie and stores it on the state blob. The callback refuses unless the two match, before the provider's single-use code is spent. Without it the callback would establish a session for whatever browser presented a crafted callback URL: an attacker runs a federated login for their own external account and hands the URL to someone else, whose browser adopts that session and then consents the application into it. The state's own ip and user agent cannot carry that weight, and #3439 already recorded why: both are chosen by whoever mints the state, and under the shipped `trustProxy: true` the address is the client-supplied left-most `X-Forwarded-For` entry.
- **`authorize-in`** swaps the nonce for the pending login id. Only the browser the callback answered carries it, and no other origin can set it, so the URL is empty-handed apart from a `provider` hint.
- **`POST /identity-providers/:id/login-complete`** reads it, clears it, and answers with the grant. It additionally requires the request's `Origin` to be `publicUrl`'s own, and that is load-bearing rather than belt-and-braces: this is authup's first cookie-authenticated endpoint, and the CORS default reflects every origin **with** credentials on the premise that none exists. `SameSite` does not save it, since the attribute is scoped to the registrable domain rather than the origin, so a sibling subdomain is same-site and its credentialed `fetch` would read the token pair.

This is the shape both cohort products use. Keycloak keeps the pending login in an auth session identified by `AUTH_SESSION_ID` and its `session_code` is a secret inside that cookie-reachable session; Authentik keeps the `FlowPlan` in the Django session behind `authentik_session` and its OAuth source callback redirects to the flow SPA with an empty-handed URL. Neither hands a front end a one-time value to trade for a session.

## The local second factor stays with the local credential

An external session is not challenged for an authup factor, and `mfaRequired` does not force local enrollment on those users. The provider is the authentication authority for that login, which is where MFA is configured and enforced for it. The route is opt-in either way: an external identity reaches an existing account only through the bearer-authenticated link flow, since a first login provisions a new user (plan 091).

The exception is an explicit `acr_values=urn:authup:mfa`. The application asked, and a local factor is the only proof authup can produce, so the step-up branch applies to an external session too.

Keycloak and Authentik default the same way, with a per-IdP post-login flow / a per-source Authenticator Validation stage as the opt-in. The boundary is written down in `.agents/architecture.md` under *Intentional enforcement boundaries* and in the operator docs, together with the residual: authup does not verify the upstream's own `amr`/`acr` (filed as #3477).

Note this is a decision not to tighten one gate, not a loosening of anything: before this PR the callback minted the code directly, so a federated login passed none of the gates above.

## Breaking

- The auth console render contract is at **version 3**. A substituted console package must complete `data.federatedLogin`, so `bindConsolePackages` fails closed on a package built against 2.
- The callback's custom-scheme interstitial is no longer rendered. Such a `redirect_uri` is navigated at the end of the ladder, the same place an interactive login navigates it. The route stays exported by the console package.
- A federated login now needs the authup origin's cookies. A browser that blocks them for that origin cannot complete one.

## Deliberate deviations from plan 094

- the callback keeps its access-policy leg rather than deferring it to `authorizeInner`
- the LOGIN event uses `data.reason: 'federated'`; the event data key allow-list is closed and `reason` is the key this leg already used
- the pending login is named in the payload by its provider only, since the id itself rides the cookie

## Also in here

- `resolveURL` for the provider callback URL. The template-literal join produced `//identity-providers/...` under a `publicUrl` with a trailing slash, which 404'd every federated login. Found by running the flow in a browser against two real instances, not by the suite.
- the cookie's `Secure` flag is read from the parsed URL protocol. `publicUrl` is validated but never canonicalized, so a configured `HTTPS://...` left a prefix test false and dropped the flag.
- `completeHandoff` enforces the pending session's own expiry. The lookup is by id alone, so it was enforced only by the cache entry agreeing, and under skew between the two stores `refresh` would have extended an abandoned login into a regular session.

## Verification

- `apps/server-core`: 201 files / 2285 tests
- `packages/client-web-kit`: 30 files / 229 tests
- `packages/core-http-kit`: 6 files / 24 tests
- `apps/client-auth-console`: 1 file / 4 tests, a new harness. The app had no test setup at all.
- builds and typechecks clean across the changed workspaces, eslint clean
- the whole flow driven through a path-aware cookie jar (`TestCookieJar`), so a wrongly scoped cookie fails the suite rather than passing a spec that echoes `set-cookie` back by hand
- red/green checked per assertion that pins a decision: the MFA bypass, the cookie path, the pending-session expiry, the `federatedLogin` payload binding, the `provider` strip
- exercised end to end in a browser against two authup instances, one brokering to the other


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai full review

PR #3476, opened 2026-08-19
fix: carry the post-login destination through the authorize flow

tada5hi:
A visitor sent from a restricted page to the login flow was meant to land back on that page. They landed on `/`.

The destination was carried in `sessionStorage` as `AuthorizationRequest.target`, so the round-trip was a convention every app had to remember rather than something the protocol carried, and each of the three consoles held its own 25-line copy of the glue (this repo, `PrivateAIM/hub`, `dnpm-dip/portal`). It now rides in the OAuth2 `redirect_uri`'s own query:

```
/users  ->  /login?redirect=%2Fusers
        ->  <idp>/authorize?...&redirect_uri=<origin>/login/callback?redirect=%2Fusers
        ->  <origin>/login/callback?redirect=%2Fusers&code=...&state=...
        ->  /users
```

Nothing about that is authup-specific, which is the point: any RP in any framework gets the behaviour by putting the destination in its own callback URI.

## Why the redirect_uri stays the fixed callback path

The destination is a parameter *on* the callback, never a replacement for it. The callback is where the exchange happens, it is what the client's allowlist registers, and keeping the authorization code off a content page is what RFC 6749 §10.5 asks for (`Referer` leakage).

Three things had to hold for a query-carrying `redirect_uri` to survive, and all three were verified rather than assumed:

| Leg | Result |
|---|---|
| Registered-pattern match | `is-simple-url-match.ts:25` keeps `url.search` in the canonical form, and `**` runs to the end of the value, so the provisioned `<origin>/**` still matches |
| `code`/`state` append | All 9 redirect-construction sites use `new URL()` + `searchParams.set()`. No string concatenation with `?` exists anywhere |
| `/token` replay | `code/verifier/module.ts:47-51` compares the raw stored string with `!==`, and the console replays the stored value byte-for-byte |

`grant-authorize.spec.ts` now exercises all three against the real server in one case, rather than leaving them to reasoning.

## Fixes to the routing interceptor

Three ways the destination was being dropped, none of which the storage change alone would have fixed:

1. **A permission denial after login bounced back to the callback route.** `from` is `/login/callback`, whose only job is to forward to `/`, so a denial read as "you were dropped somewhere random". Now it names the home route directly, which is also the first use of the `homeRoute` option: it was plumbed through the module and the constructor and then never read.
2. **`{ path: from.fullPath }` silently dropped the query and hash.** vue-router runs a `path` through `parseURL`, which keeps only the path, so bouncing off `/roles?page=2#row-7` landed on a bare `/roles`. Verified by executing `router.resolve()` against the installed vue-router 5.2.0, not only by reading it.
3. **`redirect` was not stripped after a failed exchange**, leaving a dead param on the callback URL.

`packages/client-web-nuxt` had no test setup at all and `RoutingInterceptor` was entirely uncovered. It has one now: 12 cases.

## Security

The destination now arrives as URL input that a crafted authorize request could shape, where before it came from same-origin storage. So the interceptor accepts a **site-relative path and nothing else** — a value not starting with a single `/` is dropped rather than reduced to its path, since an attacker-chosen path is no better than an attacker-chosen host. `https://evil.test/x`, `//evil.test/x` and `javascript:alert(1)` are all pinned as ignored.

The PKCE `code_verifier` and `state` still live in `sessionStorage`, because they must: a public client cannot redeem without the verifier, and it must never leave the browser. That is also why losing storage was never the cause of the reported bug — the login would have failed outright rather than landing the user on `/`.

## Deployment consideration

A client whose registered redirect pattern is the **exact** callback URL with no wildcard will reject a login that carries a destination, while a login started directly from `/login` keeps working — so it presents as an intermittent login failure rather than a config error. The provisioned system clients (`admin-console`, `account-console`) register `<origin>/**` and are unaffected; a hand-registered pattern (reachable via `CLIENT_ID` / `NUXT_PUBLIC_CLIENT_ID`) needs widening.

## Deliberate behaviour change

The old login page folded **every** sibling param on the login URL into the destination (`/login?redirect=/users&invite=x` → `/users?invite=x`). Only `redirect` is forwarded now. Nothing in the app has ever produced siblings — the routing interceptor and the session-expiry plugin each emit a lone `redirect` — so this drops a feature with no known caller rather than reinstating it. Noted in the code so it stays a decision rather than a deletion nobody noticed.

## Also removed

`extractOAuth2QueryParameters` (`client-web-kit`) lost its last call site in #2669 and was never re-exported from a barrel, so only an orphan `.d.ts` remained. It was also wrong: it ran `decodeURIComponent` over vue-router `LocationQuery` values that `parseQuery` had already decoded, so a correctly encoded `%` threw `URIError` and a value that still looked like an escape sequence was silently mangled (`myapp://cb?x=%321` → `?x=21`). Harmless while every caller passed a `redirect_uri` with no percent-encoding, and a trap for the next one. Removing it drops the only vue-router type reference in the built declarations.

## Follow-ups, deliberately not in this PR

- `routing-interceptor.ts:126,141`: when the attempted route already carries a `redirect` query param, `hasQueryRedirectProperty` omits the param entirely rather than forwarding it, so `/users?redirect=x` bounces to a bare `/login`. Same defect class, but the guard may be load-bearing and I did not want to widen scope.
- `routing-interceptor.ts:179-186`: a permission denial on a hard page load answers with a full `store.logout()`. `from.path === to.path` is true on *every* initial load in Nuxt (the router replays the resolved initial route with `force: true`), so this is the dominant path, not an edge case. Left alone: dropping the `redirect` there is what currently stops a re-login walking straight back into the same denial.
- `hub` and `portal` need the same one-line login-page change once this ships. Both currently write `target`, which no longer exists.

## Verification

- `apps/server-core`: 197/197 files, 2257 tests. (An earlier run hit an `afterAll` teardown timeout in `account-pages.spec.ts`; it reproduced with these changes stashed and did not recur.)
- `packages/client-web-kit`: 217 pass.
- `packages/client-web-nuxt`: 12 pass (new).
- Full `npm run build` and `eslint` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved post-login navigation while preserving destination paths, query parameters, and URL fragments.
  - Added safeguards to prevent redirects to external or malformed destinations.
  - Fixed federated and authorization-code login flows to reliably preserve redirect information.
  - Improved permission-denied navigation, including route details and fallback behavior.
  - Removed obsolete login destination handling that could cause inconsistent redirects.

- **Documentation**
  - Documented the updated post-login redirect behavior and validation rules.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
Investigated all three review comments.

**`grant-authorize.spec.ts` — byte-different `redirect_uri`: valid, fixed.** The round-trip test only showed the original value is accepted, which a verifier that normalized the query would satisfy equally. Added a negative case: `redirect=/users` for `redirect=%2Fusers` (identical once parsed, different bytes) is refused with `invalid_grant`. The strictness holds, and is now proven rather than asserted in prose.

Not repeated on the federated spec as suggested. Both legs redeem through the same `OAuth2AuthorizationCodeVerifier` comparison (`code/verifier/module.ts:47-51`), so a second case would cover the same `!==` twice; the federated leg's own risk is the code-request re-serialisation, which its positive test already covers.

**`architecture.md` — MD040: applied.** The file has ~10 other unlabeled fences and the repo has no markdownlint config or CI step, so this is cosmetic, but `text` is right for a diagram and the idiom already exists at line 1386.

**`package.json` — `npm run build` in the test script: declining.** No package in this repo does that (`client-web-kit`, `kit`, `access`, `core-kit`, `errors`, `specs`, `server-kit` all run vitest directly), and `nx.json` declares `dependsOn` only for `build`, never `test` — CI satisfies the guideline by running Build as a separate step before `nx run-many -t test`.

The `nuxt-module-build prepare` prefix is deliberately narrower than a full build. It exists only because this package's tsconfig extends the gitignored `.nuxt/tsconfig.json`; `npm run build` here is `prepare && build`, so the suggestion would add a module bundle to every test run to obtain a file the prepare step alone already produces.

PR #3479, opened 2026-08-19
feat(server-core): verify the upstream provider's amr/acr per identity provider

### LINKED ISSUE #3477: Verify the upstream provider's amr/acr per identity provider
Follow-up to #3454 / #3475, which shipped the posture this issue is about.

A federated login is authenticated by the external provider, and authup stacks no local factor on top. That is what Keycloak and Authentik default to, and it matches authup's linking model, where an external identity reaches an existing account only through the bearer-authenticated link flow.

But "MFA is guaranteed by the external provider" is a statement the operator makes, not one authup checks. An upstream that permits password-only silently downgrades every authup account reachable through it, and nothing in the callback notices. Since #3475 made that the shipped default, this is the one unchecked premise of the design.

## Where things stand after #3475

Half of part 1 already exists. `IdentityProviderOAuth2Authenticator.buildIdentityWithTokenGrantResponse` (`apps/server-core/src/core/identity/provider/authentication/protocols/oauth2/module.ts:204`) already decodes the upstream `id_token`:

```ts
if (typeof input.id_token === 'string') {
    // ... extractTokenPayload(input.id_token) merged into `claims`
}
```

The claims are merged into a local `claims` object, used for the name ladder, and then dropped: `identity.data` is deliberately the access-token payload alone, because mappers pattern-match over it and widening it would change every existing mapper's input. So `amr` / `acr` are in hand at exactly the right moment and simply not kept.

Note the guard on that block is already the OIDC-only condition this issue needs: a plain `oauth2` provider returns no `id_token`, so there is nothing to verify.

## Part 1 - verify what came back

1. Carry the two claims out of the builder. Either a dedicated field on `IdentityProviderIdentity` (`apps/server-core/src/core/identity/provider/types.ts:11`), e.g. `assurance?: { amr?: string[], acr?: string }`, or a second return value. A dedicated field keeps `identity.data` untouched, which is the constraint above.
2. Add the operator allow-list to the OAuth2 provider attributes: `OAuth2IdentityProviderBase` (`packages/core-kit/src/domains/identity-provider/oauth2/types.ts`), its validator (`.../oauth2/validator.ts`) and the preset validator's key list, then the admin form fields (`packages/client-web-kit/src/components/entities/identity-provider/AIdentityProviderOAuth2*Fields`). Two nullable fields, e.g. `requiredAmr` and `requiredAcr`, comma-separated like the existing `scope`. Null on both keeps today's behaviour, so this is opt-in and needs no migration beyond the columns.
3. Gate in `OAuth2FederatedLoginService.complete` (`apps/server-core/src/core/oauth2/federated-login/module.ts`), beside the two refusals that are already there: the disabled provider at line 173 and the inactive user at line 190. The new one belongs after the exchange, since the claims only exist then, so it is in the same position as the inactive-user check: the provider's code is already spent, but no authup session exists and no code is issued.
4. Refuse the same way those do, by bouncing to the hosted page with a marker `serve()` maps onto neutral server-side copy. `access_denied` already exists and reads correctly here. Do not echo the upstream's claim values into the redirect: the marker set in `AuthorizeController.serve` is deliberately closed.

A provider that returns no `id_token`, or an `id_token` carrying neither claim, must refuse when the allow-list is set. Failing open there would make the feature decorative.

## Part 2 - ask on the way out

Send `acr_values` (and optionally `max_age`) on the outbound authorize request, so the upstream is asked to step up rather than merely observed. The injection point is `buildRedirectURL` (`.../protocols/oauth2/module.ts:74`), which passes `Partial<AuthorizeParameters>` straight to the hapic client; today `authorize-out` passes `{ state }` alone (`apps/server-core/src/adapters/http/controllers/entities/identity-provider/module.ts`, the `sendRedirect(event, authenticator.buildRedirectURL({ state }))` line).

Independent of part 1 and worth less on its own: `acr_values` is voluntary in OIDC Core §5.5.1.1, so a provider may ignore it and return a token anyway. Only part 1 turns it into a guarantee. Ship part 1 first, or both.

## Decisions to make

- **One field or two.** `amr` is a list and `acr` a single value with provider-specific vocabulary (`acr: "1"`, `amr: ["mfa"]`, `["pwd","mfa"]`, ...). Simplest that works: require any intersection with `requiredAmr`, and exact membership for `requiredAcr`. Authup cannot normalize these, so it is an operator-filled allow-list either way.
- **Whether the check also applies to the account-link flow.** `link-confirm` admits an external identity to an existing account, so the same argument applies, and `resolveIdentity` is shared. Probably yes, but it is a second refusal path with its own UX.
- **Whether a refusal should be distinguishable from the other refusals.** The current set is deliberately opaque to the browser. An operator debugging a misconfigured allow-list needs the reason somewhere, which argues for a `describeError`-style log line rather than a new marker.

## Tests

The federated e2e specs already stand up a fake provider that returns an unsigned three-segment `id_token`, so a claim can be planted without any crypto: see `test/unit/http/controllers/entities/identity-provider/login-completion.spec.ts` and the fake token endpoint in `login-cookie-flow.spec.ts`. The matrix worth pinning: allow-list unset behaves as today, a matching claim completes, a mismatching claim refuses with no session and no code, and a provider returning no `id_token` refuses when the allow-list is set.

## Docs

`.agents/architecture.md` under *Intentional enforcement boundaries* and `docs/src/guide/user/identity-providers.md` both state today that authup trusts the upstream without checking. Both need the new opt-in described, and the boundary restated as "unchecked unless you configure this".

## Notes

- OIDC only. A plain `oauth2` provider has no `id_token` to read, so the fields must be optional and the check must state clearly that it cannot apply.
- Neither Keycloak nor Authentik verifies upstream `amr`/`acr` by default, so this is a differentiator rather than parity work.
- Pairs with the existing exception: an application sending `acr_values=urn:authup:mfa` already triggers a local factor when the user holds one, which is authup's own proof rather than the upstream's.

tada5hi:
Closes #3477.

A federated login is authenticated by the external provider and authup stacks no local factor on top. That is the Keycloak/Authentik default and it matches authup's linking model, but "MFA is guaranteed by the provider" was a statement the operator made and nothing checked. An upstream that permits password-only silently downgraded every authup account reachable through it. Since #3475 made that the shipped default, it was the one unchecked premise of the design.

## What this adds

Two optional attributes on an OAuth2/OIDC identity provider:

| Field | Passes when |
|---|---|
| `requiredAmr` | the upstream `id_token`'s `amr` intersects the list |
| `requiredAcr` | the upstream `id_token`'s `acr` is a member of the list |

Both take a comma- or space-separated list and are **null by default**, which is exactly today's behaviour. They are extra-attribute rows, so there is **no migration**. Authup cannot normalize either vocabulary (`acr: "1"`, `amr: ["mfa"]`, `["pwd","mfa"]`, ...), so the values are operator-filled either way.

`requiredAcr` is additionally sent **outbound** as `acr_values` on the authorize request (part 2 of the issue), so the upstream is asked to step up rather than only observed. That is voluntary per OIDC Core 5.5.1.1, so only the inbound check turns it into a guarantee; it is derived from the same field rather than a third one.

## Where the gate sits, and why not where the issue proposed

The issue sketched carrying an `assurance` field out on `IdentityProviderIdentity` and gating in `OAuth2FederatedLoginService.complete`, beside the inactive-user refusal. This instead gates inside `IdentityProviderOAuth2Authenticator.resolveIdentity`, for three reasons:

- **It refuses before anything is written.** `complete()` gates *after* `authenticator.authenticate()`, which has already run `accountManager.save` - so the issue's position would provision a user for a login it then refuses. `resolveIdentity` runs before that.
- **It covers the account-link flow for free.** `resolveIdentity` is shared with `link-confirm`, which admits an external identity to an *existing* account and needs the same check (the issue listed this as an open decision). No second refusal path to design.
- **A preset cannot skip it.** It reads the token response rather than the built identity, because the five presets (github, facebook, instagram, paypal, google) override `buildIdentityWithTokenGrantResponse`. None overrides `resolveIdentity` or `authenticate`.

It **fails closed** on every uncertainty: the claims come from the `id_token` alone, so a provider that returns none - a plain `oauth2` provider, or an OIDC one whose scope lost `openid` - or one whose token cannot be decoded or carries neither claim, is refused once an allow-list is set. Failing open there would make the feature decorative.

`OAuth2FederatedLoginService.complete` catches `IdentityProviderAssuranceError` and returns a refusal that bounces to the hosted page with `access_denied`, like the inactive-user and access-policy refusals. The reason reaches the operator's log through `describeError` and never the browser, since the marker set `AuthorizeController.serve` recognizes is deliberately closed - which is also why the error carries no `ErrorCode`: both callers swallow it into a redirect, so the `markInstanceof` symbol is the only discriminator anyone reads.

## Tests

- `core/identity/provider/assurance.spec.ts` - the decision matrix (13 rows over amr/acr/both), the parser, and the `acr_values` rendering.
- `core/oauth2/federated-login/module.spec.ts` - the refusal maps to `access_denied` with no session and no pending login; every other authenticator failure keeps throwing.
- `http/.../identity-provider/module.spec.ts` - the attributes survive create -> read (the effective allow-list is the validator, and a key nobody mounted is stripped **silently**), and the outbound URL carries `acr_values`.
- `http/.../identity-provider/login-completion.spec.ts` - end to end: the provider is contacted, the login is refused, no pending-login cookie is set and no user is provisioned.

Both HTTP tests were confirmed non-vacuous by mutation: dropping the validator mount fails the round-trip assertion, and commenting out the gate fails the completion test (plus three collateral tests, because the refused login then provisions a user).

Full `apps/server-core` suite green (2312 tests), `packages/{core-kit,i18n,client-web-kit}` green, lint clean.

## Upstream

hapic's `buildAuthorizeURL` emits only the keys it knows and silently drops the rest, so `acr_values` is appended by hand with a removal trigger - filed as tada5hi/hapic#1071.

## Docs

`.agents/architecture.md` (*Intentional enforcement boundaries* restated as unchecked-unless-configured, plus a section on where the gate sits), `docs/src/guide/user/identity-providers.md` (a new *Verifying what the provider did* section) and `docs/src/guide/deployment/configuration-server-core-mfa.md`.

## Summary by CodeRabbit

- **New Features**
  - OAuth2/OIDC providers can require specific authentication methods or assurance levels through AMR and ACR allow-lists.
  - Configuration fields are available in the identity-provider setup form, with validation and multilingual labels.
  - Configured assurance requirements are forwarded to upstream providers during login.

- **Bug Fixes**
  - Federated logins and account linking now fail securely when required assurance claims are missing or don’t match.
  - Rejected authentication returns an access-denied response without creating sessions or provisioning accounts.

- **Documentation**
  - Updated MFA and identity-provider guidance to explain upstream assurance verification and fallback behavior.


tada5hi:
Ran a 4-lens adversarial audit over the diff (security / correctness / design / tests), each finding refuted by an independent agent before being accepted. 28 raised, 15 survived. Everything actionable is now in `555f5a3`; the notes below are what I decided *not* to change, and one thing that turned out to be someone else's problem.

## The one that mattered

`requiredAcr: "1"` bricked the provider, and the operator's natural fix disabled the gate.

The allow-lists are extra-attribute rows, and that value column round-trips through `destr`. A stored `"1"` — the canonical PAPE / ISO-29115 acr level, and the shortest value the validator deliberately allows — comes back as the **number** `1`:

```
destr("1") -> number 1
destr("0") -> number 0
destr("mfa") -> string "mfa"
```

`splitList` then called `.split()` on it. Through `buildRedirectURL` the TypeError landed inside the existing try and was swallowed into *"the identity provider is misconfigured and has an invalid or missing authorize URL"*, so authorize-out answered 400 blaming a URL that was fine — and clearing `requiredAcr` "fixed" it, which is exactly the downgrade the feature exists to prevent. Through the callback it escaped as a 500. `"0"` never even errored: falsy, so the check silently never ran.

The unit table could not see any of this, because it passes JS strings straight into the function. The round-trip is now asserted **behaviourally** — an authorize-out emitting `acr_values=1` — since that is the only form that crosses the persistence boundary. Confirmed non-vacuous by reverting the coercion.

## Also fixed

- A `null` id_token payload parses fine, so the claim reads threw a TypeError out of a callback that must answer with a refusal.
- `requiredAmr: ","` parsed to zero tokens and disabled the check while looking configured. Both mounts now require a token.
- **The accepting direction was never exercised end to end.** A gate wired to the access token instead of the id_token would have refused every login forever with a green suite, because the only integration case expected refusal anyway. Added the accepting case; verified it fails under that exact mutation.
- The account-link flow had no test, though the gate covers it by construction. Added one that also asserts the reason reaches the log (the browser only sees `link_failed`).
- Preset providers are validated by their own attribute validator, so that mount needed its own round-trip. Verified non-vacuous by removing the mount.
- acr matching semantics (membership — never prefix, containment, or coercion) are pinned by table rows instead of a comment.
- **`requiredAmr` is a disjunction, and the label did not say so.** An operator copying `["pwd","mfa"]` off a real login and pasting `pwd, mfa` would reasonably read that as "both", while a password-only login intersects on `pwd` and is admitted. The label now reads *any of* in all four locales, both fields carry a hint stating that empty trusts the provider, and the docs note that an update omitting a field clears it (true of every provider attribute, but only here does it silently turn off a security control).

The full `apps/server-core` suite caught a regression the targeted runs did not: the new accepting case provisions a second user for the same external subject under a different provider, and a later test addresses the federated subject as "the first non-admin user". It cleans up after itself now. 2323 tests green, lint clean, `check:types` clean.

## Deliberately not changed

- **No audit event or metric on this refusal.** Consistent with the federated access-policy leg beside it, which is a documented gap rather than an oversight.
- **No distinct `linkError` marker for an assurance refusal on the link path.** The issue settled this: the marker set is closed and the reason belongs in a `describeError` log line.
- **The fields stay in the OAuth2 client sub-form** rather than getting their own. They render under the Security heading, which is where they belong; a separate component costs a file plus parent wiring for no behaviour.
- **A numeric `acr` claim still fails closed.** OIDC says `acr` is a string; coercing it would be speculative. Now a pinned table row rather than an accident.
- **The id_token is decoded twice per login.** Threading claims out would change `buildIdentityWithTokenGrantResponse`'s signature, which the five presets override — the precise thing the current placement is designed to prevent.

## Out of scope, filed separately

The security lens surfaced a **pre-existing** vulnerability unrelated to this change: `GET /identity-providers/:id` is anonymous and returns the provider's `clientSecret` in plaintext. I reproduced it end to end against the test application. Filed as #3480 — it wants its own fix, and the trimmed-DTO shape that fix implies would also cover the two new fields riding that route today.

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3483, opened 2026-08-19
fix(server-core)!: gate the identity provider record read

### LINKED ISSUE #3480: Anonymous GET /identity-providers/:id returns the provider's clientSecret
`GET /identity-providers/:id` is anonymous and returns the provider's `clientSecret` in plaintext.

Found while auditing #3477; it predates that work and is unrelated to it.

## Reproduction

Against a running server, with no `Authorization` header at all:

```
GET /identity-providers            -> 200, lists every provider with its id
GET /identity-providers/<that id>  -> 200
```

The second response body:

```json
{
  "id": "3f63d2e7-...",
  "name": "dmtvgivrxskxzrus",
  "protocol": "oauth2",
  "realmId": "0c8990ec-...",
  "clientId": "Jennifer.Stoltenberg75",
  "clientSecret": "super-secret-value",
  "tokenUrl": "https://upbeat-sundae.name",
  "authorizeUrl": "https://jumbo-pepper.name/",
  "scope": "openid profile email"
}
```

Verified end to end against the test application (a spec issuing the request with no bearer, asserting on the body).

## Why it happens

Three things line up:

- `@DGet('/:id', [])` in `apps/server-core/src/adapters/http/controllers/entities/identity-provider/module.ts` carries no `ForceLoggedInMiddleware`, unlike the write routes on the same controller.
- The route resolves through `findOneByIdOrName` -> `findOneById`, which calls `extendOneWithEA(entity)`. The OAuth2 attributes are extra-attribute rows rather than columns, so that call loads `clientSecret` (and `tokenUrl`, `authorizeUrl`, `userInfoUrl`) onto the returned object. There is no attribute filter on that path.
- The `IDENTITY_PROVIDER_READ` evaluation on the route is wrapped in `catch { // do nothing }`, so a denied permission does not change the response.

The anonymous **collection** route is fine: it does not extend with EA, and the rapiq projection covers it. Only the single read leaks. But the collection route is what hands out the ids, and the hosted login page calls it, so the id is not a secret.

## Impact

The client secret authup uses to authenticate to an upstream identity provider is readable by anyone who can reach the server, including from the public internet for any deployment whose IdP is internet-facing. Depending on the upstream, that secret allows minting tokens as authup's registered client.

`tokenUrl` / `authorizeUrl` / `userInfoUrl` leak with it, which is lower impact but still internal topology for a private upstream.

## Notes on a fix

The single read is anonymous on purpose: the hosted login page renders the provider list and needs the presentational fields. So the fix is presumably a trimmed DTO for the unauthenticated case (the `ClientSummary` precedent at `/authorize`, which exists for exactly this reason), rather than putting the route behind auth.

Whatever shape it takes, the EA extension is the part that needs a deliberate allow-list: any attribute added later rides this route automatically. #3477 added two, and `requiredAmr` / `requiredAcr` being public is harmless, but they got there without anyone choosing it.

tada5hi:
Closes #3480.

## The leak

`GET /identity-providers/:id` was anonymous and returned the provider's extra attributes, which carry the OAuth2 `clientSecret` and the LDAP bind password. Three things lined up, and only one of them mattered:

- **the swallowed check (the actual bug)** — the route's `IDENTITY_PROVIDER_READ` evaluation sat inside `catch { // do nothing }`, so the permission decided nothing;
- the missing `ForceLoggedInMiddleware` — real, but cosmetic on its own: an anonymous actor holds no grants and is denied once the check is enforced;
- the ungated EA splice — deliberately left alone. `findOneById` is the same call the federated login path uses, and it genuinely needs `clientSecret`/`tokenUrl`. Making the splice conditional needs a second port method and buys nothing once the route is gated.

Enforcing the check also activates its `REALM_MATCH` term, which is what stops a realm admin reading another realm's provider: on the flat mount `getRequestRealmID(event)` is `undefined`, so `findOneByName(name, undefined)` matches any realm's row by guessed name.

`evaluateOneOf` over READ|UPDATE|DELETE mirrors `UserService.getOne`, so a realm admin holding only UPDATE can still open the admin edit page.

## The collection stays anonymous

The hosted login page lists a realm's providers before anyone signs in. That is safe because `findMany` does not EA-extend, unlike `findOneById`/`findOneByName`/`findByProtocol`, so it ships the schema's `fields.default` projection alone. An anonymous caller wanting a single provider reads it from the collection: `?filter[id]=<uuid>` or `?filter[name]=<name>`, both filterable and index-leading, same 9 columns.

Nothing in-repo used the record route anonymously. Its only caller is the admin console edit page, already logged-in and permission-gated. `getAuthorizeUri()` builds its URL locally with no request.

## Two sibling egress paths for the same value

The record route was one of three places the attribute value escaped, so the other two are closed here rather than filed separately:

- **`auth_events` recorded a rotated `clientSecret` in plaintext**, default-on, readable by any `EVENT_READ` holder. `buildEntityDiff` denylists by key *name*, and an attribute row's payload column is generically called `value`, so no regex over that name can decide what it holds. Updates to the four `*_attributes` types now record no diff at all. Fail-closed, rather than widening a regex that structurally cannot see the key.
- **the identity provider attribute subscriber shipped the row verbatim** to redis and socket consumers. It now nulls `value` on both the current and the previous payload. (Only the strip. A realm-namespaced destination was tried and reverted: `buildEntityDestinations` always keeps the global destination, so adding one only widens delivery and closes nothing.)

Also drops a dead per-row permission loop in the collection handler that evaluated a verdict and discarded it.

## Verification

- `2330 passed (203 files)`, lint and `check:types` clean.
- The new cases were each confirmed to **fail without their fix**: source reverted, re-ran, 6 failures; restored, green.
- A full-suite run showed 8 failures at one point, all `requiredAmr`/`requiredAcr` assertions. The cause was a stale `packages/core-kit/dist` predating #3479, not this branch. Rebuilding core-kit turned the suite fully green.

## Deliberately not in scope

`GET /users` splices every `auth_user_attributes` row onto the user record under `USER_READ`, while `/user-attributes` gates foreign rows behind `USER_UPDATE`. That asymmetry is real, but it is a permission-model decision rather than a credential leak, and it can be resolved in either direction (strip attributes from the user record, or let `USER_READ` list them). It does not belong in a security fix that changes a public response shape. Worth its own issue.

## Summary by CodeRabbit

* **Security**
  * Protected individual identity-provider details behind authentication and appropriate permissions.
  * Prevented sensitive provider secrets and attribute values from appearing in API and realtime event data.
  * Anonymous collection queries remain available with safe default fields.

* **Documentation**
  * Documented the updated access requirements, error responses, and migration guidance for identity-provider lookups.

* **Bug Fixes**
  * Prevented audit diffs from exposing opaque attribute values.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -2951,6 +2951,18 @@ a guaranteed 400 is worse than no button. The hosted `/authorize` page always
 passes one; a host embedding `ALoginForm` without one gets the password login
 alone.
 
+**The provider LIST is the anonymous surface, the provider RECORD is not
+(#3480).** `GET /identity-providers` stays ungated: the hosted login page
+renders its provider buttons before anyone has authenticated, and `findMany`
+ships the schema's `fields.default` projection alone.
+`GET /identity-providers/:id` requires an identity and one of
+`IDENTITY_PROVIDER_READ`/`UPDATE`/`DELETE` with a realm match, because the
+record read resolves through `findOneById` and splices the extra attributes on
+after the query (`extendOneWithEA`), so the OAuth2 `clientSecret` and the LDAP
+bind password sit outside every rapiq field gate. An anonymous caller wanting
+one provider reads it from the collection (`?filter[id]=<uuid>`), which is the
+same projection.
+
 ## Federated Identity Claims
 
 `IdentityProviderOAuth2Authenticator.buildIdentityWithTokenGrantResponse`
@@ -4406,7 +4418,14 @@ hub lacks: a **closed taxonomy** (`EventName`/`EventScope` enums in
   any key matching the secret denylist
   `/(password|secret|hash|token|credential)/i` are dropped fail-closed,
   strings truncated to 512; `sanitizeEventData`'s dedicated `diff` branch
-  re-checks the same regex at the write boundary. Created/deleted rows carry
+  re-checks the same regex at the write boundary. **A generic `(name, value)` row
+  gets no diff at all** (`ENTITY_OPAQUE_VALUE_TYPES`: the four `*_attributes`
+  types): their payload column is literally called `value`, so a key-name
+  denylist cannot see that it holds an identity provider's `clientSecret` or
+  LDAP bind password. The same reasoning applies one layer out, on the
+  realtime bus: `IdentityProviderAttributeSubscriber` overrides `publish` to
+  null the `value` on both the current and previous payload, since the
+  domain-event content is shipped verbatim to redis and socket consumers. Created/deleted rows carry
   `data: null` (no column dumps). Rows self-prune on a short per-row TTL via
   `EventRecordInput.retentionDays` (config `eventLogEntityEnabled` default
   `true` / `eventLogEntityRetentionDays` default `7` days, env
--- a/apps/server-core/src/adapters/database/domains/identity-provider-attribute/subscriber.ts
+++ b/apps/server-core/src/adapters/database/domains/identity-provider-attribute/subscriber.ts
@@ -5,7 +5,7 @@
  * view the LICENSE file that was distributed with this source code.
  */
 
-import type { IdentityProviderAttribute } from '@authup/core-kit';
+import type { EntityDefaultEventName, IdentityProviderAttribute } from '@authup/core-kit';
 import { EntityType } from '@authup/core-kit';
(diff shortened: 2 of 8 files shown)


tada5hi:
@coderabbitai pause

PR #3484, opened 2026-08-19
fix(client-account-console)!: render a retryable error state on a failed page load

### LINKED ISSUE #3482: account console: a failed page load leaves the page blank with only a toast

**Reported in:** PR #3481, review comment by @coderabbitai
**File:** `apps/client-account-console/src/pages/overview.vue#72`

## Description

Every account console page loads its data in `onMounted` and handles a failure the same way: show a toast, leave the content refs empty. There is no retry action and no distinction between "the request failed" and "there is nothing here".

- `pages/overview.vue` — the profile form is behind `v-if="entity"`, so a failed `/userinfo` load leaves the page **completely blank** but for a toast that fades. The only recovery is a manual reload.
- `pages/connected-accounts.vue` — already tracks an `errored` ref precisely so a failure does not fall through to the "no providers configured" empty state, but still offers no way back.
- `pages/applications.vue`, `pages/sessions.vue` — same shape via the kit collection's `failed` emit.

The overview page is the worst case because its fetch is the page's only content. The others at least keep their chrome.

## Suggested fix

One shared treatment for the whole console rather than a per-page patch:

- an error state per page that replaces the content with a short message plus a retry control, reusing the loader the page already has;
- route an authentication failure (401) through the session logout flow instead, since it means the session died and the router guard would clear it on the next navigation anyway.

This needs a new `authupAction` retry/reload key (no `RETRY` key exists today — the closest are `RESET` and `BACK`), authored across all four locales, which the locale-parity test enforces.

## Notes

Not fixed inside #3481: the fix belongs to all four pages, and the PR only introduced the fetch on `overview.vue` (it previously read the user out of the session store). Patching one page would leave the console inconsistent.

tada5hi:
Closes #3482.

## The problem

Every account console page handled a failed load the same way: show a toast, leave the content refs empty. The toast fades and the page it leaves behind says nothing.

- `overview.vue` was the worst case. Its form sits behind `v-if="entity"`, so a failed `/userinfo` load left the page completely blank.
- `applications.vue` and `sessions.vue` rendered the kit collection's "no entries" copy, reporting an error as an absence.
- `connected-accounts.vue` already tracked an `errored` ref to avoid that misreport, but offered no way back.

## The change

The four pages render a shared `PageError` panel in place of their content: one localized line plus a Retry control.

- `usePageError()` (`src/pages/utils.ts`) holds the error. `overview` / `connected-accounts` retry by calling `load()` again; the two collection pages retry by clearing the error, which flips the `v-if` and re-mounts the kit collection. Its initial load runs in setup, so a fresh instance re-fetches, and there is no other reload handle outside the default slot.
- The error is cleared **on success**, not before the attempt, so a retry keeps the panel on screen instead of dropping back to a blank page (overview) or to the "no providers configured" empty state (connected-accounts) for the duration of the request.
- Action failures (a revoke, a disconnect, a form submit) keep toasting. They leave the page intact, and their server message is the actionable one ("already linked to another user").

**A 401 is not offered as retryable.** `capture()` runs `store.logout()`, which drops the shell for the sign-in state, the same place the router guard's failed `resolve()` lands. The kit's auth hook already covers the common case (a failed background refresh unsets the header and the kit answers with a logout); this catches the request that still 401s after it, and the case where the store holds no refresh token to retry with at all.

**The panel does not render the failure's own message.** A page load here fails for reasons its reader cannot act on, and the messages carry the request line: hapic builds `Failed to fetch (GET <url>)` and `502 Bad Gateway (GET <url>)` from the request itself, so the API url, the subject uuid and the rapiq filter expression ride along. That was tolerable in a toast that faded and is not in a panel that stays. Note that gating on the error's `code` does **not** separate the two: hapic stamps `ECONNRESET` on a transport failure and `HTTP_RESPONSE_ERROR` on a status whose body is not JSON.

**The nested token inventory flags per row.** `sessions.vue` hosts a second collection inside each expanded row. It had the same defect, but a page-level error there would take the whole session list down for one row's failure, so it gets its own scoped panel and retry.

## i18n

New `authupAction.RETRY` and `authupCommon.LOAD_FAILED`, authored in all four locales (the parity test enforces it). `@vuecs/elements` is now declared: the app already imported `VCAlert` from it in `pages/index.vue`.

## Verification

Built, linted, `vue-tsc` clean, unit tests green (17 account console, 77 i18n parity), full `nx` build across 24 projects.

Driven end to end in a browser against a real server-core (sqlite), for each state:

| state | before | after |
|---|---|---|
| overview, API down | blank page + fading toast | panel + Retry |
| applications, API down | "You have not granted any application access yet." | panel + Retry |
| connected accounts, API down | fading toast, empty list | panel + Retry |
| sessions, API down | empty list | panel + Retry |
| expanded session row, API down | empty token inventory | panel + Retry **inside that row**, list intact |
| Retry once the API is back | n/a | content loads |
| session revoked server-side | n/a | hosted login, not a dead-end Retry |

## Known limitations

- A non-initial collection load failure (a pagination click) replaces the list rather than keeping the rendered page. That is deliberate: the alternative is silently showing page 1 while the user asked for page 2.
- `AUserAuthenticators` (the Authenticators tab) keeps its own inline alert with no retry, and `ARealmGrid` its own error handling. Both are kit components shared with the admin console and are out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Added localized inline page-load error states with retry actions across account console pages.
  * Added retry support for collection and session token loading failures.
  * Authentication failures without a renewable session now automatically sign users out.
  * Added English, German, Spanish, and French translations for loading errors and retry actions.

* **Bug Fixes**
  * Improved handling of page and row-level loading failures without relying solely on toast notifications.
  * Corrected HTTP responses for invalid access tokens and refresh-token failures.

* **Tests**
  * Added coverage for retryable errors, authentication failures, token handling, and error normalization.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

tada5hi:
## Follow-up: the 401 gate was weak in both directions

Reviewing `usePageError`'s `status === 401` check turned up two problems, both fixed here.

### 1. It could not see the case it was written for

`JWT_EXPIRED` / `JWT_INACTIVE` / `JWT_INVALID` were **absent** from `ERROR_CODE_TO_STATUS`, so they took the unlisted-code 400 fallback. Probed against a live server before the fix:

| request | status | code |
|---|---|---|
| no `Authorization` header | 401 | `identity_unauthorized` |
| garbage bearer | **400** | `invalid_token` |
| malformed 3-segment bearer | **400** | `invalid_token` |

So a *dead* bearer answered 400 while a *missing* one answered 401 — no client could tell "your credential died" from "your request was malformed" by status alone, and the console's branch only ever caught the missing-header case (where the kit had already logged out). The three codes now map to **401** per RFC 6750 §3.1.

The token endpoint is the exception RFC 6749 §5.2 carves out, and the global map cannot know which surface it is on, so it is handled at the call site: the refresh grant catches a verification failure and re-throws `OAuth2GrantError.invalid()`. **The full suite did not catch this** — 2323 tests passed with `/token` regressed to 401; a live probe found it. It also removes a split the endpoint already had, where an unparsable refresh token answered `invalid_token` while a wrong-kind one answered `invalid_grant`.

After, verified live:

```
resource:  no header 401 identity_unauthorized | garbage 401 invalid_token
token:     garbage 400 invalid_grant | 3-seg 400 invalid_grant | wrong-kind 400 invalid_grant
           valid refresh -> 200
```

### 2. It logged out sessions that were still renewable

An expired access token is **not** a dead session while a refresh token answers for it, and renewal belongs to the kit's auth hook. The gate now requires the session to be *unrenewable*:

```ts
if (extractErrorContext(e).status === 401 && !store.refreshToken)
```

The one thing the hook cannot answer is a store holding no refresh token: `refreshSession` throws before anything is emitted, so no logout follows and a retry would fail on every press. That is the only case left.

### Verification

Full server-core suite green (2325), plus errors / specs / server-adapter-kit / client-web-kit / i18n / account console. New tests pin both halves and were mutation-checked (inverting each guard fails them). Two existing specs moved 400 → 401, both on resource routes, which is the intended change.

**Breaking:** a resource request with an expired/revoked/unverifiable bearer now answers 401 instead of 400 (code unchanged). The token endpoint still answers 400, but an unparsable refresh token now carries `invalid_grant` rather than `invalid_token`. Title marked `!` accordingly.

### Unrelated, not chased

While probing I saw a revoked session's access token keep working for 90s+ against `/users/@me`, and `/token/introspect` report `active: true` for it. Not touched here and not claimed as a bug, but it does not match my reading of the architecture doc's "authup's own API rejects immediately via the authorization middleware's session check" — worth a separate look.

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3492, opened 2026-08-23
fix(server-core): make the just-in-time dev mode reachable

### LINKED ISSUE #3382: server-core: the just-in-time dev mode is unreachable from every script
Split out of #3380, which repaired the JIT branch but left it unreachable.

## What

`isCodeTransformation(CodeTransformation.JUST_IN_TIME)` gates two things in server-core:

- `adapters/http/middleware/built-in/assets.ts:40` decides between mounting the auth console's prebuilt dist and creating an embedded vite dev server.
- `adapters/http/ui/auth-console/module.ts:37` decides between reading `dist/server/server.js` and loading `/src/server.ts` through vite.

typeorm-extension returns `jit` only under ts-node or tsx. Every server-core script runs compiled JS from `dist/`:

| script | command |
|---|---|
| `cli` | `node dist/cli/index.mjs` |
| `cli-dev` | `npm run build:server && npm run cli` |
| `cli-dev-src` | `nodemon --watch src --exec "npm run build:server && node dist/cli/index.mjs"` |
| `debug` | `node --inspect dist/cli/index.mjs start` |

The documented development command (`docs/src/guide/development/quick-start.md`) is `cli-dev`. So the vite branch has never run in any documented workflow, which is why the `exports is not defined` breakage #3380 fixes went unnoticed.

## Why it matters

Auth console development currently means a `vite build` plus a server-core restart, because server-core memoizes the dist template, manifest and render function for the process lifetime. The JIT path exists to avoid exactly that, and now works, but nothing can reach it.

## What is needed

1. A way to run server-core from TypeScript source. `ts-node` is already a devDependency but its ESM support is awkward for this app; `tsx` is the more likely choice.
2. Verification that the whole boot path survives it, not just the vite branch. The same `CodeTransformation` gate drives typeorm entity and migration loading, so this is not only a vite concern.
3. A decision on whether the gate should stay implicit (runtime detection) or become an explicit opt-in. Implicit detection is what made the branch invisible; an explicit flag would need a `Config` key and a `ConfigValidator` mount.
4. Documentation, so the mode has a stated entry point.

If nobody wants to own the dev loop, the honest alternative is deleting the branch, which would also drop `vite` from server-core's devDependencies and simplify both files. Worth deciding rather than leaving a repaired path nobody runs.

tada5hi:
Closes #3382.

## What

- `cli-dev` in `apps/server-core` now runs the CLI from TypeScript source: `node --loader ts-node/esm --require ts-node/register src/cli/index.ts`. It is the command `docs/src/guide/development/quick-start.md` already documents, so the JIT dev mode has a stated entry point without a new script name.
- `registerAssetsMiddleware` logs at boot which branch serves the auth console (`source (vite dev server)` / `the package dist`). The gate stays implicit, so this is what makes a dark gate visible; #3380 was found broken only because someone read the code.
- Docs: quick-start (what `cli-dev` does and how to run the built server instead), the architecture paragraph that described the branch as unreachable, a testing.md note that the suite always takes the dist branch, and `.agents/references/locter.md` for the runtime probes consulted.

## Why this shape

The issue asked for a runner, a boot-path verification, and a decision on implicit vs explicit gating. Three runners were tried from source:

| runner | module graph | gate (`isCodeTransformation(JIT)`) |
|---|---|---|
| tsx | dies on the first entity: esbuild emits no `design:type`, and 219 `@Column`s have no explicit type (`ColumnTypeUndefinedError`) | true, but moot |
| `@swc-node/register` | boots (needs a scratch shim) | false, nothing detects it |
| `node --loader ts-node/esm` | boots, metadata honoured | **false**: ts-node sets its marker on the loader thread (Node >= 20), never on the main process |
| the same `+ --require ts-node/register` | boots | **true**; the `--require` does nothing but set the marker on the main thread |

So the issue's "tsx is the more likely choice" is inverted, and the already-present `ts-node` devDependency is the runner. The `--require` half is a workaround for a locter gap, filed as tada5hi/locter#884 (its removal trigger); detection already lives in locter, typeorm-extension merely wraps two of its probes, so no new package.

Gate design: kept implicit. Deriving it from `CODE_PATH` (src vs dist) flips true in every vitest worker (`CODE_PATH` is `<pkg>/src` there), which would spawn a vite dev server per `createTestApplication` suite. An explicit config key is ~40 lines across ten files for a switch that is monorepo-only by construction (`vite` is a devDependency, a published install carries no auth console source).

## Verified

From a clean worktree, sqlite scratch DB:

- `npm run cli-dev -w apps/server-core -- start` boots (~12s) and logs `Serving the auth console from source (vite dev server)`.
- `GET /register` is SSR'd from the auth console source through vite (`/public/@vite/client`, `/public/src/client.ts`).
- Editing `apps/client-auth-console/src/pages/register.vue` shows on the next request without a restart; reverting it disappears again.
- SIGTERM releases both the http port and vite's HMR websocket port (24678).
- The built CLI (`npm run cli -- start`) still takes the dist branch and logs it.
- `cli-dev -- migration run` against a scratch postgres database applies all 17 migrations from `src/*.ts` through the loader.
- `npm run test -w apps/server-core`: 204 files, 2359 tests, green.

## Known edges (documented in architecture.md)

- The `--loader` flag prints Node's `ExperimentalWarning`; the `--import` + `module.register` form avoids it and behaves the same, but still needs the `--require`.
- Transpile-only mode (~3s boot) is blocked by TS 6's `baseUrl` deprecation (TS5101) until the root tsconfig carries `ignoreDeprecations`; the type-checked boot is what ships.
- The dev render has no `<link rel="stylesheet">` (CSS rides the module graph) and emits two benign `[Vue warn] onScopeDispose()` lines per render.
- Two dev instances on one machine collide on vite's `*:24678`.

## Summary by CodeRabbit

* **New Features**
  * Development mode now serves the authentication console directly from source with improved startup visibility.
  * Added clearer guidance for running the development server, including initial setup and expected loader warnings.

* **Documentation**
  * Documented development-server behavior, authentication console serving paths, testing coverage, runtime limitations, and troubleshooting guidance.
  * Added reference documentation for runtime environment detection and module loading behavior.


tada5hi:
The locter side is up as tada5hi/locter#885 (closes tada5hi/locter#884). Once it is released and typeorm-extension/authup are bumped past it, the `--require ts-node/register` half of `cli-dev` can be dropped, plus the matching sentences in `.agents/architecture.md` and `.agents/references/locter.md`.

tada5hi:
@coderabbitai pause

tada5hi:
locter 4.1.1 shipped tada5hi/locter#885, so the follow-up landed in this PR after all (7f00bf520): `cli-dev` is now plain `node --loader ts-node/esm src/cli/index.ts`, and server-core's `locter` floor moved to `^4.1.1` (load-bearing: below it the gate is dark under the loader-only form). Verified: loader-only boot logs the source branch and serves /register through vite; the built CLI still takes the dist branch.

PR #3493, opened 2026-08-24
feat(server-core)!: require authorization on the token introspection endpoint

### LINKED ISSUE #3489: Introspection endpoint requires no authorization (RFC 7662 §2.1)
`POST /token/introspect` (and `GET /token/introspect`) mounts no authorization: `@DController` registers it as `@DPost('/introspect', [])`, and the route answers a request carrying no credentials at all.

RFC 7662 §2.1 makes this a MUST:

> To prevent token scanning attacks, the endpoint MUST also require some form of authorization to access this endpoint, such as client authentication as described in OAuth 2.0 [RFC6749] or a separate OAuth 2.0 access token such as the bearer token described in OAuth 2.0 Bearer Token Usage [RFC6750].

`POST /token/revoke` has the same shape. RFC 7009 §2.1 is weaker there ("The client also includes its authentication credentials as described in Section 2.3 of [RFC6749]"), but the endpoint is equally open.

## Why it matters

Anyone who holds a token string can introspect it and receive the subject id, realm, session id, the resolved permission set and the subject's OpenID claims, without authenticating as anything. Holding a live token is already close to game over, so the marginal disclosure there is small — but §2.1 exists to stop *scanning*, and nothing rate-limits or gates the endpoint today.

PR #3488 widens what an unauthenticated caller can learn slightly: an expired token is now read and reported with its payload and claims, where it previously answered 401. An old, expired token that was useless to a holder now discloses who it belonged to. That change is deliberate and wanted (a relying party can tell the person whose session ended who they were), and §2.1 is exactly the control that would make it safe.

## The part that makes this non-trivial

The obvious fix — require an authenticated identity on the route — collides with how the endpoint is actually called.

`@authup/client-web-kit`'s store introspects with `Bearer <the token being introspected>` (`packages/client-web-kit/src/core/store/create.ts`, `fetchTokenIntrospection`). So:

- Requiring a valid **bearer** would reject exactly the expired-token case the endpoint was just changed to serve, since the bearer *is* the expired token.
- Requiring **client authentication** (`client_id` + `client_secret`) would lock the kit out entirely: the consoles authenticate against public, secret-less clients (`admin-console`, `account-console`).

So closing this needs a decision about who is allowed to introspect, not just a middleware.

## Options

1. **Client authentication, and move the kit off introspection.** Conformant, and it makes the fat inactive response safe (the caller stops being "a third party", which is the §2.2 / §4 SHOULD NOT's own rationale). Cost: the kit store currently derives realm, subject, permissions, `session_id` and `acr` from the introspection response, so it would need another way to resolve a session. Breaking for any existing integration calling the endpoint anonymously.
2. **Accept a bearer OR client credentials, and allow a self-introspection exemption.** A caller may introspect a token it presents as its own bearer even when that token is expired; anything else needs credentials. Keeps the kit working and closes anonymous scanning. More surface to get right.
3. **Bearer required, and drop the expired-payload feature.** Simplest gate, but it removes the reason the endpoint was changed.

Option 2 looks closest to intent, but it is a design call.

## Scope

Pre-existing — this predates PR #3488, which only widened the disclosure. Deliberately not fixed there.

tada5hi:
Closes #3489.

## What

Two layers on `POST`/`GET /token/introspect`, per RFC 7662 §2.1 ("to prevent token scanning attacks, the endpoint MUST also require some form of authorization"):

**Layer 1, WHO may call** (`assertIntrospectionAuthorized`), an **independent credential**:

- a **request identity** the global authorization middleware resolved: a live bearer (the introspected token itself, the kit's shape, or the caller's own) or Basic auth;
- or **confidential client credentials**, read exactly the way the grants read them (`extractClientCredentialsFromRequest` + `OAuth2ClientAuthenticator`, realm hint via `readRealmHint`). A resolved `authMethod: none` client is refused with `invalid_client`: a bare public `client_id` identifies and proves nothing.

Nothing at all answers `401 identity_unauthorized` with the bare `WWW-Authenticate: Bearer` challenge.

**Layer 2, WHOSE tokens** (`isIntrospectionAllowed`): the token's own subject, the client the token was issued for (its `client_id` claim), or an actor granted the new **`token_introspect`** permission, whose realm reach is matched against the token's realm claim (`admin` reaches everything at `any`; a default grant covers the caller's own realm). A caller failing all three receives the bare `{"active": false}` §2.2 prescribes for a resource "not allowed to introspect" (indistinguishable from a dead token, no oracle); the denial is logged server-side, since the response gives the caller nothing to diagnose with. The permission is a `PermissionName` entry and auto-provisions like every other.

## Design decisions (settled with the maintainer)

- **The issue's option 2 (self-introspection exemption for an expired bearer) cannot exist.** Possession of the token string is exactly what a finder has, and the global middleware already answers 401 to an expired bearer before any route runs (now pinned by a spec). The kit never consumed the expired report anyway: it throws on `active: false`, and its expired call already fell through to `refreshSession()`.
- **The expired-token report (#3488) stays**, now reachable only by a caller that proved who it is and may introspect that token.
- **`POST /token/revoke` deliberately stays open** (documented on the handler as authup's own choice, not an RFC 7009 requirement): §2.1 asks a public client to identify itself by `client_id` and the server to verify token ownership, but a public `client_id` proves nothing, the consoles are public clients whose kit revokes anonymously during logout teardown, and the uniform 200 leaves a scanner nothing to learn.
- Cohort posture (recorded in `.agents/references/`): Keycloak is client-auth only and refuses public clients; Authentik requires `client_id` + secret and is same-provider-scoped. Authup additionally accepts a bearer (RFC 7662 names it; the secret-less console clients have nothing else) and mirrors the same-provider bound through the issued-for rule, with the permission as the escape hatch.

## Callers

- `@authup/client-web-kit`: unchanged (introspects its own live token as the bearer; self rule).
- Downstream RPs introspecting tokens issued to their own client: no grant needed (issued-for rule).
- **Resource servers using `@authup/server-adapter-*` remote verification: action required** — grant `token_introspect` to the client behind the verifier's `creator`, or foreign tokens verify as inactive. Documented in the upgrade guide and all three adapter pages.
- Anonymous integrations are broken on purpose (BREAKING CHANGE in the commits).

## Tests

- `introspect.spec.ts` (25 cases): authorization matrix for both layers — anonymous 401 + challenge, expired-token-as-own-credential 401, GET gated, bare public `client_id` refused, wrong secret refused, body + Basic client credentials accepted, live bearer self and foreign, adapter-shaped lazy hook authentication, non-string realm hint ignored, foreign token denied bare without the grant, issued-for client allowed without it, an `own`-reach grant bounded to its realm.
- `mfa-login-ticket.spec.ts`: the MFA ticket is refused as an introspection credential.
- Mutation-checked: the `authMethod: none` refusal, the hook replay, the `readRealmHint` guard and (via the realm-reach case) the `realmMatch` wiring each have a test that fails without them.
- Full server-core suite green (204 files / 2374 tests), core-kit suite green, lint + `check:types` + build clean.

## Summary by CodeRabbit

- **Changes**
  - Token introspection now requires authorization for the requested token.
  - Callers may introspect their own tokens, tokens issued to their client, or permitted realm tokens.
  - Supported authentication includes bearer tokens, confidential client credentials, and TLS certificates.
  - Unauthorized or inaccessible tokens return an inactive response.
  - Revocation remains available without authorization; invalid or missing parameters continue to return errors.

- **Documentation**
  - Added upgrade guidance and clarified token endpoint authorization behavior.

- **Tests**
  - Expanded coverage for authorization, credentials, bearer inheritance, expired tokens, and invalid scenarios.


tada5hi:
@coderabbitai pause

PR #3497, opened 2026-08-24
feat(server-core): optional worker role

tada5hi:
## Summary

Implements plan 095: an optional worker role for server-core. Same binary, role by command. With no worker deployed nothing changes; deploying one moves the background components (the cron sweeps) into a dedicated process.

- `authup-server worker`: a new CLI subcommand booting a minimal module graph (config, logger, cache, database, components). No HTTP listener, no provisioning, no mail.
- `COMPONENTS_ENABLED` (default `true`): disables the background components on API replicas when a worker runs them. The worker role runs them regardless, so a shared env block across containers works.
- `MIGRATION_ENABLED` (default `true`): when `false`, startup applies no migrations. It verifies none are pending and fails loud otherwise, so a migration Job can be the only DDL path and the app DB user can lose DDL rights. The worker role never migrates, regardless of the flag. sqlite is exempt: it has no migrations and keeps synchronizing in both roles.
- The expired session and session-token sweeps are now batched at the repository layer (select ids, delete, stop on an empty batch), the shape the event sweep already had. The old unbatched delete path is gone.
- Deployment surfaces: entrypoint usage example, a `MIGRATION_ENABLED=false` boot phase in `verify-latest-migration.mjs`, a worker-mode deployment guide, config key docs, and the `.agents` updates.

Out of scope by decision: the `authup` launcher's third child (follow-up PR), helm chart changes (follow-up in authup/helm), and any job queue (plan 096 records that design as a triggered contingency; it has zero tenants today).

## Verification

- Full monorepo test suite green (20 projects; server-core 208 files / 2383 tests), `check:types` clean, eslint clean.
- Real `worker` boot verified: SIGTERM round-trip exits 0; the sqlite path verifies, then synchronizes.
- The `MIGRATION_ENABLED=false` boot runs in the `tests-migrations` CI job against a genuinely migrated schema (it needs live databases, so it is CI-verified).

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated worker command for background cleanup tasks without an HTTP listener.
  * Added configurable controls for background components and startup migration behavior.
  * Added batched cleanup for expired sessions, tokens, and events.
  * Added graceful shutdown handling for long-running processes.

* **Documentation**
  * Added worker deployment guidance and configuration examples.
  * Documented the new worker command and updated Redis configuration examples.

* **Bug Fixes**
  * Improved schema verification when migrations are disabled, including SQLite synchronization.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3498, opened 2026-08-24
feat(server-core)!: authenticate the account console with an opaque session cookie

tada5hi:
Plan 088 Stage 1. The account console authenticated to the API the way a third-party RP does: a JWT bearer in a JS-readable cookie. It now presents an opaque `HttpOnly` credential naming an existing `auth_sessions` row, and the code exchange moves server-side so no token reaches JavaScript.

Bearer stays the contract for every other consumer. Cookie mode is opt-in, and only a surface served from the API's own origin can present the credential, so the restriction is enforced by the browser rather than by remembering to check.

## Surface

Two routes on the **existing** `SessionController` — no new controller, and sign-out needed no new route:

- **`GET /sessions/@me/introspect`** — sibling of `@me`, which returns the row; this returns the *context* a console hydrates from (identity claims + permission projection), i.e. what `POST /token/introspect` answers minus everything token-shaped.
- **`DELETE /sessions/@me`** — already revoked the current session. The only reason the plan rejected it for sign-out was that it left a dead cookie the browser kept presenting; it clears the cookie now.

Per-console `GET /account/login|callback` stay on the account controller, because they genuinely differ by OAuth2 client and redirect target.

The permission read passes **no client scope**, which is what lets the route be console-neutral: `reduceBindingsByIdentityClient` keeps only permissions whose own `clientId` matches, so naming a console's client drops every global permission. A token is issued to one client; a session is not.

## What it buys, stated honestly

Immediate revocation and ~1.5-4 KB less per request are unqualified wins. The XSS story is narrower and the plan says so: an ambient origin cookie still lets an XSS act as the user while the page lives. What it stops is exfiltration of a *portable* credential — **and only for a user who cannot mint credentials through the API**. A user holding `CLIENT_CREATE` can still be driven through ordinary CRUD to `POST /clients` with a chosen secret and then a `client_credentials` grant, which touches no deny-listed route. No deny-list can close that; tracked as #3499.

## Two properties that each nearly sank it

Both found by adversarial review and verified against the code.

**The primary flow never reached cookie mode.** The kit pins `COOKIE_PATH = '/'`, so the hosted login's tokens are visible to the console; seeding them gave it a bearer and header-wins precedence meant the cookie branch never ran. Cookie mode now seeds nothing — and deliberately does **not** clear those cookies, which are the hosted SSO session `prompt=none` / `prompt=select_account` depend on for every other RP.

**The cookie converted back into a bearer.** `POST /authorize` is `ForceUserLoggedIn` and reads its subject from `useRequestIdentityOrFail`, so one script execution could mint a code with its own PKCE challenge and trade it at `/token`. Cookie authentication never reaches the OAuth2 issuance surface (`/authorize`, `/token*`, `/logout`), matched on a path normalized the way routup routes. An invariant, not a detail: `HTTPOAuth2IdentityGrantType` is unregistered today but mints a grant straight from the request identity.

## The gate

`SameSite=Strict` + an `Origin` check against publicUrl + a mandatory `Sec-Fetch-Site: same-origin` that **fails closed when absent** — the last is what `SameSite` cannot carry, being registrable-domain scoped. `Allow-Credentials` narrows to publicUrl's origin. `trustedOrigins` is deliberately not consulted: it grants a consented redirect target, not ambient authority.

Cookie mode requires **both** the server vouching for the routes (injected capability) and the client confirming same-origin (applicability). Either alone is wrong: injected-only made the broken pairing representable and silently fatal; derived-only assumed a capability it cannot verify.

## At rest

`auth_sessions.secret` stores a **SHA-256 digest**, not the credential — `select: false` governs ORM projection, not what the table contains. Plain SHA-256 rather than a keyed HMAC: keying defends low-entropy inputs, and this is ~248 bits of nanoid.

## Session lifetime

Cookie mode removes the refresh grant, which is what slides `expiresAt`. The cookie branch calls `refresh()` on a throttle **and re-arms the cookie** — without that the slide is invisible to the browser and an active user is signed out at a hard cap. Pinned by a test verified to fail without the fix.

## Verification

- Build 24/24 · server-core **2402 tests** · kit 241 · core-http-kit 24 · account console 25 · eslint clean · `check:types` clean
- Migration `1787571998009-SessionSecret` replays run → revert → run on **both** MySQL and Postgres; schema-drift gate clean on both. DDL CLI-generated, committed as emitted.
- **Verified in Chrome** against a real server: deleting every JS-readable token cookie left a gated route returning 200; the session survived reload; `POST /authorize`, `/authorize/` and `/AUTHORIZE` all returned **401 with no code minted** while the control route stayed 200; sign-out produced a real server-side 401.

## Behaviour notes

Sign-out answers `202` (the entity delete's status). A dead credential answers `401` rather than `{active: false}`, since the route carries `ForceLoggedIn` like its siblings — the kit store already reads a 401 there as the no-session answer.

## Known limits

`loggedIn`, `usePermissionCheck` and the socket manager read `accessToken` and stay inert in cookie mode. Verified the account console uses none of them; the admin console gates on permission checks in 25 files, so plan 088 Stage 2 (gated on plan 081) must re-derive them. A browser not sending Fetch Metadata can never authenticate — intended, and a support-visible cliff. Standalone cross-origin hosting stays on the JS-token path.

BREAKING CHANGE: `IClient` gains a required `account` member.


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

PR #3503, opened 2026-08-25
feat(server-core,client-web-kit)!: mount the consoles under /console/{admin,account,auth}

tada5hi:
## Summary

Plan 099 PR 1 (follows #3502).

- The served consoles move under one prefix: admin console `<publicUrl>/console/admin`, account console `<publicUrl>/console/account`, auth console assets `<publicUrl>/console/auth/assets/` (vite base `/console/auth/`). The auth **pages** stay on their protocol routes (`/authorize`, `/logout`, `/register`, `/activate`, `/password-forgot`, `/password-reset`): discovery advertises them, mail deep links target them, they share GET-HTML/POST-JSON paths, and the cookie deny-list pins them. `/console` and `/console/auth` answer 404.
- server-core spells the three mounts once (`adapters/http/ui/constants.ts`); controller mounts, login cookie scopes, callback URLs, asset mounts, vite-base rebases and the identity-provider link callback redirect derive from them. The empty `PACKAGE_PATH/public` mount is deleted; the auth console mount serves `dist/client/assets` only.
- Static consoles default `basePath` to the new mounts and strip the full two-segment suffix in the standalone API-URL derivation; the kit's `buildConsoleLoginURL` emits `/console/<name>/login`.
- Gate fix: `AccountController` honours `ACCOUNT_CONSOLE_ENABLED` on `/login` and `/callback` like `AdminController` does; a disabled account console no longer mints pending logins and session secrets on a direct hit.
- Docs swept (16 pages, 4 READMEs, `.agents/*`); new `docs/src/guide/deployment/console-replicas.md` (console replica set on the one origin: Redis required, both flag pairs on both sets, what the flags do not unmount); `upgrading.md` entry.
- The launcher smoke runner fetches one entry script per console shell (and drains it), so a bundle built for another base fails the packed smoke instead of serving a blank console.

## Breaking

`/admin`, `/account` and `/public` move to `/console/admin`, `/console/account` and `/console/auth/assets` with no redirect. Rebuild and redeploy every console together with the server (a dist built for the old base serves a blank console); update proxy rules and bookmarks. `buildConsoleLoginURL`'s output changed, so kit and server must be on the same release. Standalone hosts injecting their own `basePath` keep working; only the defaults changed. Helm chart follow-up: authup/helm (plan 098 A1) routes `/console/**`.

## Test plan

- Full server-core suite (sqlite): 212 files, 2449 tests pass. Kit: 34 files / 247 tests. Admin console 30, account console 26, auth console 4.
[...]

PR #3502, opened 2026-08-25
fix(server-core)!: read the CLI meta from the package, refuse stray positionals on start/worker

tada5hi:
## Summary

- `createCLIEntryPointCommand` read `process.cwd()/package.json` for citty's meta, so `authup-server --help` (and the quick start `npx authup start` in a fresh directory) died with ENOENT in any directory without one. It now reads `PACKAGE_PATH/package.json`.
- citty 0.2.2 never throws on undeclared positionals, so `authup-server start client.admin-console` (a leftover launcher selector) booted a server. The root command's `setup` now refuses a stray positional on `start` and `worker` and exits 1 having started nothing. The guard lives on the root command on purpose: citty re-parses the subcommand tail against the subcommand's own (empty) arg list, so the documented space form `start --configDirectory <path>` lands `<path>` in the subcommand's `_` while the root's `_` is `['start']`. `migration` keeps its required positional.
- The `-cD` / `-cF` aliases are dropped from `CLI_CONFIG_ARGS` and the docs: citty parses a two-character short alias as `-c -D`, so the documented spelling never set anything (the undocumented `--cD` / `--cF` did, and is gone too).

First step of plan 099 (console mounts / launcher retirement / console role).

## Breaking

`authup-server start|worker` refuse positional arguments; the undocumented `--cD` / `--cF` spellings are removed. Use `--configDirectory` / `--configFile`.

## Test plan

- `apps/server-core/test/unit/cli/module.spec.ts`: meta pinned to `@authup/server-core`; the guard fed by citty's real root parse (`start client.admin-console` / `worker x` refused; `start --configDirectory <path>` space form, `--configFile`, `migration status`, `migration run --configDirectory ...` accepted); end-to-end `runCommand` rejections for `start` and `worker`.
- `apps/server-core/test/unit/cli/index.spec.ts`: spawns `node dist/cli/index.mjs --help` from an empty temp dir, expects exit 0 + usage.
- `npm run test -w apps/server-core -- test/unit/cli`: 3 files, 16 tests pass; `check:types`, build and eslint clean.
- Manual from `mktemp -d`: `--help` exit 0; `start client.admin-console` exit 1 with `Unexpected argument "client.admin-console" for command "start".`, nothing listening on 3001.

## Summary by CodeRabbit

- **Bug Fixes**
  - The CLI now rejects unexpected positional arguments for `start` and `worker` commands.
  - CLI metadata is loaded correctly regardless of the current working directory.
[...]

PR #3504, opened 2026-08-25
feat(server-core): console role

tada5hi:
## Summary

Plan 099 PR 1b, stacked on #3503 (retarget to `master` once that merges).

`authup-server console [admin|account ...]` is a third long-running role next to `start` and `worker`: the IdP surface without the management API, serving the static consoles. Same binary, same image (`entrypoint.sh` forwards the subcommand; only its usage text changed).

- Any request may land on a console replica and is answered correctly, except management-API routes, which 404: an admin-UI pod that cannot answer `/users` even when misrouted.
- `createConsoleApplication()` is `createApplication` minus provisioning and the cron components, with the worker's verify-only schema step (never migrates). Mail and LDAP stay: the workflow controllers resolve the mail client at mount, the password grant needs LDAP.
- `HTTPModule({ managementApi: false })`: `mount()` builds every controller and filters by class through `classifyController`, which reads two explicit lists in `http/modules/classification.ts` and throws at mount for a controller on neither, so a new controller has to be classified. `classification.spec.ts` walks the controllers barrel and fails when a class is on neither or both lists.
- The positionals are sugar over the two console flags: `console admin` forces `adminConsoleEnabled=true` and `accountConsoleEnabled=false`, `console account` the inverse, both names both on, no name leaves them as configured. `auth` is refused with a message (the auth pages are the issuance surface and ride every role); unknown names are refused. Overrides ride a second parameter of `createCLIConfigModule`, spread over the normalized config.

## Keep/drop audit

The unit of exclusion is the controller. Everything the hosted auth pages self-call during render, the console login, the session sign-in/out routes and the protocol routes stay.

| Controller | Role | Why |
|---|---|---|
| Authorize, Token, Jwk, OpenID, Logout, UserInfo | keep | protocol surface; `ConsoleLogin.callback` redeems through the loopback `/token`; `/userinfo` is in discovery |
| Register, Activate, PasswordForgot, PasswordReset | keep | workflow pages (GET html + POST json on one path) |
| AuthenticatorChallenge | keep | second factor on the authorize ladder, MFA-ticket completion |
| Account, Admin, Status | keep | console shells + sign-in routes; `GET /` is the image HEALTHCHECK |
| Realm | keep | `/realms/:id/.well-known/openid-configuration`, `/realms/:id/jwks`, the login form's realm picker |
| IdentityProvider | keep | anonymous provider list on the login page; `authorize-out/in`, `login-complete`, `link-request/confirm` |
| Client | keep | kit `AAuthorize` reads `GET /clients/:id` when a host passes `clientId` (fallback path) |
| ClientScope | keep | `AuthorizeScopes` reads `GET /client-scopes` when `scopesAvailable` is absent (same fallback shape) |
| Consent | keep | consent covering probe before auto-consent |
| Session | keep | `GET /sessions/@me/introspect` + `DELETE /sessions/@me` (console sign-in/out) |
| UserAuthenticator | keep | inline MFA enrollment (`/users/@me/authenticators`, `/confirm`, recovery lookup) |
| User | drop | no hosted page reads `/users/@me` (the store derives the user from introspection; the profile form loads `/userinfo`); the served consoles' own API calls are routed to the API set |
| IdentityProviderAccount, SessionToken | drop | account console pages: API calls of a served SPA, not self-calls |
| Role, Permission, Policy, Scope, Key, TrustAnchor, Event | drop | entity CRUD |
| ClientPermission, ClientRole, RolePermission, UserRole, UserPermission, PermissionPolicy, IdentityProviderRoleMapping, RoleAttribute, UserAttribute | drop | junction/attribute CRUD |

## Docs

`deployment/console-replicas.md` rewritten around the role (the flag-only split stays under "Without the role"; Redis required, both flag pairs on both sets, migration run, sqlite cannot split, what stays served everywhere), `configuration-server-core.md` command list, `worker.md` cross-link, `entrypoint.sh` usage, `.agents/{architecture,structure,testing}.md`.

## Notes

- `/docs` (swagger) on a console replica still describes the routes it 404s; `middlewareSwagger=false` turns it off. Left as is.
- `mount()` builds all 40 controllers and filters instances (`ponytail:` the factories construct services over registered adapters and do no I/O, so the flat list stays the one place a controller is added).
- The root positional guard from #3502 covers `start`/`worker` only; `console` validates its own positionals and must stay out of that set when the branches meet.

## Test plan

- Full server-core suite (sqlite): 215 files, 2468 tests pass; `check:types`, build, eslint clean.
- `test/unit/http/console-role.spec.ts` boots the console preset: `/console/admin` shell 200, `/authorize` renders, `POST /token` client_credentials issues, `/sessions/@me/introspect` answers, `GET /` answers, `/users` / `/roles` / `/keys` 404; `console admin` makes `/console/account/login?realmId=` render the disabled shell while `/console/admin/login?realmId=` kicks.
- `test/unit/app/factory.spec.ts` pins the console module list; `cli/{console,config,module}.spec.ts` cover the selectors (`auth` and unknown refused) and the config overrides.
- Manual boot `node dist/cli/index.mjs console admin --configDirectory <scratch>`: `/console/admin` 200, `/users` 404, `/` 200 with `"adminConsole":true,"accountConsole":false`, SIGTERM exits 0 in under a second.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated `console` server role for serving console and authentication pages without management API routes.
  * Added `console admin` and `console account` startup options for selecting the available console.
  * Disabled management endpoints return 404 while identity, authentication, status, and console workflows remain available.
* **Documentation**
  * Updated deployment, configuration, worker, testing, and command-line documentation for console replicas and their behavior.
* **Tests**
  * Added coverage for console routing, role configuration, controller availability, and CLI selection validation.


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
Closing without merge: superseded by plan 101 (internal .agents/plans/101-service-family.md). In stage D2 the consoles leave server-core and become their own services, so this role's identity (a console pod that cannot answer /users) becomes structural, and the managementApi flag + controller classification exist only to serve a flag no role asks for anymore. The branch stays as the ready-made shape for the one real trigger: a stated deployment requirement for management-API-less public replicas (an internet/internal split; note the management API is browser-facing, so that is a VPN-shaped topology). Reopening is a rebase, not a rewrite. Do not delete the branch.

PR #3509, opened 2026-08-26
feat(server-core,authup)!: one authup.yml replaces the conf file family

tada5hi:
Plan 101 stage C-2. The `authup.conf` family is retired in favour of one `authup.yml`, and the CLI gains `config validate` / `config schema`.

Follows #3508, which made the config schema one registry. This PR gives every entry a place in the document.

## The document

```yaml
# yaml-language-server: $schema=https://authup.org/schema/config.json
publicUrl: https://idp.example.com
env: production

db: { type: postgres, host: 127.0.0.1 }
redis: redis://127.0.0.1
smtp: { host: smtp.example.com }
trustedOrigins: [https://app.example.com]

theme:
  directoryPath: /etc/authup/theme

server:
  core:
    port: 3001
    host: 0.0.0.0
  adminConsole:   { enabled: true }
  accountConsole: { enabled: true }
```

Where a key sits is one registry field, `path`: the absolute dotted location in the document. An entry without one resolves through the reading pass's prefix, which for server-core is `server.core`, so only the 14 keys that live outside its own section spell a path out. A section is per console, never per implementation package. **No environment variable name changed**, so env-driven deployments (docker, helm, `.env`) feel nothing.

The plan asked for a `section` field. A bare section cannot express the mapping the spec requires: `adminConsoleEnabled` reads `server.adminConsole.enabled`, so the member NAME differs from the config key, not just the prefix. One `path` field covers both, and `composeSchemas` resolves paths at merge time, which is why the override is absolute rather than section-relative.

## Mechanism

`@authup/server-config-kit` gains three passes and keeps its dependency set (`envix`, `validup`, `@validup/zod`, `zod`, pinned by its own spec):

- `readSchemaFromFileTree` reads a parsed document into a partial config, verbatim (coercion and validation stay downstream), through own properties only.
- `buildSchemaJSONSchema` emits the document SHAPE rather than a flat key list, so a `$schema` line validates the real nested file, and refuses to overwrite a location.
- `composeSchemas` merges several registries into one whose entries carry resolved absolute paths, refusing a key two registries declare with a disagreeing path, environment variable, default or reader. One caller today; D2 gives it three.

## Commands

`authup config validate` reads the file and the environment, normalizes, and reports what does not hold: one line per issue, exit 1. `authup config schema` prints the JSON Schema document, in process from the same builder the build artifact comes from.

Both are `defineCLIConfigCommand` in server-core rather than in the CLI package: command bodies stay with the service, as every command has since #3507, and this one reads server-core's registry.

## The migration is made visible

The review pass found four ways an upgrading operator would have been told nothing. Each is fixed and pinned:

- `--configFile authup.conf` still LOADED, and every moved key was dropped in silence, so the service came up on a derived issuer and an empty sqlite database while the rest of the file applied. A named `.conf` is refused now; a stray one in the discovery directory keeps its warning.
- The retirement check was keyed on the `.conf` extension, but the documentation page's leading tab was `authup.server.core.ts` and the per-component family is retired in every format.
- The documented Docker mount `/usr/src/app/authup.yml` was never read, and had not been since before this plan: `entrypoint.sh` cds into `apps/server-core` (load-bearing, typeorm resolves the nested better-sqlite3 install through the cwd) and discovery globs the cwd. The CLI is given `--configDirectory` explicitly.
- `config validate` could not see a key left at its old location (the read skips what no entry claims, so a document written for a newer version still boots), reported a mistyped `--configDirectory` as valid, and printed a parse failure without its reason.

## Also

- `readEnvBoolOrString` returned `''` for a blank value where every other reader returns undefined, so `REDIS=` wrote an empty connection string instead of leaving the default. Latent (every consumer gates on truthiness) and preserved through #3508 for behaviour neutrality.
- `normalizeConfig` warns loudly when a production boot DERIVED `publicUrl` instead of being told it.
- The `$schema` URL is real: the emit script writes `docs/src/public/schema/config.json` next to `dist/config-schema.json`, a spec fails when it goes stale, and vitepress serves it from the site root.
- The ignore rules cover every discovered extension, since the documented first tab is `authup.ts` and such a file carries the admin password and the database credentials.

## Breaking

`authup.conf` and `authup.server.core.conf` are no longer read. `upgrading.md` carries the full key-move table.

## Verification

server-core 214 files / 2498 tests green (sqlite), server-config-kit 31, authup unit 2, `check:types` green for both apps, `test:smoke` proving the yml round trip and env-wins precedence, the docs build serving `/schema/config.json`, and the built CLI exercised by hand across valid, zod-invalid, cross-key-invariant, unparsable, half-migrated, mistyped-directory, named-`.conf` and stray-`.conf` inputs.

## Summary by CodeRabbit

* **New Features**
  * Added support for a unified `authup.yml` configuration file with nested settings.
  * Added `authup config validate` for checking configuration files and unknown options.
  * Added `authup config schema` to display the available JSON Schema.
  * Added support for YAML, JSON, JavaScript, and TypeScript configuration files.

* **Bug Fixes**
  * Empty environment variables now preserve configured defaults.
  * Production startup warns when `publicUrl` is not explicitly configured.

* **Documentation**
  * Updated deployment, Docker, theming, migration, and configuration guides for the new format.


tada5hi:
On the MD014 comment (shell prompts in the `authup config validate` / `schema` examples): leaving as is. `markdownlint` is not configured or run in this repo, and the `$ ` prefix is the established convention in these pages, which carry ten other lines in that shape (`bare-metal.md` 9, `docker.md` 1). `apps/authup/README.md` used it before this branch too. Dropping the prefix on two lines would make them inconsistent with their siblings for a rule the project does not enforce.

PR #3510, opened 2026-08-26
feat(server-core,core-http-kit): authorize info endpoint

tada5hi:
Plan 101 stage D2, PR 1. Additive: it adds the endpoint stage D2 is built on and changes nothing about how the `/authorize` page behaves.

## Why

The hosted `/authorize` page assembles its render input inside `AuthorizeController.serve()` and hands it straight to the in-process SSR renderer. Stage D2 moves that renderer out into `@authup/server-auth-console`, which needs the same input over HTTP. This PR exposes it and nothing else.

## What

`GET /authorize/info` answers that input as JSON.

- **One producer.** The assembly moved into `buildAuthorizeInfo(event)`, and `serve()` is now nothing but the render around it, so the page and the endpoint cannot report different things. A spec asserts the two answer equal objects for the same query, which is the guard that keeps them from drifting once a second renderer exists.
- **Anonymous and `Cache-Control: no-store`**, like the page GET. Everything it discloses, the page already discloses to the same anonymous visitor, through the same trimmed client DTO.
- **It takes the page's own query**, the `provider` hint and the closed `error` marker set included, because the answer is derived from all of it.
- **A refused request is a 200 carrying the refusal** under `error`, not an error status. A refusal is what the page renders, so a caller has to be able to render one too.

## Two things worth reviewing closely

**`requestPath` is anchored on `/authorize`, not on the route that was called.** The page renders it as the `redirect` parameter on its register and password links, so it has to lead back to the page; leaving it derived from the request path would make `/authorize/info` name itself there. For the page this changes nothing for any request whose path is already `/authorize`, and normalizes the pathological ones (a case- or slash-variant that routup still matches). Pinned by *should point requestPath at the page, not at itself*.

**The shape is deliberately the complete `payload.data`,** `federatedLogin` included, rather than the field list the plan sketched. That is the point of the endpoint: D2-2 must not need a contract bump. I checked the page reads its entire render input from the payload (its one `window.location` use is a browser-only history cleanup on mount), so `AuthorizeInfo` plus `config.baseURL`, which is the future service's own api url, is the whole of it.

`ClientSummary` and `RealmSummary` were local `Pick` aliases duplicated across three files; they move into `@authup/core-http-kit` next to the new `AuthorizeInfo` and `client.authorize.getInfo()`. The client method also accepts a raw search string, so a renderer can forward a browser query untouched instead of round-tripping it through a record.

## Verification

- Full server-core suite green: **215 files / 2508 tests** (baseline 214 / 2498, so the 10 new ones and no regression). The existing `ui-pages.spec.ts` authorize assertions are the additive-only gate and are untouched.
- `npm run build` green across the monorepo; `check:types` green for server-core.
- trapi picked the route up on its own: `/authorize/info` with a `$ref` to a generated `AuthorizeInfo` schema.
- Reviewed adversarially across four lenses with per-finding refutation; nothing survived. Two low findings were taken anyway: a wrong prerequisite claim in the suite docblock, and the raw-string form of `getInfo`.

`docs/` is deliberately untouched. `/authorize/info` is a renderer contract, not an RP endpoint, and documenting it in the OAuth2 guide would invite relying parties to build their own login UI against it.

## Summary by CodeRabbit

* **New Features**
  * Added an authorization information endpoint that returns authorization-page data as JSON.
  * Added client API support for retrieving authorization details using query strings or parameters.
  * Included authorization status, client and realm details, redirect verification, errors, and federated login information.
  * Authorization responses are not cached, and rendered authorization pages now use the same data source.

* **Bug Fixes**
  * Improved request-path handling and consistency between authorization JSON responses and rendered pages.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -771,7 +771,15 @@ into it):
   parameter leading back to the page.
   The trimmed anonymous DTOs (`ClientSummary`, `RealmSummary`) moved from
   local `Pick` aliases in the controller into `@authup/core-http-kit`, since
-  a second consumer now types against them.
+  a second consumer now types against them. `error` is `AuthorizeInfoError`
+  and deliberately NOT `Error`: the value is `sanitizeError`'s own
+  attributes plus `message`, so an `instanceof Error` check against it is
+  always false and `name` / `stack` are absent. `code` is the discriminator
+  a renderer branches on, which is why it is declared rather than left to
+  the index signature. Note the client side still declares this one `Error`
+  (`authorize.vue`, the kit's `Authorize.vue`), as it does `client: Client`
+  against the five-field summary: both are casts over an untyped payload,
+  so nothing type-checks across the boundary and neither was swept here.
 - **Template splicing goes through `replaceTemplateMarker`** (never a bare
   `String.prototype.replace` with a string replacement). A string
   replacement expands `$&`, `` $` ``, `$'` and `$$` **in the replacement
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -752,6 +752,26 @@ into it):
   console shares (Nuxt's `__NUXT__` model; each console is its own
   document, so the shapes never coexist — the account console injects its
   runtime config under the same name).
+- **`GET /authorize/info` answers that page's render input as JSON**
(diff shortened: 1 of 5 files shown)


tada5hi:
@coderabbitai pause

PR #3511, opened 2026-08-27
feat(server-auth-console,server-core)!: the auth pages render in their own service

tada5hi:
Plan 101 stage D2, PR 2. The hosted auth pages stop rendering inside server-core and move into a service of their own.

## Why

server-core renders the login and consent pages itself, which is why it carries the console bundle, a vite dev server and a loopback HTTP client that calls the server's own listen address. Stage D2 separates the protocol from the render: server-core keeps the endpoints, the issuer and every POST that mints or ends something; a service renders the pages against the public API like any other client.

## What

**`@authup/server-console-kit`** (new, Apache-2.0). The page-serving mechanism every console service shares: cookie-derived shell stamping, security headers, asset rebasing and the template splice. It exists as a package rather than a copy per service because `replaceTemplateMarker` is what defends the `$'`-expansion trap that broke `/authorize` once already, and a helper like that must exist exactly once. The helpers move verbatim, so the diff reads as a relocation and the existing spec travels with it.

**`@authup/server-auth-console`** (new, AGPL). A routup service rendering the same bundle through the same contract, hydrating **anonymously** over HTTP: `/authorize` from the context endpoint D2-1 added, the four workflow pages from the public feature flags plus their own query, and `/logout` from nothing at all. No credential, no loopback, no database.

**The six page GETs become 302s**, carrying the request's own parameters. The hop is stateless on purpose: the request is the application's own public URL, so a cookie handle would clobber across tabs and a TTL would break the mail-borne `redirect=` chain that arrives days later.

**The CLI composes the monolith.** `createApplication` grows a generic `mounts` option and mounts what it is handed under a path it is told, learning nothing about consoles; `apps/authup` hands it the auth console handler. A split deployment mounts nothing and runs the services behind their own listeners.

## Three things worth reviewing closely

**The plan's `/logout` shape was wrong, and this corrects it.** `GET` and `POST /logout` are the *same handler* today: the POST is the OIDC `form_post` binding, so **both are browser navigations**. Moving the revoke into "the existing POST" would have silently dropped a spec binding. Both forward now, and the JSON call the rendered page makes is discriminated by content type, which a navigation can never satisfy.

**The end-session flags must answer the page's own request.** `serverRevoked`, `hintSub` and `hintSubKind` decide whether a browser may tear its own session down unattended. Carried in a redirect URL they would be attacker-suppliable, and the forced-logout CSRF gate inside `AEndSessionForm` would be deciding on operands its attacker chose. `redirect` still rides regardless of the revoke, because the page navigates it only *after* its own sign-out has run; what `serverRevoked` gated was the server's automatic bounce, so a hint-less round-trip could not make an RP believe a logout it never got.

**`baseURL` and `basePath` came apart.** They were one value only while server-core served the pages on its own origin path. `baseURL` stays the API (the console derives its http client and cookie path from it); `basePath` is where the service is served, which the router and every inter-page href need. `payload.config` carries an index signature, so this is additive and moves no contract version, and the app falls back to the old derivation for a host that predates the split.

## Consequences to know

- **The revoke is now JavaScript-dependent.** A browser with JS disabled no longer revokes on `/logout` — it also never returns to the RP, so both sides stay consistent rather than the RP believing a logout it did not get.
- **A verified hint plus a valid redirect renders the page briefly** instead of bouncing instantly.
- **The auth service renders unthemed for one release window.** Theme application follows the consoles in D2-3, which is when both sides of it have a home.

## Verification

- Full monorepo build and test suite green: **server-core 211 files / 2447 tests**, service 29, kit 12, auth console 10, CLI 2.
- **The end-session matrix is re-pinned against the two-step shape** (14 tests), including both forwarding cases. The plan's named fallback was not needed.
- **The smoke runner proves the composition end to end**: `authup start` boots one process, `/logout` hands over to `/console/auth/logout`, the service renders the shell and its asset is served. Both halves are asserted separately, because a forward that lands nowhere answers the API-side probe exactly like a working one.
- New SSR-cleanliness guard: every page renders server-side without a DOM. The app is also shipped as a static shell where no SSR service runs, so that path would keep working while the SSR path silently stopped.

## Not in this PR

server-core still holds `adapters/http/ui/**` and the console dependencies; shedding those, the static console services and the theme extraction are D2-3. The `ui-pages*` coverage that asserted server-core rendering is gone or retargeted here — the `$`-expansion guard it exercised end to end is covered directly in the kit's own suite.

## Summary by CodeRabbit

* **New Features**
  * Added a standalone Auth Console service with server-side rendering, health checks, asset serving, and supported authentication workflow pages.
  * Authentication workflows now redirect to the configured Auth Console while preserving relevant request details.
  * Added configurable Auth Console URL support, including environment and configuration-file options.
  * Logout now supports JSON end-session responses and browser redirects.
  * Added shared console serving capabilities for consistent preferences, asset paths, security headers, and hydration.

* **Bug Fixes**
  * Improved console routing when deployed under a subpath and strengthened redirect validation.
  * Added coverage for server-side rendering and console redirects.


tada5hi:
@coderabbitai pause

PR #3515, opened 2026-08-31
feat!: configuration is a property of the document, and a console is a service

tada5hi:
Configuration stops being something server-core owns and hands out, and becomes a property of the document that every service reads for itself. A console is then a service in its own right: server-core neither imports nor mounts one.

## Normalization is declared on the key

An entry gains `resolve`, and `resolveSchemaData` is the pass that runs it:

```ts
publicUrl: resolve: ({ value, get }) => value ?? derive(get('server.core.host'), get('server.core.port'), get('env'))
url:       resolve: ({ value, get }) => value || `${get('publicUrl')}${SEGMENT}`   // and asserts its own origin
```

All eleven rules that lived in `normalizeConfig` moved onto the keys they constrain — the issuer, the canonicalized and dev-seeded trusted origins, each console url and the refusal of a foreign one, every path made absolute against `rootPath`, the trustProxy shapes, the secret byte length, the three flag implications. `normalizeConfig` is now the three passes any service runs, and folded into `read.ts`.

References are by absolute **document path**, never by entry object: `defineSchema` clones an entry when it stamps a prefix, and a service composes its selection by spreading, which flattens sections. `server.core.port` addresses the same key whether its reader spread that section flat or nested it. Resolution is lazy and memoized, so declaration order never matters; a cycle throws naming the chain.

`alt` is deleted along with the three recursions it needed (`resolveSchemaEntryPaths`, the env chain, the multi-path file read). It was a *source-address* chain — borrow another key's path or variable — so it could never borrow a computed value. Its one use, `host` inheritance, is a two-line `resolve`.

## A console is a service

Each console is an orkos `Application` with `ConfigModule`, `ThemeModule` and `HTTPModule` behind `InjectionKey`. `authup console` and the `authup-<name>-console` binary start the same thing, so the two paths cannot diverge, and a pre-registered token wins over the module that would register it — the test seam server-core has and the consoles did not.

`readConsoleConfigs` lost its `core: Config` parameter. A console standing alone with nothing configured now resolves:

```
apiUrl http://localhost:3000 · url .../console/account · trustedOrigins ["http://localhost:3010"]
```

`authup start` composes those Applications rather than reaching past them. It sets each console up with `listen: false` and mounts the app that graph built, so the composed topology runs a console's own config and theme modules like the two standalone paths do. Asking each console for a bare handler would have made `authup start` a third way to start a console that only resembled the supported ones. Teardown runs the consoles first: they are mounted ON server-core's listener, so tearing that down under them would leave their modules unwinding against a socket that is gone.

`ApplicationMount` is gone. `HTTPModule` takes a `mount` callback invoked in the one window a sub-application belongs in — after the controllers, so a console's wildcard shell route cannot shadow `/console/<name>/login/start`, and before the error middleware, so it inherits the error handling. `authup core` is its own command again, mounting nothing, so the split topology exists.

## Defects fixed on the way

Several were only reachable through the smoke runner; every unit suite passed against a CLI that could not boot.

- `authup start` crashed with `No registration found for: TypedToken(IApp)` — `createApplication` declared a `container` and an `http` module and used neither.
- Consoles were mounted at their absolute public url rather than its path component, after the listener had already started.
- `authup core` was an alias of `start` and served every console, so the split topology could not exist.
- Console services were handed a raw document instead of resolved values: a bare-host `trustedOrigins` entry left the account console matching `ref` against a scheme-less pattern that matches nothing, and an unset `publicUrl` refused to start a console that used to derive one.
- `mountSchema` built a child container per section and never attached it, so validup stripped every section from the parsed configuration.
- `healthcheck` layered its passes defaults-last, so `PORT` never reached the probe.
- Both static consoles passed `THEME:` to a `serve()` that reads `theme`, so the operator theme never applied; the account console read `values.distPath`, which its namespace does not carry, so `ACCOUNT_CONSOLE_PATH` was inert.
- The `mtlsPublicUrl` invariant was dropped and restored; all five throwing invariants are now asserted to fire from the schema alone.

## Naming

A package does not repeat its own name in its exports. `@authup/server-account-console` exports `createApplication`, `createHandler`, `createServer`, `resolveConfig`, `readConfigFromEnv` and `CONFIG_SCHEMA`. Identical names across siblings are the intended outcome: `server-adapter-node` and `server-adapter-socket-io` have both exported `createMiddleware` since long before the consoles existed, and the two only ever meet in a consumer importing several, where an alias at the import site says which is which. `apps/authup` is that consumer, and its type imports already aliased `Config` / `ConfigInput`, so those files are now uniform rather than half-and-half. The client bundles keep their qualified names (module-internal, not package exports), and `@authup/server-config` keeps `ACCOUNT_CONSOLE_BASE_PATH`, since it declares all three sections and the console name is what tells them apart.

The `Config`/`CONFIG_` prefix leaves `server-config-kit` and `server-config`, and `Console` leaves `server-console-kit` — the package names already say it, and server-core does not call its own `CoreHTTPModule`. `defineStaticConsole` keeps its word: that is the *kind* of console, not the package name.

## Breaking

- **Default port is 3000.** This is what the container always used internally (`3001:3000` in every compose file); only the host-side number was 3001. Docker deployments need nothing.
- **A console started through its own binary on a foreign domain refuses to start.** It used to boot and half-work. Already refused through the CLI; a console under a path of its own is unaffected.
- Renamed exports across `server-config`, `server-config-kit`, `server-console-kit` and the three `server-*-console` services.

Both operator-facing breaks are written up in `docs/src/guide/deployment/upgrading.md`.

## Verification

`nx run-many -t test` green across **27 projects** · lint 0 errors · smoke green on both the composed and split topologies · **packed smoke green**, which is the gate that reproduces install-layout breakage a workspace run passes straight through.

`readConsoleConfigs` gained the tests it never had, each verified by deleting the line it defends.

## Notes for review

- Two WIP commits (`fix: save intermediate changes`, `fix: commit intermedaite changes`) are early in the history and can be squashed on merge.
- `apps/server-core/test/unit/modules/provisioning.spec.ts` flaked once under a loaded parallel run with `SqliteError: FOREIGN KEY constraint failed`. This branch has an empty diff against provisioning, and it passes in isolation and on re-runs, so it is pre-existing — flagging rather than hiding it.


tada5hi:
## Audit follow-up

Two defects found by auditing the branch, both fixed and pushed.

**A console refused to boot on server-core's constraints.** The console registries selected all of `CORE_SCHEMA`, and `resolveSchemaData` runs every resolver in a registry — so a static file server with no database, key store, MFA or event log refused to start on a short `SECRETS_ENCRYPTION_KEY`, on `mfaRequired` without `mfaEnabled`, and on the throttle without the event log. A boot failure in exactly the topologies this branch enables. Nothing in any console reads `values.core`; the section was there for the two listener keys `publicUrl` derives from, so that is what each console selects now.

**`ThemeModule`'s provider reached nothing.** None of the three consoles forwarded the second `createHandler` argument, so every handler fell back to building its own provider from the configuration — which renders a perfectly good un-themed page, which is why it looked like it worked. `createApplication` also took a config argument beside an options bag; it takes one `CreateApplicationContext` now, matching server-core's shape, and folding them is what surfaced the missing forward.

Both regression tests are mutation-verified: each fails when its fix is reverted, and nothing else does.

## Known, not fixed here

- `resolveSchemaData` shallow-copies, so it writes through into the nested sections of the data it is given. Harmless at every current call site (all pass freshly merged data) but it contradicts the fresh-per-call discipline `buildSchemaDefaults` keeps. Belongs in the planned `resolve` refactor.
- A composed production start emits four identical `publicUrl` warnings (one per service) where it emitted one, now that each service derives independently.

PR #3522, opened 2026-09-01
feat(authup,server-console-kit,server-admin-console,server-account-console,server-auth-console,server-core)!: an experimental dev command serves the consoles from source

tada5hi:
`authup dev` is exactly `authup start`, except that every console whose package resolves to a **source checkout** is served through a vite dev server with hot module replacement instead of from its built `dist/`. In this workspace that is all three consoles; for a published operator with no source checkout it is none, and the command says so and opens no HMR port.

Detection needs no configuration key: a published console ships `files: ["dist"]` and so carries no `vite.config.ts`, while a workspace symlink and an integrator's substituted checkout both do. The presence of that file is the fact, not a declaration of it.

This restores the development loop that plan 101 D2 removed with server-core's console serving, and extends it to the two static consoles, which never had it.

## Why it is worth having

Editing `packages/client-web-kit/src` now hot-updates the admin console **and** the server-rendered auth pages at once, because each console's vite config already aliases the kit to source.

It also fixes a fidelity problem. `npm run dev -w apps/client-admin-console` serves the console cross-origin on `:3010`, which forces the *standalone* path: browser PKCE, `cookieSession: false`. Because `authup dev` puts everything on one origin, the dev loop exercises the **cookie-session** path a served console actually uses in production.

## How it avoids duplicating each console

Three optional seams, all defaulting to today's behaviour, so an operator who never runs `dev` observes nothing:

- `@authup/server-console-kit`: `StaticConsoleDefinition.readShell?` overrides where the shell HTML comes from. Everything after the read is unchanged, which is what keeps the operator theme, the account console's `ref` validation and the console security headers working in dev.
- `@authup/server-{admin,account}-console`: `createHandler(config, theme?, readShell?)` forwards it.
- `@authup/server-auth-console`: `createHandler(config, theme?, render?)` substitutes the page render, so its six routes and their payload assembly stay owned by the service.

Vite runs `appType: 'custom'` and is mounted **in front of** each console's own handler, never instead of it. With `spa` it would answer the shell itself and silently drop the theme, the `ref` check and the security headers.

## Breaking change

`@authup/server-core` gains an `exports` map, which narrows what was previously reachable through `main`: `require()` now resolves via a `default` condition, and subpaths are no longer exported. Nothing in this repo used either.

The map carries an `authup-source` condition pointing at `src/index.ts`, which is how the workspace runs server-core from TypeScript under `ts-node/esm`. It is deliberately **not** named `development`: that is a public condition Vite and Vitest set by default, and the tarball ships no `src/`, so a published consumer resolving under it would have hit `ERR_MODULE_NOT_FOUND`.

## Security

An adversarial audit of the branch found two defects in the dev path, both confirmed against a running server and both fixed here:

- The vite dev servers had no `server.fs` restrictions, and in `middlewareMode` the listener is server-core's, which binds `0.0.0.0` in the shipped container. `@fs` served the whole workspace, including `apps/server-core`'s sqlite database and the plaintext signing key inside it. Now denied, with the default deny list restated (vite **replaces** rather than extends it) and extended.
- Vite registers `/__open-in-editor` unconditionally. It spawns a local process and is reachable cross-origin from any page a developer visits. It is now refused on every console mount, case-insensitively, because vite's connect lowercases before matching.

`authup dev` additionally refuses to start under a production environment. The image does `COPY . .` and `npm ci` before `ENV NODE_ENV=production` with no prune, so a production container started with `dev` would otherwise have found the source and the devDependencies present.

## Verified

Booted end to end: three consoles served from source, real HMR (captured `js-update` frames), console security headers intact, operator theme applied, a complete cookie-session login, server-core provably running from TypeScript with `dist` untouched, and a packed install falling back to built bundles. Repo-wide: 2285 tests, 0 lint errors, `test:smoke:packed` green.

## Known issues, not addressed here

- **Pre-existing:** `apps/server-core` declares `better-sqlite3@^13.0.3` while typeorm peers it at `^12.0.0`, so npm nests it where typeorm's resolution cannot reach it. This breaks a built `authup start` from most working directories. `apps/authup` declares the driver as a devDependency so the dev command boots, but the root cause deserves its own fix.
- A vite transform error answers `200 text/html`, and ts-node type errors print `[Object: null prototype]`. Both are upstream behaviours; they make dev failures harder to read.
- `dev.ts` reaches past `createApplication` for a bare console handler, which `start.ts`'s own comment argues against. No behavioural loss today.

`authup dev` is marked EXPERIMENTAL on the console-theming precedent. The parts most likely to be reshaped are the source-checkout detection rule and the coupling to whichever vite major the caller has installed.

## Summary by CodeRabbit

* **New Features**
  * Added the experimental `authup dev` command.
  * Run server-core and console applications from source with hot module replacement on a single origin.
  * Console replacements can now be served directly from source checkouts instead of built files.
  * Added safeguards preventing development mode in production and blocking sensitive files and editor endpoints.
* **Documentation**
  * Updated setup, deployment, console replacement, and development guides with the new workflow and behavior.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai pause

PR #3529, opened 2026-09-02
feat(client-web-nuxt): allow namespacing the session cookies

### LINKED ISSUE #3527: client-web-nuxt: allow namespacing the session cookies (cookiePrefix) so a widened cookieDomain cannot collide with another authup client
### Summary

`@authup/client-web-nuxt` (and the `client-web-kit` install beneath it) writes its session cookies under fixed names — `access_token`, `refresh_token`, `id_token`, `realm`, … (`core-http-kit` `CookieName`). The only scoping knob is `cookieDomain`.

That combination means an embedding app cannot widen `cookieDomain` without risking a collision: any *other* authup client reachable under that Domain writes the same names, so the browser holds two records under one `(name, domain, path)` key. A read takes the first — which can be the older one — so each side can end up driving on, refreshing and cleanup-revoking the other's tokens. It surfaces as being signed out on the next page load.

### Where we hit it

FLAME Hub ([PrivateAIM/hub](https://github.com/PrivateAIM/hub)) embeds `client-web-nuxt`. Layout:

```
app.example.com        <- Hub UI, embeds client-web-nuxt
auth.example.com       <- authup
storage.example.com    <- another first-party service
```

The UI sets `cookieDomain: example.com` so its session cookie also reaches `storage.example.com` (a file download is a top-level navigation, so it cannot carry an `Authorization` header and the service falls back to the cookie). That same Domain also delivers the cookie into authup's origin, where the console wrote the same names — and we got the sign-out loop. Observed, not theoretical.

### Already improved by #3498 — thank you

`d84b730b` (`feat(server-core)!: authenticate the account console with an opaque session cookie`) plus `8c7cf388` move the server-served consoles onto `authup_session` (`httpOnly`, `SameSite=Strict`, path-scoped to the deployment base path). For the default deployment mode that removes the console-vs-embedding-app collision entirely.

### What still seems open

1. Per `apps/client-admin-console/README.md`, a console hosted **standalone on a foreign origin** still signs in through the browser-side authorization-code flow "instead of the session cookie" — so it still writes the fixed-name cookies and can still collide with an embedding app under a shared Domain.
2. Two different embedding apps on sibling hosts under one Domain still collide with each other, with no way to tell them apart.

### Proposal

A `cookiePrefix` (or explicit `cookieNames`) option on `client-web-nuxt`, forwarded into the `client-web-kit` install, so an embedding app can namespace its cookies:

```ts
['@authup/client-web-nuxt', {
    apiURLRuntimeKey: 'authupUrl',
    cookieDomainRuntimeKey: 'cookieDomain',
    cookiePrefix: 'flame_',
}]
```

Every write already funnels through a single `cookieSet(key, value, options)` in `packages/client-web-nuxt/src/runtime/plugins/kit.ts`, and `cookieGet`/`cookieUnset` alongside it, so this looks like it could be a narrow change at that seam rather than anything structural.

Happy to open a PR if you'd like it shaped a particular way.

### Also

Which release is #3498 targeted at? We'd like to know whether to wait for it or namespace on our side first.

tada5hi:
Closes #3527.

## What

`cookiePrefix` on `@authup/client-web-nuxt`, prepended verbatim to every store
cookie name, so an embedding app that widens `cookieDomain` no longer writes
the same names as another authup client reachable under that domain.

```typescript
modules: [
    [
        '@authup/client-web-nuxt',
        {
            apiURLRuntimeKey: 'authupUrl',
            cookieDomainRuntimeKey: 'cookieDomain',
            cookiePrefix: 'flame_'
        }
    ]
]
```

`flame_access_token`, `flame_refresh_token`, and so on. Unset keeps today's
bare names, so nothing changes for an existing consumer.

## Why it lands in the plugin and not in the kit

The reporter's read of the seam is exactly right.
`packages/client-web-kit/src/core/store/install.ts` is the only source file in
the repo that names `CookieName` outside the enum and the tests, and all 16
touches inside it go through the caller-supplied `cookieGet` / `cookieSet` /
`cookieUnset`, which this plugin supplies all three of. So a prefix applied
here is total, not partial. The two cleanup sweeps (`dropShadowingCookies`
and the legacy `user` sweep) inherit it for free, which is the desirable
narrowing: today, under a widened domain, the sweep deletes BARE names at
`Domain=example.com` on every ancestor path, meaning a sibling app's or the
IdP's own cookies.

Nothing server-side reads any of these names. The complete set of cookies read
server-side is `vc-locale`, `vc-color-mode`, `authup_session`,
`authup_console_login` and `authup_federated_login`, and the three server
adapters parse no cookies at all, so the rename crosses no server contract
inside authup.

Only a consumer that supplies its own `cookieSet` can set a `Domain`, since
the kit assigns one nowhere. So in every cross-host collision the widening side
IS the side that can prefix, and breaking name equality on one side is enough
to dissolve it.

This deliberately does not revive the parked kit-level `cookiePrefix` of #3403.
That one was about authup's own consoles sharing the IdP origin, where a second
cookie set roughly doubles header weight on every request including assets, and
that reasoning still stands. It says nothing about a downstream app widening
its domain across sibling hosts, which is what this is.

## What is deliberately not here

- **No automatic sweep of the pre-prefix cookies.** Tempting and wrong: under a
  widened domain it would delete a not-yet-migrated sibling's cookies, and on
  the IdP origin it would clear the hosted auth pages' own SSO cookies, signing
  the visitor out as a side effect of loading an app. Documented instead. The
  stranding is bounded: only `access_token` and `access_token_expire_date`
  carry a lifetime, the rest are session cookies.
- **No `cookiePrefixRuntimeKey` twin.** `cookieDomain` has one because the
  domain varies per deployment. A prefix is the application's identity.
- **No charset validation, no imposed separator.** #3403 joined with `.`; this
  concatenates verbatim, matching the issue's `flame_`. The value is authored
  in build config and a bad one fails on first login. The constraint is stated
  in the docs.
- **No new spec.** `buildApiUrl` and `buildCookieOptions` in the same file are
  untested for the same reason: `plugins/kit.ts` imports `#imports` at module
  scope, which the package's vitest config cannot resolve, so a spec would cost
  a stub module and an alias to cover one ternary. Verified instead by running
  the built plugin against a stubbed `useCookie`: with a prefix set, all seven
  names touched carry it, none leak bare, and a prefixed install does not read
  a seeded bare `access_token`, which is the adoption that produces the
  reported sign-out loop.

Still unsupported, and now written down in `.agents/architecture.md`: two
kit-DEFAULT consumers colliding on ONE host with no widening involved. Both
take the host-only `useCookies()` default, and a console builds its install
from `window.__AUTHUP__`, which carries no such key. That is what a kit-level
option would be for.

## Second commit

`fix(client-web-nuxt): forward every declared module option` is a one-line
prerequisite, split out. `runtimeOptions` was a hand-written list of six fields
annotated as `RuntimeOptions` itself, so it could only fall behind, and it
already had: `serverApiURL` is declared and read by the plugin, yet never
reached `runtimeConfig`. Spreading removes the class rather than adding a
seventh line to it.

## Verification

- `npm run build -w packages/client-web-nuxt`
- `npm run test -w packages/client-web-nuxt` (15 passed)
- `npm run test -w packages/client-web-kit` (247 passed, the suite that pins
  the seam this relies on)
- `npx eslint packages/client-web-nuxt/src`
- The stubbed-`useCookie` run described above, against `dist/`

## Summary by CodeRabbit

* **New Features**
  * Added an optional cookie prefix setting for Nuxt clients.
  * Session cookies now consistently use the configured prefix, helping prevent collisions when multiple clients share a cookie domain.
  * Runtime configuration now supports additional module options, including the server API URL.

* **Documentation**
  * Added configuration guidance, naming rules, migration considerations, and coordination requirements for prefixed cookies.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -3139,12 +3139,28 @@ needs next to it (`CLI_CONFIG_ARGS`, `applyCLIConfigArgs`,
 `createCLIConfigModule`, `assertNoStrayPositionals`, `describeConfigError`)
 and no config command of its own.
 
-**Unsupported:** sharing one cookie domain between a standalone-hosted
+**Partly unsupported:** sharing one cookie domain between a standalone-hosted
 console (or a downstream kit app on its own origin) and the hosted auth
-pages — both surfaces embed the kit store under identical cookie names, so a
+pages. Both surfaces embed the kit store under identical cookie names, so a
 widened cookie domain has the two apps clobbering each other's session
-cookies. The served consoles are same-origin and hold no token cookies at
-all (cookie mode), so the question does not arise for them. The same-ORIGIN variant of this collision — a host
+cookies. A downstream app on `@authup/client-web-nuxt` has a way out since
+issue #3527: `cookiePrefix` prepends a namespace to every store cookie name
+inside that plugin's own `cookieGet` / `cookieSet` / `cookieUnset`, which is
+total because `installStore` routes all cookie I/O through those three, and
+the kit therefore needs no option of its own. Only a consumer supplying its
+own `cookieSet` can set a `Domain` at all (the kit assigns one nowhere), so
+in every cross-host collision the widening side IS the side that can prefix,
+and breaking name equality on one side is enough. Two things stay
+unsupported. A prefix is a contract across the adopting app's OWN services:
+a resource server of theirs reading the access token out of the cookie (a
+file download is a top-level navigation, so it carries no `Authorization`
+header) must read the prefixed name. And two kit-DEFAULT consumers on ONE
+host still collide with no widening involved, since both take the host-only
+`useCookies()` default and a console builds its install from
+`window.__AUTHUP__`, which carries no such key; that is what a kit-level
+option would be for (PR #3403, parked). The served consoles are
+same-origin and hold no token cookies at all (cookie mode), so the question
+does not arise for them. The same-ORIGIN variant of this collision — a host
 application at `/` embedding authup under a sub-path — is defused by the
 consoles scoping their cookies to the base path (see *Account Console →
 Session cookies are scoped to the deployment base path*); the residual
--- a/docs/src/sdks/javascript/client-web-nuxt/index.md
+++ b/docs/src/sdks/javascript/client-web-nuxt/index.md
@@ -75,6 +75,14 @@ export type RuntimeOptions = {
     cookieDomainRuntimeKey?: string,
 
     /**
+     * Prefix prepended to every session cookie name
+     * (client-side & server-side)
+     *
+     * See "Namespacing the session cookies" below.
+     */
+    cookiePrefix?: string,
+
+    /**
      * Path of the home route
      * Default: /
      */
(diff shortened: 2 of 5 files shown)


tada5hi:
@coderabbitai pause

PR #3533, opened 2026-09-02
refactor(server-config)!: drop the server. prefix from every config section

tada5hi:
`authup.yml` nested four of its six sections under `server.` — `server.core`,
`server.authConsole`, `server.adminConsole`, `server.accountConsole`. This drops
the prefix, so every section is written at the document root.

```yaml
# before                          # after
server:                           core:
  core:                             port: 3000
    port: 3000                    adminConsole:
  adminConsole:                     enabled: true
    enabled: true
```

## Why

- **It never discriminated.** There is no `client.` sibling and cannot be: plan
  081 retired `client.admin-console`, and `entrypoint.sh` exits 1 on it. A
  namespace with one member is not one.
- **It was half applied.** `theme` and the seven deployment keys (`host`,
  `publicUrl`, `db`, `redis`, `smtp`, `trustedOrigins`, `rootPath`) always sat at
  the root. Only 4 of 6 sections wore the prefix.
- **The document was the odd surface out.** `SECTION_KEY` and `AuthupConfig` were
  already `core` / `adminConsole`, services already read `config.adminConsole.url`,
  and no environment variable ever carried a `SERVER_` qualifier. `types.ts` claims
  "the document and the type read alike" — now they do.

The docker entrypoint's `server/core` is a service SELECTOR (`<service> <command>`),
not a configuration path, so it is unchanged; `.agents/conventions.md` no longer
claims the two mirror each other.

## Breaking change

An `authup.yml` section moves to the document root. No environment variable
changes.

| Was | Is |
|---|---|
| `server.core.*` | `core.*` |
| `server.authConsole.*` | `authConsole.*` |
| `server.adminConsole.*` | `adminConsole.*` |
| `server.accountConsole.*` | `accountConsole.*` |

**No migration is offered.** The prefix arrived with `authup.yml` in the current
beta and never shipped in a release. Note the file read is permissive by design,
so a document still carrying `server:` has that whole subtree skipped in silence —
`authup config validate` reports it as a path nothing reads.

`authup/helm` writes this document into a ConfigMap and follows in its own PR.

## What was deliberately left alone

Every `server.core` that names the **retired** world keeps its spelling: the
`authup.conf` migration table in `upgrading.md`, the `authup.server.core.conf`
file family, and the retired `authup start server.core` selector.
`@authup/server-config-kit`'s own test fixtures keep `pathPrefix: 'server.core'` —
that package is the generic mechanism and the prefix there is an arbitrary
example.

## Verification

- `npm run build` — 30 projects, clean
- `npm run check:types --workspaces` — clean
- `apps/server-core` — 201 files, 2295 tests passing
- `apps/authup`, `packages/server-config{,-kit}`, the three console services — passing
- `npm run test:smoke -w apps/authup` — both scenarios, including
  *workspace/config: authup.yml was applied*, which is the assertion that proves
  the new `core:` section is read from the file

## Summary by CodeRabbit

- **Configuration**
  - Configuration sections now use a flattened structure: `core`, `adminConsole`, `accountConsole`, and `authConsole` are top-level keys.
  - Existing settings, environment-variable mappings, validation, and precedence behavior remain unchanged.
  - Configurations retaining the previous wrapper are no longer processed.

- **Documentation**
  - Deployment guides, examples, and configuration references now reflect the updated key paths.

- **Tests**
  - Configuration, schema, console, CLI, and smoke-test coverage was updated for the new layout.


tada5hi:
@coderabbitai pause

PR #3538, opened 2026-09-02
feat!: the worker is a flag on start, and componentsEnabled becomes core.worker.enabled

tada5hi:
## Summary

- **`authup worker` is retired; the worker is `authup start --worker`** (`core --worker` is equivalent, a worker mounts no console either way; container: `server/core start --worker`). In both cases the operator starts the app, so the flag says what the process runs. A boolean on purpose: a value flag (`--mode worker`, the first draft) has to be declared on the root command too, or citty's root parse reads the value as a stray positional and `assertNoStrayPositionals` refuses it. Pinned by *does not read --worker as a stray positional*.
- **`core.componentsEnabled` / `COMPONENTS_ENABLED` becomes the nested `core.worker.enabled` / `WORKER_ENABLED`**, and the semantics change with the name. The key reads the same in both shapes: without `--worker` the API runs the worker alongside itself while it is true (a plain `start` needs nothing set), an API replica hands the sweeps over by setting it false, and `--worker` REQUIRES it: `ComponentsModule({ required: true })` throws at boot naming the key, instead of coming up idle. The old key was one the worker role ignored.
- **Every role has a `--help` description.** `start`, `core`, `migration` and `healthcheck` had none, which is what made the two shapes hard to find.
- **server-config-kit**: `readSchemaFromFileTree` leaves a section the document does not spell absent rather than `{}`, so a nested section reads like a key. A nested section inside a section declares its own `defineSchema` with a `pathPrefix`: the outer stamping does not recurse.
- Docs (`worker.md`, `bare-metal.md`, `configuration*.md`, `console-replicas.md`, an `upgrading.md` entry), `entrypoint.sh` usage and the `.agents` files follow.

## Breaking

- `authup worker` / `server/core worker` fail with `Unknown command worker` (exit 1). Use `start --worker`.
- `COMPONENTS_ENABLED` / `core.componentsEnabled` are no longer read: an old `COMPONENTS_ENABLED=false` means every API replica runs the sweeps again (batch-safe, N-fold); a stale `componentsEnabled:` is reported by `authup config validate` as a path nothing reads.
- A worker process fed `WORKER_ENABLED=false` refuses to start.

## Verification

- Builds, `check:types` and eslint clean on server-config-kit, server-config, server-core, authup.
- Full server-core suite green after the CLI spec update (2293 + 17); kit 32/32, server-config 15/15, authup 40/40 (new regression test red-green verified).
- `npm run test:smoke`: both scenarios pass.
- Live against the built dist: `WORKER_ENABLED=false … start --worker` exits 1 naming `core.worker.enabled`; `core --worker` boots and logs `Background components started: oauth2-cleaner, event-cleaner`; plain `start` still listens and runs the worker in-process; `config validate` flags a stale `core.componentsEnabled`.

## Follow-up

- Helm: https://github.com/authup/helm/issues/28

## Summary by CodeRabbit

* **New Features**
  * Worker mode is now launched with `authup start --worker` or `authup core --worker`.
  * Worker mode runs background tasks without opening an HTTP listener.

* **Breaking Changes**
  * The standalone `worker` command has been removed.
  * Configure background workers with `core.worker.enabled` or `WORKER_ENABLED` instead of the previous setting.
  * Worker mode now refuses to start when disabled.

* **Bug Fixes**
  * Unspecified nested configuration sections now correctly preserve default values.


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai pause

PR #3551, opened 2026-09-03
fix(server-config,server-auth-console,authup): reach the API on an internal address

### LINKED ISSUE #3550: The auth console fetches PUBLIC_URL from inside the container, so a port-mapped deployment 502s the hosted auth pages
## What is wrong

`@authup/server-auth-console` resolves its API base from `publicUrl`:

```ts
// apps/server-auth-console/src/config.ts:78
apiUrl: values.publicUrl,
```

`publicUrl` is the address a **browser** reaches the deployment at. The auth console is the one
console that fetches server-side (it renders `/authorize` and the workflow pages from
`GET /authorize/info` and `GET /`), so it needs an address reachable from **inside** the container.
Under a port mapping that publishes a port other than the one the listener binds, those two
diverge and every hosted auth page fails.

## Reproduced

Against an image built from `master`, with a postgres alongside it:

```
docker run -p 3001:3000 -e PUBLIC_URL=http://localhost:3001 -e DB_TYPE=postgres ... authup/authup:latest start
```

| path | status |
|---|---|
| `/` | 200 |
| `/console/admin` | 200 |
| `/console/account` | 200 |
| `/console/auth/logout` | 200 |
| `/console/auth/authorize?client_id=admin-console` | **502** |

`/console/auth/logout` is the tell: it is the one page that drives its call from the browser, so it
renders. The server's own log names the cause:

```
error: fetch failed (GET http://localhost:3001/authorize/info?client_id=admin-console)
  cause: TypeError: fetch failed <- AggregateError: (code=ECONNREFUSED)
    at async OAuth2AuthorizeAPI.getInfo (.../packages/core-http-kit/dist/client-*.mjs:636:11)
    at async fn (.../apps/server-auth-console/dist/server-*.mjs:317:10)
```

Nothing listens on 3001 inside the container. With `PORT=3001` and `-p 3001:3001`, every page
answers 200.

## Why it is worth fixing rather than documenting

server-core already solves exactly this for its own loopback self-calls.
`createPublicToInternalURLRewriter` (`apps/server-core/src/adapters/http/internal-client/`) rewrites
any request targeting `publicUrl` onto the server's own listen address, specifically so a self-call
does not depend on the public hostname resolving from inside the deployment (and so it skips TLS
termination). The console services have no equivalent, so under `authup start`, where the consoles
are mounted onto server-core's own listener, the auth console takes the long way round to a server
in the same process.

The two static consoles are unaffected: they inject `apiUrl` into the page for the browser to call,
so `publicUrl` is the correct value there. Only the SSR auth console fetches server-side.

## Scope

Affects `authup start`, the batteries-included topology. `authup start console` in a separate
container is a different case: the console genuinely must reach the API over the network, and
`console-replicas.md` documents `PUBLIC_URL` as the API address deliberately.

Also affected, and reported here rather than fixed:
`docs/src/guide/deployment/worker.md` has the inverse inconsistency. It publishes `"3001:3000"`
while setting `PUBLIC_URL=http://localhost:3000`, so the server-side fetch resolves and the
browser-facing URL is wrong instead.

## Options

1. In `authup start`, hand the composed console services an internal `apiUrl` (the real listen
   address) rather than `publicUrl`. The CLI is the only place that knows both, which is the same
   reason it owns the mount composition.
2. Give the console services the public-to-internal rewrite server-core's internal client already
   has, so it also covers a proxied deployment whose public hostname does not resolve internally.
3. Docs only: require the published port to equal `PORT`. Done in #3549 as an interim measure, but it
   constrains a mapping Docker users reasonably expect to be free.

## Touch points

- `apps/server-auth-console/src/config.ts` (`apiUrl: values.publicUrl`)
- `apps/server-core/src/adapters/http/internal-client/` (the existing rewriter to mirror)
- `apps/authup/src/console/config.ts` and `src/console/applications.ts` (where the CLI composes them)

tada5hi:
Closes #3550.

## The bug

`@authup/server-auth-console` is the one console that renders server-side: it
builds `/authorize` from `GET /authorize/info` and the four workflow pages from
`GET /`. It made those calls against `publicUrl`, which is where a **browser**
reaches the deployment, and nothing entitles the inside of a deployment to that
address.

Published as `-p 3001:3000` with `PUBLIC_URL` naming 3001, nothing listens on
3001 inside the container:

| path | before | after |
|---|---|---|
| `/` | 200 | 200 |
| `/console/admin`, `/console/account` | 200 | 200 |
| `/console/auth/logout` | 200 | 200 |
| `/console/auth/authorize` | **502** | **200** |
| `/console/auth/{register,activate,password-forgot,password-reset}` | **502** | **200** |

`/logout` is the tell: it drives its end-session call from the browser, so it is
the one auth page that renders without reaching the API.

## The fix

The API address is two values, not one.

**`internalUrl` (env `INTERNAL_URL`)** joins the root of `authup.yml` next to
`publicUrl` and resolves to it when unset, so a deployment reachable at one
address from both sides configures nothing. It is where a service *inside* the
deployment reaches the API, and it is what the auth console's
`createAPIClient` dispatches against.

**`publicUrl` keeps every job it had** — the issuer, the cookie scope, and the
`baseURL` the rendered page hands the browser — so nothing derived from the
internal address can reach a caller.

This is the shape [@tada5hi asked for][k8s] and the one PrivateAIM/hub already
uses (a server-only runtime-config key beside its `public` mirror): on a
cluster network the render reaches `http://authup-api:3000` directly instead of
leaving for the ingress and coming back.

```yaml
authup-console:
    image: authup/authup:latest
    environment:
        - PUBLIC_URL=https://auth.example.com
        - INTERNAL_URL=http://authup-api:3000
    command: start console
```

[k8s]: https://github.com/PrivateAIM/hub/blob/master/apps/client-ui/nuxt.config.ts

**The composed roles need no configuration at all.** `authup start` and
`authup dev` override the resolved value with this process's own listen
address, since there the API *is* that process, and the CLI is the only place
that knows both — the same reason it owns the mount composition. An explicitly
configured `internalUrl` still wins there (a composed process may be told to
take an egress route); because the key resolves to `publicUrl` when unset, that
comparison is exactly the "was one configured" question.

Two details worth stating, both mirroring `createPublicToInternalURLRewriter`:
the internal address carries **no path** (server-core mounts every route
root-relative, so a self-call must not re-add the prefix the proxy strips), and
the swap is on `baseURL` rather than at the transport layer, because nothing
derived from it is user-facing — there is no `redirect_uri` here to be compared
byte for byte at redemption.

## Not in scope, deliberately

`server-core`'s own internal client is **not** repointed at the key: it
resolves its actual listen address per request, which is strictly better than
any configured value. The two static consoles never read it either — they hand
`apiUrl` to the browser and make no call of their own.

## Also fixed here

A consequence of the change rather than a pre-existing bug: the console
services register no error middleware, so routup answers with the error's own
message, and hapic puts the resolved request URL in it. On a split deployment
that URL now names a service on the operator's network, and the auth pages are
the public login surface — an API that was merely down would have published
its internal address to every anonymous visitor:

```
{"status":500,"message":"fetch failed (GET http://internal-api.cluster.local:3000/authorize/info)"}
```

It is `sanitizeError` / `describeError`'s rule and remedy: the visitor gets
`The identity provider could not be reached.`, the detail goes to the log.
Harmless before this change, since the URL was one the browser already had.

## Verification

Everything below was run, not reasoned about.

- **The reported bug, against the built CLI.** `PORT=4401` with `PUBLIC_URL`
  naming a dead 4499: every route including `/console/auth/authorize` answers
  200, and the rendered payload still carries `"baseURL":"http://127.0.0.1:4499"`.
- **The Kubernetes case.** `start core` + `start console` as separate
  processes: `/authorize` answers **500 without** `INTERNAL_URL` and **200
  with** it, while the payload keeps the public address either way.
- **The leak.** Reproduced (`internal-api.cluster.local` in the response body),
  then confirmed gone, with the detail in the server log instead.
- **Red/green on both new behaviours.** Reverting the `payload.ts` change makes
  the two new `handler.spec.ts` cases fail (`500 !== 200`, and the address back
  in the body) rather than passing either way.
- `npm run build`, `npm run test` (2343 tests), `npm run check:types
  --workspaces`, `npm run lint`, `npm run test:smoke -w apps/authup` — all
  green.

Reviewed by a multi-agent pass with adversarial verification; five confirmed
findings are fixed in the commit (IPv6-literal listen host producing an
unparseable URL, the leak above, and three documentation inaccuracies).

## Tests

- `apps/authup/test/unit/console-api-url.spec.ts` — wildcard and single-interface
  hosts, IPv6 bracketing, no path, and that an explicit `internalUrl` survives
  the composed override.
- `apps/server-auth-console/test/unit/handler.spec.ts` — the render dispatches
  against `apiInternalUrl` while the browser payload keeps `apiUrl`; the key
  defaults to `publicUrl` and takes `internalUrl` when named; a failure names
  no internal address.
- The composed smoke scenario now probes a workflow page. It does not reproduce
  #3550 and does not claim to (this run names publicUrl and the listen port the
  same); what it covers is that the SSR fetch path is exercised at all, which is
  how the bug went unnoticed there — `/logout` renders with no backend.

## Docs

`docker.md` drops the "publish the port the listener binds" constraint #3549
added as an interim measure (never released), `console-replicas.md` gains
`INTERNAL_URL` with the reason it matters to the auth console alone, and
`configuration-server-core.md` documents the key.

The helm chart will want a values key for `INTERNAL_URL` (`authup/helm`); not filed yet.

## Summary by CodeRabbit

* **New Features**
  * Added an `INTERNAL_URL` deployment setting for server-side API requests.
  * Auth console rendering now uses the internal API address while preserving the browser-facing URL.
  * Added automatic internal URL resolution for composed and development deployments.
* **Bug Fixes**
  * Improved authentication page reliability when public URLs, ingress routes, or certificates are not reachable internally.
  * Server-side API failures now return a generic error without exposing internal addresses.
* **Documentation**
  * Updated configuration, Docker, and console deployment guidance for `INTERNAL_URL`.
* **Tests**
  * Added coverage for internal URL resolution, server-side requests, error handling, and password-reset rendering.


tada5hi:
@coderabbitai pause

PR #3555, opened 2026-09-05
fix!: work the beta.64 audit backlog

tada5hi:
Works `.agents/plans/104-beta64-audit-backlog.md`, the low/informational remainder of the `v1.0.0-beta.63..master` pre-release audit. Every item was re-verified against master before it was implemented; sections A and D carry a spec that failed before the fix.

## Correctness (A)

- **A1** `backchannelLogoutUri` is capped at the column's 2000 characters, so an over-long value is a 400 instead of a driver error mapped to 500.
- **A2** `authup healthcheck` probes the configured bind address: `core.host` falls back to the root `host`, and a wildcard bind is looped back the way the composed roles reach their own listener.
- **A3** A federated login whose attribute mapping rewrites `email` clears `emailVerified`, the rule `UserService.save` already applies.
- **A4** `POST /identity-providers/:id/login-complete` rides `isSameOriginRequest`: `Sec-Fetch-Site: same-origin` is required and fails closed when absent, matching the console session credential and what `cors.ts` and the architecture doc already claimed.

## Silent failures (B)

- **B2** A configured provisioning directory that does not exist is logged at boot.
- **B3** The auth console service asserts the bundle's `CONTRACT_VERSION` at boot with an actionable message, and a dist without `dist/server/server.js` no longer resolves.
- **B4** Each console service validates its own configuration section before layering the defaults, so the boot path and `authup config validate` agree. See the breaking change below.

## Drift (C), client stability (D), operational (E)

- **C1** The auth console's render state is a per-handler closure (`createRenderPage`), no module scope; `authup dev` consumes the pure resolver.
- **C2** The auth console's hashed assets are served immutable for a year like the static consoles.
- **C3** The three `authup-<name>-console` bins start through `createApplication`, the path `authup start console` takes.
- **C4** The dead `AUTH_CONSOLE_SEGMENT` is gone.
- **C5** `setRealm` narrows to `{ id, name }` at the sink like its siblings.
- **D1** Chunk-load recovery is one-shot per target per document (a `sessionStorage` marker), so a persistently 404ing asset no longer reloads in a loop.
- **D2** `serverApiURL` lives in the private `runtimeConfig.authup`, never in the public block serialized into every page.
- **E1** Bulk session revoke fans out to 5, below the default pool of 10.
- **E2** `migration generate` is refused by the `authup` binary; only server-core's dev CLI opts into it.

## Documentation and packaging (F, G)

- Upgrading guide: stale `:3010` entry removed (F1), `server.<name>Console.path` corrected (F2), the shipped beta.64 block titled and the post-tag changes noted.
- Agent docs describe the real `packages/server-config/src/sections/{root,...}` layout (F6) and record every fix above.
- `DEPENDENCY_GRAPH.md` regenerated (F4). F3 had already landed in `08738ad49`.
- READMEs for `server-config` and the three console services (G2); `@types/node`, `cross-env` and `validup` declared where they are used (G1).
- F5: upstream issue tada5hi/javascript#940 asks the shared commitlint config to reject the bare `BREAKING:` token. The release body of v1.0.0-beta.64 still lacks the "disabled console routes answer 404" note; that is a release-page edit, not a repo change.

## Verification

`npm run build` (30 projects), `npm run check:types --workspaces --if-present`, `npm run lint` (0 errors) and `npm run test` (27 projects, 0 failures, 8 skipped), all in a clean worktree at this branch.

BREAKING CHANGE: `resolveConfig` and `readConfigFromEnv` of `@authup/server-admin-console`, `@authup/server-account-console` and `@authup/server-auth-console` return a `Promise`; an embedder calling either directly needs an `await`.


tada5hi:
@coderabbitai pause

PR #3556, opened 2026-09-05
feat(create-authup): add the install wizard package

tada5hi:
## Summary

Plan 100: `npm create authup` (the `create-authup` package) runs a short prompt flow and writes the deployment files for one of four targets into cwd. It scaffolds only; the `authup` CLI or the `authup/authup` image runs the result. Zero runtime dependencies (`node:readline/promises`, `node:util` `parseArgs`, string templates).

| Target | Files written | Printed next step |
|---|---|---|
| docker | `authup.env` | `docker run --env-file authup.env -p 3000:3000 authup/authup:<version> start` |
| compose | `docker-compose.yml`, `.env` | `docker compose up -d` (a worker split prints the one-off `migration run` first) |
| helm | `values.yaml` | `helm repo add` + `helm install ... -f values.yaml` |
| bare-metal | `package.json`, `authup.yml`, `.env` | `npm install`, `npx authup config validate`, `npm start` |

Image tag and npm range are the package's own `pkg.version`, so release-please linked versions keep the wizard in lockstep with the release it scaffolds.

## Decisions

- **Monorepo, not an own repository.** Every template value is a function of this repo's deployment surface (role commands, `WORKER_ENABLED`, `MIGRATION_ENABLED`, `INTERNAL_URL`, the `authup.yml` sections), version pinning is free through linked versions, and the drift spec below is only possible in-repo without a one-release lag. Extraction later stays cheap because there are no deps.
- **Drift spec.** Every emitted `authup.yml` path is checked with `findUnknownSchemaPaths(tree, SCHEMA)` from `@authup/server-config-kit`, every env name against the `EnvironmentVariable` enum (plus the typeorm-extension `DB_*` set and the bundled images' own variables). Helm values are the one target whose keys track another repository and are deliberately not drift-checked.
- **Bare metal carries the database as `DB_*` in `.env`, no `db` block.** `DatabaseModule.buildDataSourceOptions` hands a present `db` block to the data source as is and never merges `DB_*` over it, so the draft's password-less block would have booted without its password; `authup config validate` cannot see that since it never builds the data source.
- **Compose.** `.env` is quoted for compose-go (double quotes with `\\`, `\"`, `$$`; a single-quoted `\'` there is an escaped quote and `$` expands inside double quotes), every operator `$` in a yaml value is doubled, the bundled database carries a TCP healthcheck (`pg_isready -h localhost` / `mysqladmin ping -h 127.0.0.1`, both images run a network-less temporary server while initializing) with `service_healthy` dependencies so the one-off `migration run` never races the engine, and `WORKER_ENABLED=false` + `MIGRATION_ENABLED=false` are emitted only under a worker split (a console-only split keeps one API process that sweeps and migrates itself).
- **Helm** is written against the chart tracking beta.64 (authup/helm#29): `worker.*` and `server.splitConsoles` do not exist on the released 0.3.0 chart, whose values schema rejects them, so a split document names that requirement in its header and a no-split document renders on either. The helm target requires an origin-root public URL (the chart refuses a console split under a sub-path) and passes the answered database password to the bundled engine (left empty the chart generates one and drops the answer).
- **Prompt flow.** `start123` and blank refused, admin password bounded to the registry's 3..256, a secret carrying both quote kinds refused (no dotenv quoting holds it), secrets never trimmed; the public URL may carry no credentials, query or fragment; the database host and redis url are prefilled with the loopback on bare metal only (inside a container that address is the container itself). Piped input works: readline drops all but the first line of a chunk under `rl.question`, so the seam buffers `line` events and treats a close mid-flow as an error rather than exit 0 with nothing written.

## Verification

- `npm run build`, `npx eslint`, `check:types` and 78 tests in `packages/create-authup` (renderers, prompt flow, drift spec, the built bin over piped stdin in a temp dir).
- Every emitted document validated with its consuming tool: `docker compose config` on four projects, `helm template` against the authup/helm#29 chart (and the released 0.3.0 for a no-split document), `authup config validate` on both bare-metal shapes, `docker run --env-file` reading a password with a space verbatim.
- The emitted worker-split compose project ran for real with the cached beta.64, postgres:17 and redis:7 images: the one-off `migration run` waited for the healthy database and applied the chain, `up -d` had the API answering in 4s with both consoles at 200, the worker running the sweeps and the API reporting the handover; adversarial values (`svc$user`, `abc\`, `it's $ecret`) came back byte for byte from inside the container.
- The emitted sqlite bare-metal project booted with the workspace CLI, served the API and both consoles, wrote `db.sqlite` into the project directory and exited cleanly on SIGTERM.
- A four-lens adversarial review (runnable output, operator input, prompt flow, conventions and docs) produced 22 findings, none high; the three verified mediums (helm sub-path split, helm bundled password, compose migration race) and the remaining concrete ones are fixed here.

## Out of scope, noted for follow-up

- `packages/client-web-kit` imports `@authup/errors` in two files without declaring it in `package.json` (surfaced by the regenerated dependency graph).
- The built CLI cannot resolve `better-sqlite3` from a cwd without its own `node_modules/better-sqlite3` (documented in architecture.md; the emitted bare-metal project is fine after its own `npm install`).

Related: authup/helm#29.

## Summary by CodeRabbit

* **New Features**
  * Added the `npm create authup@latest` interactive wizard.
  * Generates deployment files for Docker, Docker Compose, Helm, or bare-metal installations.
  * Supports configuration prompts for databases, Redis, SMTP, workers, consoles, feature flags, and credentials.
  * Pins generated projects and container images to the wizard’s release version.
  * Prevents accidental overwrites, with an optional force-overwrite flag.

* **Documentation**
  * Added deployment guidance and examples for using the wizard across supported deployment targets.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3558, opened 2026-09-05
fix(server-core): serialize provisioning across replicas with an advisory lock

### LINKED ISSUE #3356: Provisioning is not concurrency-safe: a multi-replica first boot can crash on unique-constraint collisions
Surfaced while reviewing #3354, which adds two more inserts of this shape. Pre-existing and systemic, not introduced there.

## The pattern

Every write in the provisioner is find-then-insert with no guard between the two statements:

- `ProvisioningJunctionSynchronizer.synchronize` (`core/provisioning/synchronizer/junction-synchronizer.ts`): `findOneBy({owner, target})`, then `create` + `save` when absent. This backs every role-permission, user-role, client-role, client-permission and now client-scope binding.
- The per-entity synchronizers (`client`, `role`, `permission`, `policy`, `scope`, `user`): `findOneBy({name, realmId})`, then `save` when absent.
- `WebClientProvisioner.ensureClient` and `ensureScopes` (`core/entities/client/web-client.ts`).
- `ProvisionerModule.assignDefaultPolicy`.

The unique constraints these race against are real: `auth_clients (name, realm_id)`, `auth_scopes (name, realm_id)`, `auth_roles (name, client_id, realm_id)`, `auth_permissions (name, client_id, realm_id)`, `auth_client_scopes (client_id, scope_id)`.

## Consequence

Two or more replicas starting simultaneously against a database that has not been provisioned yet can both take the "absent" branch and both insert. The loser gets a driver-level unique-constraint error out of `save`. `ProvisionerModule.setup` does not catch, and it runs as a boot module (`ApplicationBuilder.withProvisioning`), so that replica fails to start.

Severity is bounded, and worth stating plainly rather than overselling:

- The window is the first boot per realm only. Once the rows exist, every branch is a no-op and the race disappears.
- It is self-healing across a restart: the crashed replica comes back, finds the rows present, and boots. So the practical symptom is a noisy failed start or a brief crash-loop in a scaled deployment, not a permanent outage or data corruption.

It is still worth closing, because the failure is confusing (a driver error at boot with no context), it scales with replica count, and "deploy N replicas of the container" is the documented production topology.

## Why it was not fixed in #3354

Guarding only the two new client-scope inserts would have been arbitrary while every sibling write has the same shape. There is also an architectural constraint worth deciding up front: the existing precedent for swallowing this is `ConsentRepositoryAdapter.insertMissing`, which catches via `isUniqueConstraintDatabaseError` (`adapters/database/errors/driver.ts`, the driver-code unwrapper covering mysql, postgres and sqlite). That helper lives in the adapter layer, and `core/` must not import from adapters, so the synchronizers cannot simply call it where they are.

## Options

1. Push the guard into the repository adapters, so a junction/entity `save` that loses the race resolves to the existing row instead of throwing. Keeps `core/` clean and fixes every call site at once. Closest in spirit to `insertMissing`.
2. Add an explicit upsert-shaped method to the repository ports (`ensure`/`findOrCreate`) that the synchronizers call instead of find-then-save, with the driver-error handling in each adapter.
3. Serialize provisioning across replicas with an advisory lock. Heavier, but it also covers ordering hazards the per-write guards do not, such as two replicas interleaving the policy tree sync.

My preference is 2: it makes the intent visible at the call site rather than hiding a swallow inside `save`, and the ports already carry per-entity semantics.

Whatever is chosen should also decide whether `ProvisionerModule.setup` failing should be fatal at all, or whether a replica that loses a provisioning race should log and continue. `RealmService.save` already takes the non-fatal stance for the runtime web-client path.

tada5hi:
Closes #3356.

## What the issue gets right, and what it misses

The issue reports a crash: two replicas booting against an unprovisioned database both take the "absent" branch of a find-then-insert pair, and the loser gets a driver-level unique violation out of `save`. That is real. `auth_realms(name)`, `auth_clients(name, realm_id)`, `auth_users(name, realm_id)` and every junction (each unique over its two id columns, both NOT NULL) genuinely collide, and the first statement in the boot graph that can reach it is the master-realm insert.

But that is only half of what happens, and the quieter half is worse:

| table | unique key | on a concurrent second insert |
|---|---|---|
| `auth_realms`, `auth_clients`, `auth_users`, junctions | NOT NULL columns | raises, replica fails to boot |
| `auth_permissions`, `auth_roles` | `name, client_id, realm_id`, **both nullable** | silently duplicates |
| `auth_scopes`, `auth_policies` | `name, realm_id`, **nullable** | silently duplicates |

`DefaultProvisioningSource` declares every permission, role, scope and policy as global, so `client_id` and `realm_id` are null. All three supported dialects treat nulls as distinct in a unique index, so nothing raises: both replicas insert the entire global permission catalogue, both global roles, every global scope and every system policy. Nothing cleans them up afterwards — `PolicyProvisioningSynchronizer.cleanupStaleTopLevel` only prunes names absent from the declared set, so a duplicate of a *declared* name survives.

This is why the issue's preferred options 1 and 2 were not taken. Both catch the duplicate-key error and re-read; for those four types there is no error to catch. Serialization does not care whether the database refuses the second insert, so it is the only shape that closes both halves.

## The change

`ProvisionerModule.setup` is renamed to `protected provision(container)`, body untouched, and a new `setup()` wraps it in `withDatabaseLock` (`adapters/database/helpers/advisory-lock.ts`): `pg_try_advisory_lock` on postgres, `GET_LOCK` on mysql, a passthrough on better-sqlite3. The helper is generic mechanism and the lock identity is a parameter, so the provisioning-specific `PROVISIONING_DATABASE_LOCK` lives with the provisioning module rather than in the database layer; the identity is bound, never interpolated into the statement. No write site is guarded, no repository port grew a method, boot stays fatal.

That last point is deliberate. Provisioning seeds the authorization graph, so a replica that comes up without the master realm, the admin user or the permission catalogue passes a container health check while being useless. On lock timeout the helper throws rather than proceeding unlocked, because proceeding *is* the pre-fix behaviour: failing is what today's losing replica does anyway, and the deployment restarts it, by which point the winner has finished and the pass is a no-op.

Four mechanics are load-bearing rather than incidental, and each is a way this could have shipped silently broken:

- The lock is **session-scoped**, so it takes a dedicated query runner. `dataSource.query()` would acquire and release on different pooled connections.
- It is **counted** in both dialects, so it is acquired exactly once.
- It is **released explicitly** before the runner goes back to the pool. Skipping that leaks it for the process lifetime, and a second `setup()` then deadlocks against itself.
- The acquisition answer is **normalized, not tested for truthiness**. Traced against a live server: postgres returns `[{"acquired":true}]` and mysql2 returns the *string* `[{"acquired":"1"}]` / `[{"acquired":"0"}]`. A truthiness check reads a lock held by another session as acquired and makes the entire mechanism inert on mysql.

better-sqlite3 creates no runner at all: one database file per container means a second replica cannot reach it, and `BetterSqlite3Driver.createQueryRunner` hands out one shared runner (`this.queryRunner ??= ...`), so holding one would nest inside whatever else is running. Same shape and same reasoning as `isDatabaseTypeRowLockable`.

## Verification

- `test/unit/adapters/database/advisory-lock.spec.ts`, 16 cases over a fake DataSource, runs on every dialect. Neutering `withDatabaseLock` to an unconditional passthrough fails 15 of them (the survivor is the sqlite case, a passthrough either way).
- `test/unit/modules/provisioning-concurrency.spec.ts` boots two `ProvisionerModule` instances concurrently against one empty database and asserts single rows for the realm/client/user **and** for a global permission, role, scope and policy — the assertions no catch-based design could pass. Gated to mysql/postgres. Neutered, it fails with `duplicate key value violates unique constraint "UQ_9b95dc8c08d8b11a80a6798a640"`, the `auth_realms(name)` collision the issue reports.
- Passes on postgres and mysql. Traced under the lock: one replica acquires immediately, the other polls (`false,false,true` on pg, `"0","0","0","1"` on mysql) and then provisions.
- Full server-core suite on sqlite: 2371 passed, 10 skipped, 0 failed. `tsc -p tsconfig.test.json` and eslint clean.

## Residuals, stated rather than papered over

- **The schema hole stays open.** The lock prevents boot-vs-boot duplication in those four tables; it does not make it structurally impossible. That needs `NULLS NOT DISTINCT` or an expression index over coalesced sentinels, which TypeORM cannot express in entity metadata, plus a dedupe migration for clusters that already raced. Filed separately.
- **Boot vs a live `POST /realms` is not serialized**, and the two sides are not symmetric: `RealmService.save` wraps each provisioner and logs, while the boot backfill loop has no catch and can still abort. The window is a realm created in the seconds a replica takes to boot, and the restart reconciles it.
- **`RealmService.save` runs no `checkUniqueness`**, so two concurrent `POST /realms` with one name give the loser an unmapped 500 rather than a 409.
- **On mysql the mutex is instance-wide**, not per-database, since a named lock is scoped to the mysqld instance while a postgres advisory lock carries the database OID. Two authup databases on one server delay each other's first boot rather than corrupting it. Confirmed by probing, and recorded in the code comment.

Docs: a `Concurrent boots: the provisioning lock` subsection in `.agents/architecture.md` and a paragraph in `.agents/testing.md`.

## Summary by CodeRabbit

* **New Features**
  * Added deployment-wide locking during database provisioning to prevent duplicate records and startup failures when multiple instances initialize concurrently.
  * Provisioning now waits for an available lock and reports a failure if it cannot be acquired within the timeout.

* **Documentation**
  * Added documentation describing provisioning concurrency behavior, locking mechanics, and related test coverage.

* **Tests**
  * Added coverage for lock acquisition, polling, timeouts, cleanup, and concurrent provisioning across supported databases.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1250,9 +1250,10 @@ own, and a client in the same realm block may bind them via `realmScopes`.
 Every write in the provisioner is find-then-insert with no guard between the
 two statements, so two replicas booting against an unprovisioned database
 interleave. `ProvisionerModule.setup` therefore holds a deployment-wide mutex
-around the whole pass (`withProvisioningLock`,
-`adapters/database/helpers/advisory-lock.ts`) and calls the untouched body as
-`provision()`. Nothing else changed: no write site is guarded, no port grew a
+around the whole pass (`withDatabaseLock`,
+`adapters/database/helpers/advisory-lock.ts`, holding the
+`PROVISIONING_DATABASE_LOCK` identity the provisioning module owns) and calls
+the untouched body as `provision()`. Nothing else changed: no write site is guarded, no port grew a
 method, and boot stays fatal.
 
 **One mutex rather than a guard per write site, because half the failures
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -1245,6 +1245,87 @@ one into every realm would fight the system MERGE on every boot).
 Scopes run first because they are leaf entities carrying no relations of their
 own, and a client in the same realm block may bind them via `realmScopes`.
 
+### Concurrent boots: the provisioning lock (issue #3356)
+
+Every write in the provisioner is find-then-insert with no guard between the
+two statements, so two replicas booting against an unprovisioned database
+interleave. `ProvisionerModule.setup` therefore holds a deployment-wide mutex
+around the whole pass (`withProvisioningLock`,
+`adapters/database/helpers/advisory-lock.ts`) and calls the untouched body as
+`provision()`. Nothing else changed: no write site is guarded, no port grew a
+method, and boot stays fatal.
+
+**One mutex rather than a guard per write site, because half the failures
+cannot raise an error to catch.** The insert sites split into two groups with
+opposite failure modes, and the issue text describes only the first:
+
+- `auth_realms` (unique on `name`), `auth_clients` and `auth_users` (unique on
+  `name, realm_id` with `realm_id` NOT NULL) and every junction (each unique
+  over its two id columns, both NOT NULL; the nullable `*_realm_id` and
+  `policy_id` companions are not part of the key) genuinely collide. The loser gets a driver-level unique violation
+  out of `save`, and since provisioning is a boot module that replica fails to
+  start. This is the reported crash, and the first statement that can reach it
+  is the master-realm insert, since the graph runs policies, permissions,
+  roles and scopes before realms.
+- `auth_permissions` and `auth_roles` (unique on `name, client_id, realm_id`)
+  and `auth_scopes` / `auth_policies` (unique on `name, realm_id`) are unique
+  over a tuple containing a NULLABLE column, and every row the default source
+  declares for them is global, so both columns are null. All three dialects
+  treat nulls as distinct in a unique index (postgres defaults to `NULLS
+  DISTINCT`, mysql and sqlite permit repeated nulls), so nothing raises and
+  both replicas simply insert. Nothing cleans them up afterwards
+  (`PolicyProvisioningSynchronizer.cleanupStaleTopLevel` only prunes names
+  absent from the declared set, so a duplicate of a declared name survives).
+
+The second group is why the issue's own options 1 and 2, which catch the
+duplicate-key error and re-read, were not taken: there is no error to catch
+for the entire global permission catalogue, both global roles, every global
+scope and every system policy. Serialization does not care whether the
+database refuses the second insert, so it is the only shape that closes both
+halves. Making the second group structurally safe needs `NULLS NOT DISTINCT`
+or an expression index over coalesced sentinels, which TypeORM cannot express
+in entity metadata; that is tracked separately and is a schema change, not a
+behaviour one.
+
+**Mechanics that are load-bearing.** The lock is SESSION-scoped in both
+dialects, so it takes a dedicated query runner for its whole lifetime:
+`dataSource.query()` would acquire and release on different pooled
+connections. It is also COUNTED in both, so it is acquired exactly once, and
+it is released explicitly in a `finally` before the runner goes back to the
+pool. Skipping that release leaks the lock onto a pooled connection for the
+lifetime of the process, and a second `setup()` in that process then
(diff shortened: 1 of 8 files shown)


tada5hi:
@coderabbitai pause

PR #3560, opened 2026-09-06
feat: record an audit event per back-channel logout delivery

tada5hi:
Closes plan 064 (OIDC back-channel logout) by delivering its last open item, stage 3: delivery outcomes become audit events, so an operator can see which relying parties actually received the push. Until now the only trace of one was a `warn` line on a failure.

## What it does

`OAuth2BackchannelLogoutNotifier` takes an optional `eventService` and records one `auth_events` row per delivery:

| | |
|---|---|
| `backchannelLogout` | the RP answered `2xx` |
| `backchannelLogoutFailed` | anything else |

The row is about the **client** (`refType: client`, `refId` and `clientId` the RP's id), carries the ended session under `sessionId`, is attributed to the session's subject like the sibling `logout` row, and is realm-scoped to the client's realm like the token's `iss`. `data.jti` is the logout token's id, so a row can be matched against the RP's own log of what it received; a failed row adds `status` (the RP answered, with that code) or `errorCode` (it never answered).

```
GET /events?filter[name]=backchannelLogoutFailed
```

## The failed row carries a code, never the rendered error

`describeCauseChain` is the log renderer: it appends `address=` and `port=`, a DNS failure names the RP hostname, a TLS failure the certificate's altnames. The row is attributed to the session's subject, whom `EventService.getMany` answers with their own rows self-service, while `backchannelLogoutUri` itself is withheld from that user (absent from `ClientSummary`, gated by `CLIENT_READ`). So the rendering stays in the `warn` line, which is the log's, and `describeFailureCode` walks the cause chain for the first `code` instead (`ECONNREFUSED`, `ENOTFOUND`, `ERR_TLS_CERT_ALTNAME_INVALID`, else the error's name, `TimeoutError` for the abort). A spec pins that the address never reaches the row.

## Deliberately not included

No `skipped-no-uri` event. A client without a logout URI never enters the audience, and a row per silent client per logout is noise rather than a trace.

## Scope

Two `EventName` members, an optional context field on the notifier, one `record()` method, one wiring line in `AuthenticationModule`. No new port and no per-site hooks: `SessionManager.revoke` was already the single chokepoint, and `IEventService.record` is contractually non-throwing, so a failed audit write can never reach the revoke. No migration: `auth_events.name` is a `varchar(64)` with no database enum.

## Verification

| Check | Result |
|---|---|
| server-core suite | 2357 passed, 7 skipped |
| notifier unit spec | 13 passed |
| HTTP back-channel spec | 5 passed, incl. two new end-to-end row assertions |
| `check:types`, eslint | clean |

Docs updated: the back-channel section of `.agents/architecture.md` and its emit-site list, plus the OAuth2 API guide.

## Summary by CodeRabbit

- **New Features**
  - Added security event logging for OAuth 2.0 back-channel logout deliveries.
  - Successful deliveries and failed or refused deliveries now generate distinct audit events with relevant session, client, token, status, or failure details.
  - Documented access rules and behavior for back-channel logout audit records.

- **Tests**
  - Added coverage for successful deliveries, refusals, delivery errors, timeouts, and signing failures.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -4034,7 +4034,7 @@ local-only — the round-trip is the chosen mechanism, **not** a
 `DELETE /sessions/@me` (which would collide with the #3191 interactive-login
 session reuse → self-DoS of fresh logins).
 
-### Back-channel logout (OIDC Back-Channel Logout 1.0, plan 064 Stages 1+2)
+### Back-channel logout (OIDC Back-Channel Logout 1.0, plan 064)
 
 `Client.backchannelLogoutUri` (`auth_clients.backchannel_logout_uri`, nullable
 `varchar(2000)`; one absolute http(s) URL of at most 2000 characters, no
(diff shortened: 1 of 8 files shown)


tada5hi:
@coderabbitai pause

Issue #895, opened 2023-02-24
Feature: Allow locking/unlocking user name modification


tada5hi:
modify user-form to allow with given permission to lock/unlock the modification of the user name.

tada5hi:
2fcb2c5e50c62aa727b0109dd1dff0647b699231

Issue #893, opened 2023-02-23
Feature: Token verification optimization


tada5hi:
decode token by requesting the public key of the realm, and verify the payload with it.

PR #3561, opened 2026-09-06
feat(server-core): project the query schema registry into the OpenAPI document

tada5hi:
Implements plan 077. The OpenAPI document becomes a build-time **projection of the runtime rapiq schema registry**, so decode, the `meta.schema` wire key and the document are three views of one source and cannot drift.

Before this, the document said nothing about the query language: `parameters: []` on every collection read, so `filter` / `fields` / `sort` / `include` / `page` were reachable only by probing an endpoint or reading `core/entities/*/schema.ts`.

## What ships

**Enrichment, with nothing after the generate.** `@DQuerySchema(EntityType.X, 'collection' | 'record' | 'filters')` marks each query-capable read. A trapi preset handler turns it into an `x-query-schema` extension **and contributes that operation's query parameters during generation**; the shared error response is document-wide; `data.extra` carries the root `x-authup-schemas` (every schema's `describeQuerySchema()` output, verbatim), the hash and the `ErrorResponse` component. `build:swagger` is a plain `trapi generate` — no script, no post-processing. The invariants live in `test/unit/http/openapi-coverage.spec.ts`.

**`GET /schemas` + `GET /schemas/:name`.** `ForceLoggedInMiddleware`, no permission gate, behind `querySchemaDiscoveryEnabled` (`QUERY_SCHEMA_DISCOVERY_ENABLED`, default true). Typed client: `client.schema.getMany()` / `getOne(name)`.

**Serving.** `GET /docs/openapi.json` (the document was previously only inlined into HTML, so none of this was machine-reachable), a boot staleness warning when the built document's hash differs from the running registry, and a guard so a missing or malformed document no longer crashes the server at boot.

**Docs.** A committed grammar page plus a generated per-entity reference, split by rot rate: the operator grammar changes when rapiq's dialect does, the per-entity keys whenever a schema does.

## Baseline OpenAPI fixes

Most of these were authup working around trapi 2.0.5 gaps. Those gaps are fixed in **trapi 2.1.0**, which this branch adopts, so the workarounds are gone and the fixes are upstream:

| Defect | Now |
|---|---|
| tsconfig `paths` inert, so typed payloads silently depended on `packages/*/dist` (204 schemas → 30 without them, exit 0) | trapi resolves `extends` (#893) |
| 61 of 152 templated operations declared no `realmId` — invalid OpenAPI, uncallable from generated clients | trapi declares every path variable (#896); we only add descriptions |
| 186 of 227 operationIds positionally numbered, so inserting a controller renamed unrelated operations | `operationIdStrategy: 'path'` (#897) |
| `tags: []` on every operation despite 29 `@DTags` | v3 tag cascade restored (#894); 208 operations now tagged |
| two console controllers emitted at the document ROOT and **half their operations were silently dropped** | mounts spelled inline (#906); suffixed ids went 2 → 0 |
| error responses | shared `ErrorResponse` component + `default` on every operation |

## Verification

- `npm run build` exit 0; full server-core suite **2384 passed, 0 failed**
- `check:types` clean across server-core, core-http-kit, server-config; eslint 0 errors
- Two consecutive `build:swagger` runs byte-identical
- `test/unit/http/openapi-parity.spec.ts` boots a real server and deep-equals the document against a **live** `meta.schema`, so the central claim is falsifiable rather than asserted
- Guard failure paths exercised by hand: mismarking a route, wrapping a `describeQuerySchema` call across lines, an unused exclusion entry, and a foreign `ErrorResponse` component each abort the generate

## Review notes

Plan 077 was written against an older toolchain and is wrong in several places; the commits record each correction. Most consequential: it specifies assigning `operation.parameters`, which would have **deleted the path parameters of 131 operations**, and a registry-hash ETag that cannot work because routup overwrites handler-set ETags.

One thing worth a look: **the query vocabulary is public.** `/docs` has no auth and `middlewareSwagger` defaults true, so `/docs/openapi.json` serves all 26 descriptions anonymously. Accepted deliberately — `/docs` already published every entity component schema — and `.agents/architecture.md` now says so rather than claiming otherwise.

## Upstream

Filed from this work, all fixed in trapi 2.1.0: #893 (tsconfig `extends`), #894 (v3 tag cascade), #895 (`draft.operationId`/`parameters` discarded), #896 (path-template variables), #897 (positional operationIds), #898 (`swagger.transform` hook).

Also filed and fixed in **trapi 2.1.1**, which this branch adopts: #906 (constant decorator paths), #907 (`draft.parameters`), #908 (document-wide responses). Those three are what let the post-generation step be deleted outright and the console mounts read their constants again.

One residual, not filed: a contributed parameter is appended after the derived ones, so it cannot describe a path variable trapi synthesizes. 61 operations therefore lose the `realmId` description, which was the only place the document said a realm is addressable by name as well as by id.

## Summary by CodeRabbit

* **New Features**
  * Added authenticated query-schema discovery through `GET /schemas` and `GET /schemas/:name`, including supported parameters and registry metadata.
  * Added schema discovery access through the JavaScript client.
  * OpenAPI documentation now includes query-schema details and is available as raw JSON at `/docs/openapi.json`.
  * Query schemas are surfaced consistently in API responses, OpenAPI documentation, and generated query references.

* **Documentation**
  * Added comprehensive Query Language and Query Reference guides.

* **Configuration**
  * Added a setting to enable or disable query-schema discovery routes.


tada5hi:
@coderabbitai pause

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3565, opened 2026-09-07
fix(server-core): enforce uniqueness of global permissions, roles, scopes and policies

### LINKED ISSUE #3559: Global permissions, roles, scopes and policies have no enforceable uniqueness: their unique key contains a nullable column
Surfaced while fixing #3356. The advisory lock landed there prevents this in practice for the boot path; the schema hole itself is untouched, and it is the reason a catch-the-duplicate-key fix was impossible.

## The hole

Four tables are unique over a tuple containing a nullable column:

| table | unique key | nullable members |
|---|---|---|
| `auth_permissions` | `name, client_id, realm_id` | both |
| `auth_roles` | `name, client_id, realm_id` | both |
| `auth_scopes` | `name, realm_id` | `realm_id` |
| `auth_policies` | `name, realm_id` | `realm_id` |

All three supported dialects treat NULLs as distinct in a unique index: postgres defaults to `NULLS DISTINCT`, mysql and sqlite permit repeated NULLs. So for any row whose `realm_id` (and `client_id`) is null, the index enforces nothing at all.

That is not an edge case. `DefaultProvisioningSource` declares **every** permission, role, scope and policy as global, so the entire built-in catalogue sits in exactly the shape the index does not cover: the whole of `Object.values(PermissionName)`, both global roles, every global scope and every system policy.

Nothing reconciles it afterwards. `PolicyProvisioningSynchronizer.cleanupStaleTopLevel` prunes only names absent from the declared set, so a duplicate of a *declared* name survives every subsequent boot.

## Why it matters beyond #3356

#3356 closed the boot-vs-boot window with a mutex, which is why this is a follow-up rather than a blocker. But the mutex is a behavioural guard on one code path, not an invariant:

- A booting replica racing a live `POST /realms` is not serialized (stated as a residual in #3356).
- Nothing prevents a future writer, a migration, an operator script or a provisioning source change from inserting a second global row.
- A cluster that already raced before #3356 shipped carries the duplicates today, and no upgrade path removes them.

## What a fix needs

Two parts, and the second is the awkward one:

1. **A real constraint.** Portably that is a unique index over coalesced sentinels, e.g. `UNIQUE (name, COALESCE(client_id, '<sentinel>'), COALESCE(realm_id, '<sentinel>'))`. Postgres 15+ also offers `NULLS NOT DISTINCT`, but mysql has no counterpart, so the expression index is the portable shape. TypeORM cannot express either in entity metadata, so this needs hand-written DDL. That is a documented exception to the generate-never-hand-write rule in `.agents/conventions.md`, alongside the existing precedents.

2. **A dedupe migration**, with an up-front pre-check in the `1786631686318-SortIndexesAndAccountUniqueness` style: abort the boot naming the duplicate groups and the required cleanup rather than failing with a hash-named driver error when the new index is created. Deduping is not merely deleting the extras: the duplicates are referenced by junction rows (`auth_role_permissions.permission_id`, `auth_client_scopes.scope_id`, `auth_permission_policies.policy_id`, ...), so the survivor has to absorb them and the losers' junction rows have to be repointed or dropped as duplicates in turn.

Because the schema-drift gate (`npm run test:schema-drift`) compares the migrated schema against entity metadata, an index TypeORM cannot describe needs deciding up front: either it is expressed in a way the entity can carry, or the gate needs an explicit exemption. That question is worth settling before any DDL is written.

The populated round-trip (`npm run test:migration-latest`) is the gate that would have to pass here, since this touches constraints and rows, and the empty round-trip cannot see either.

tada5hi:
Closes #3559.

## Summary

- `auth_permissions`, `auth_roles`, `auth_scopes` and `auth_policies` are unique over a tuple that contains a nullable column, and every supported dialect treats NULLs as distinct in a unique index, so the key never refused a second row whose tuple holds a NULL: every global row (the whole built-in catalogue), and on permissions and roles every realm-scoped row with no client. Migration `1788793885495-GlobalEntityUniqueness` adds one unique index per table over the same tuple with the NULLs coalesced onto `''` (`::text` on postgres, where the columns are `uuid`), on mysql and postgres. The `UQ_*` constraints stay: they are what the entity metadata describes and they still serve the scoped rows.
- Pre-existing duplicates of any such tuple (the pre-check groups by the full key, NULLs included) abort the migration before any DDL with one message naming every affected table (the `1786631686318` precedent). The merge is left to the operator: the losers are referenced by junction rows, and a migration cannot know which of two same-named global roles a deployment meant. `upgrading.md` carries the detection queries and the complete list of referencing tables and columns.
- The drift question the issue asked to settle first: each entity declares the index as `@Index('IDX_auth_<table>_global_name', { synchronize: false })`, typeorm's name-only unsynced overload. The schema builder matches such an index by name and neither creates nor drops it, so `test:schema-drift` and `migration generate` see nothing while the DDL is hand-written. The name is given rather than derived, for two reasons found by trying the derived one first: `IndexMetadata.build` names an unsynced index from `givenName` alone, and on mysql the derived `IDX_<hash>` over that column list is already the `@Unique`'s own index (`1740991051622-Default`). Recorded in `conventions.md` as the one hand-named constraint, with the typeorm facts in `references/typeorm.md`.
- sqlite keeps the nullable unique alone: it never runs migrations, and one database file per container leaves no replica to race, the same reason the #3356 provisioning lock is a passthrough there.

## Verification

- `apps/server-core/test/unit/adapters/database/global-uniqueness.spec.ts`, gated to mysql and postgres: boots an empty synchronized database, runs the migration's own `up()`, and asserts the entity-declared name matches the migration's, that `createSchemaBuilder().log()` reports the same drift before and after, that typeorm's own loader sees the index, that a second global row with one name is refused on all four tables while one name per realm plus once globally stays allowed, that `down()` is real, and that the pre-check names the tables holding duplicates. Green on both.
- Full server-core sqlite suite (214 files) and `check:types` green.
- `migration run`, `migration revert`, `migration run`, `test:schema-drift` and `test:migration-latest` on postgres and mysql against scratch databases.

## Summary by CodeRabbit

- **New Features**
  - Added database-enforced name uniqueness for permissions, roles, scopes, and policies, including globally scoped records, on MySQL and PostgreSQL.
  - Migrations now detect existing duplicate records and provide actionable errors before applying constraints.

- **Documentation**
  - Updated upgrade guidance and database documentation to explain uniqueness behavior, affected records, and required cleanup.
  - Added coverage for database-specific uniqueness and migration scenarios.


tada5hi:
@coderabbitai pause

PR #3569, opened 2026-09-07
feat(server-core): encrypt identity-provider secrets at rest

tada5hi:
Plan 070 Stage 2, the last of the realm-cipher consumers.

## Summary

- The OAuth2/OIDC `clientSecret` and the LDAP bind `password` of an identity provider are stored as realm cipher blobs (`v1.<key_id>.<payload>`) under the provider realm's encryption key, the same auto-generated per-realm key that protects MFA seeds and client secrets in encrypted mode. Exactly these two attribute names (`IDENTITY_PROVIDER_SECRET_ATTRIBUTES`) are affected; every other attribute is stored as written.
- `IdentityProviderRepositoryAdapter` is the one choke point: no API writes provider attributes directly, so `saveWithEA` encrypts on a copy (and any secret riding the entity in place, since the extra-attribute adapter lifts non-column properties into the rows too) and puts the plaintext back on the returned entity, so the write's response and the admin form never see a blob; every extra-attribute extension (`findOneById`, `findOneByName`, `extendOneWithEA`, `findByProtocol`) decrypts afterwards, so the protocol factories and the account manager are untouched. A legacy plaintext passes through a read and is encrypted by the next save (no migration; the blob prefix is the discriminator, as for client secrets). A blob the cipher cannot open (unknown, disabled or foreign key) is dropped from the entity, so a login through that provider fails closed rather than presenting ciphertext upstream.
- The adapter moved behind `IdentityInjectionKey.ProviderRepository` as a lazy factory: the identity module is set up before the oauth2 module that registers `OAuth2InjectionToken.RealmCipher`, so the cipher can only be resolved on first use. The controller factory and the LDAP collection authenticator both resolve the token; the duplicate construction in the controller factory is gone.
- `KeyRepositoryAdapter.countBlobReferences` counts provider attribute blobs, so `DELETE /keys/:id` answers `409` while a provider depends on the key and `force` crypto-shreds the secret like an MFA seed.
- Read surface unchanged, per the plan's leaning and the plan-105 precedent: a reader the #3480 gate admits gets the plaintext back on `GET /identity-providers/:id`; masking would break the console's edit form for a reader who could rotate the value anyway.
- Docs: `upgrading.md` (what changes, the lazy re-encrypt and how to finish it, the key-lifecycle coupling), `configuration-server-core.md` and `api-oauth2.md` (the KEK and the 409 now cover provider secrets), `architecture.md` (Realm Key Store consumers).

## Verification

- New `test/unit/http/controllers/entities/identity-provider/secrets.spec.ts`: the OAuth2 secret and the LDAP password are stored as blobs and answered as plaintext on create and read; a rotation re-encrypts and a partial update that omits the secret keeps the blob and still decrypts; a legacy plaintext row passes through and is encrypted by the next save; the realm key's lifecycle binds the secret (delete answers 409, disabled withholds it, re-enabled recovers it, `force` shreds it).
- The provider CRUD, account-link, cookie-login, federated authorize-grant, account-manager and key specs pass unchanged, so the encrypted secret reaches the fake upstream exactly as the plaintext did.
- Full server-core sqlite suite, `check:types`, lint and the docs build green.

## Summary by CodeRabbit

* **Security**
  * OAuth2/OIDC client secrets and LDAP passwords are now encrypted at rest.
  * Existing plaintext provider secrets remain readable and are encrypted when the provider is next saved.
  * Disabled or deleted encryption keys prevent provider authentication and may block key deletion while secrets remain referenced.

* **Documentation**
  * Updated configuration, upgrade, and API guidance to explain identity-provider secret encryption and migration.

### DIFF
--- a/.agents/architecture.md
+++ b/.agents/architecture.md
@@ -6548,10 +6548,17 @@ must be visible in `auth_events`). The table was folded into migration
   the LDAP bind `password`) inside `saveWithEA` and decrypts them after
   every extra-attribute extension, so the protocol factories and the
   admin form see plaintext while `auth_identity_provider_attributes`
-  holds blobs. A legacy plaintext passes through a read untouched and is
-  encrypted by the next save; a blob the cipher cannot open (unknown,
-  disabled or foreign key) is DROPPED from the entity, so a login through
-  that provider fails closed instead of presenting ciphertext upstream.
+  holds blobs. The write side never sniffs its input, the plan-105
+  `protect()` rule: every value a caller writes is encrypted, a secret
+  that starts with `v1.` included, because a read reveals and so nothing
+  legitimately echoes a blob back, and keeping a prefix-matched value raw
+  would let a caller bypass the encryption while a genuine `v1.` secret
+  could never round-trip (stored raw, dropped by the next read). The
+  prefix stays the discriminator on the READ side only: a legacy plaintext
+  passes through a read untouched and is encrypted by the next save; a
+  blob the cipher cannot open (malformed, unknown, disabled or foreign
+  key) is DROPPED from the entity, so a login through that provider fails
+  closed instead of presenting ciphertext upstream.
   The adapter is registered under `IdentityInjectionKey.ProviderRepository`
   as a lazy factory, because the identity module is set up before the
   oauth2 module that registers the cipher; `countBlobReferences` counts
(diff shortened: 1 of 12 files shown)


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3584, opened 2026-09-11
feat: expose complete authorization through introspection

### LINKED ISSUE #3581: Expose complete authorization through introspection for scope-aware resource servers
## Problem

Authup's token-introspection permission extension does not provide enough information for a resource server to enforce Authup's authorization decisions. It exposes permission names and definition namespaces, but loses the actor's grant scopes, grant policies, and permission-definition policies.

This blocks a scope-aware fix for [PrivateAIM/hub#1867](https://github.com/PrivateAIM/hub/issues/1867): Hub's audit-event reads currently check permission names without server-side realm containment. We want to enforce Authup's actual grants, including `none`, `own`, `ownOrNull`, and `any`, instead of introducing another master-realm-name exception.

Verified against **Authup 1.0.0-beta.65** (clean installed packages and the matching source checkout).

### Reproduction

Give an identity a permission through two grants:

- An unrestricted `own` grant.
- An `any` grant restricted by an attribute policy.

Introspection reduces those grants to one permission entry with `name`, `realm_id`, and `client_id`. A downstream service cannot distinguish that identity from one with an unrestricted `any` grant, or recover which policy gates cross-realm access.

Copying `realmScope` into `PermissionMemoryProvider` alone is also insufficient. This probe uses the public API of the installed package:

```ts
import {
    BuiltInPolicyType,
    PermissionEvaluator,
    PermissionMemoryProvider,
    PolicyData,
} from '@authup/access';

for (const realmScope of ['none', 'own', 'ownOrNull', 'any'] as const) {
    const evaluator = new PermissionEvaluator({
        provider: new PermissionMemoryProvider([{
            permission: { name: 'event_read' },
            realmScope,
        }]),
    });

    const data = new PolicyData({
        [BuiltInPolicyType.IDENTITY]: {
            id: 'user-a',
            type: 'user',
            realmId: 'realm-a',
        },
    });

    console.log(realmScope, await evaluator.compile({ name: 'event_read', data }));
}
```

All four results are `{ verdict: 'allow' }`. This is a boundary mismatch, not a claim that the generic evaluator should unconditionally change semantics: Authup's application-level permission-binding evaluator currently supplies the separate realm-reach enforcement.

## Root Cause Analysis

There are three relevant boundaries:

1. **Introspection projection is lossy.** Identity bindings are collapsed by permission key. Their scopes and policies are omitted; the singular optional `policy` in the wire type is not populated.
2. **Authorization has two layers.** Authup applies permission-definition policies, whose permission-binding checks resolve the identity's grants. Each grant's reach is paired with its own policy. Exporting only identity-junction grants would still omit definition-level restrictions.
3. **The reusable evaluator is not the entire server authorization stack.** It evaluates grant policies; Authup's application-level binding evaluator supplies the scope-and-policy composition. A resource server needs reusable equivalent semantics as well as sufficient wire data.

Permission **namespace** (`name`, `realm_id`, `client_id`) must remain distinct from grant **reach** (`realmScope`). A globally defined permission does not imply global access for every holder. Neither namespace flattening nor independent merging of scopes and policies is safe.

## Proposed Contract

Expose a versioned or explicitly identifiable authorization representation through authenticated introspection, preserving:

- The permission-definition namespace and applicable definition policies/decision semantics.
- The lossless disjunction of identity grants, with each grant's realm reach paired with its own policy.
- Enough context for a resource server to evaluate a concrete resource or compile exact row conditions using supported Authup APIs.

An explicit definition-plus-grants representation is one option; an equivalent effective-policy representation is also acceptable if it preserves both layers and nested composition. The requirement is decision parity, not a prescribed DTO or class layout.

Keep name-only compatibility, if needed, explicitly distinguishable from the new representation. Consumers must be able to reject an old, incomplete, or unsupported authorization payload rather than interpreting missing restrictions as unrestricted access. Inactive tokens must continue to expose no authorization grants.

## TDD Fix Plan

1. **RED:** Through introspection and the supported resource-server evaluator, verify the same permission under `none`, `own`, `ownOrNull`, and `any` yields the expected own/foreign/null-realm decisions. Master-realm membership must not override `own` or `none`.
   **GREEN:** Transport and enforce explicit grant reach with a detectable contract version/capability.

2. **RED:** An unrestricted `own` grant plus a policy-bound `any` grant allows foreign resources only when the latter policy passes; a passing narrow-grant policy cannot borrow another grant's wider reach.
   **GREEN:** Preserve and evaluate the OR of complete scope-and-policy grant terms.

3. **RED:** A permission-definition policy that denies a resource still denies it after the actor's otherwise sufficient grant is transported through introspection.
   **GREEN:** Preserve definition-level policy composition as well as identity grants.

4. **RED:** Same-name permission definitions in different realm/client namespaces cannot authorize one another. A global definition with an `own` grant remains realm-restricted.
   **GREEN:** Preserve namespaces and define matching lookup semantics without name-only collapse.

5. **RED:** Applying compiled authorization to a collection returns only authorized rows and an authorized-row total, consistent with individual evaluation. An unlowerable policy cannot become `allow` or an unrestricted query.
   **GREEN:** Expose reusable evaluation/compilation that preserves grant semantics and explicitly reports policies requiring post-evaluation, allowing consumers to reject unsupported collection queries safely.

6. **RED:** Old, malformed, or unsupported authorization representations cannot silently grant unrestricted access; inactive introspection responses carry no grants.
   **GREEN:** Validate the new contract at the consumption boundary and preserve inactive-token behavior.

**REFACTOR:** Share the grant-enforcement semantics between Authup and external resource servers where practical, avoiding a second independently maintained interpretation.

## Acceptance Criteria

- [ ] A resource server can enforce both authorization layers from a supported introspection contract without querying Authup's management API per request.
- [ ] Scope, policy, mixed-grant, and namespace tests demonstrate parity with Authup's authorization behavior.
- [ ] `own` excludes realm-less resources; `ownOrNull` includes them; `any` permits cross-realm access; `none` permits none. No master-realm-name bypass is required.
- [ ] Compilation preserves the same decisions as concrete-resource evaluation where lowering is supported, and reports unsupported cases explicitly.
- [ ] Legacy payloads are distinguishable; missing authorization data cannot be mistaken for unrestricted permission.
- [ ] Migration guidance explains the wire/library upgrade order and how existing name-only consumers should migrate.
- [ ] New and existing tests pass.

tada5hi:
Introspection currently collapses grants to permission names, preventing resource servers from enforcing realm reach and policy restrictions. This adds a versioned authorization snapshot and `createAuthorizationEvaluator`, preserving definition policies, exact namespaces, and each grant's paired scope and policy.

The server and resource-server consumer share the binding evaluator. Legacy `permissions` remains available; the new consumer rejects incomplete snapshots, unsupported policies, non-access credentials, and tokens without `global` scope. Compiled conditions retain the same resource decisions, with an explicit `post` result when a policy cannot be lowered. Includes migration guidance and fixes for binding inversion and realm-less compilation parity.

Validation:

- Full server-core suite: 2,487 passed, 16 skipped.
- Access and specs suites: 207 passed.
- Affected builds, source/test type checks, ESLint, and documentation build passed.
- Regression coverage includes HTTP introspection round trips, scopes, mixed grants, definition restrictions, namespaces, collection filtering/totals, malformed snapshots, and credential restrictions.

Closes #3581

## Summary by CodeRabbit

- **New Features**
  - Token and console-session introspection responses now include versioned authorization snapshots with identity, permission policies, and realm-scoped grants.
  - Added authorization evaluation for validating permissions, checking individual resources, and filtering collections before pagination.
  - Authorization snapshots support nested policies, grant restrictions, and compiled resource queries.

- **Documentation**
  - Added guidance for consuming authorization snapshots with the JavaScript SDK.
  - Clarified when to use snapshot-based authorization instead of name-only permission checks.

- **Bug Fixes**
  - Improved handling of inactive credentials and null or restricted realm matches.


tada5hi:
@coderabbitai pause

PR #3588, opened 2026-09-13
feat: serve the authorization catalog at GET /authorization and the identity's grants on introspection

### LINKED ISSUE #3581: Expose complete authorization through introspection for scope-aware resource servers

tada5hi:
Closes #3581 (the transport #3584 chose is reshaped here; #3584 is unreleased, so nothing below is a wire break for a released consumer).

## What changes

Authorization outside the Authup process is built from three inputs, two of which every consumer already has.

- **The catalog**, `GET /authorization`: every permission definition (`name`, `realm_id`, `client_id`, `decision_strategy`, `policies` ids) plus each policy tree a definition or a grant can name, once, under `policies`. It is identity-free, so a process fetches it once and caches it (`Cache-Control: private, no-cache`). It is gated by the same `PERMISSION_READ` / `PERMISSION_UPDATE` / `PERMISSION_DELETE` pre-gate `GET /permissions` runs, so a resource server reads it with its own client credential.
- **The grants**: the introspection responses (`POST /token/introspect`, `GET /sessions/@me/introspect`) report the identity's grant list under `permissions`, one entry per junction row with its `realm_scope` and junction `policies` ids. The array used to collapse grants to names.
- **The identity**: `sub`, `sub_kind`, `realm_id`, `realm_name` from the same introspection.

`createAuthorizationEvaluator({ catalog, grants?, identity? })` in `@authup/access` rebuilds the server's own raw binding model (definitions into the memory provider, grants into the binding evaluator) and runs the same aggregation and evaluators, so decisions match by construction. A policy tree on the wire is the output of its type's validator, so entity columns never travel and the server-side projection and the consumer-side validation are one function. Both `grants` and `identity` are optional: without an identity nothing is injected and no grant is bound, so a definition whose policies need no identity (a `time` or `date` policy alone) passes while a `system.default`-bound one denies, exactly as an anonymous request evaluates server-side.

The kit store keeps one catalog per signed-in session, derives identity and grants from the introspection it already runs, and commits an evaluator into one stable `store.permissionEvaluator`. `usePermissionCheck` keeps calling `preEvaluateOneOf`, and a check carrying `realmMatch` settles reach per row.

## Failure posture

Every failure denies; none of them ends a session or fails the catalog as a whole.

- A definition whose policy tree the server cannot project is not omitted: it travels with `policies: null`, and a consumer denies that permission and drops its grants. An ABSENT definition therefore means exactly one thing, a catalog older than the definition, so the consumer refetches once.
- A consumer that still cannot resolve a grant after that refetch, or that meets a policy type it does not carry, denies the affected permission and keeps the session; the kit commits a deny-all evaluator rather than reverting the login.
- A caller without the permission family reads no catalog; a console then falls back to name-only gating, which is advisory, and the server enforces every decision regardless. A resource server must fail closed instead.
- The consumer follows the server's three-way `REALM_MATCH` rule: a realm id, `null` for a global row (which `own` denies), or no key for a resource with no realm dimension, which neutral-passes.

## Verification

Full builds, the `access`, `core-http-kit`, `client-web-kit` and `server-core` suites, `check:types` across workspaces, lint, the docs build and both smoke scenarios are green. Two independent three-lens Opus audits with adversarial refutation ran over the branch, one over the original design and one over the catalog split; their confirmed findings are fixed here. The most consequential was a parity break the suite could not see: the introspection asked for the grants of the token's CLIENT rather than of the identity, so an authorization-code token, which is how the consoles and every relying party authenticate, reported almost no grants.

## Follow-ups (own issues, not in this PR)

- `projectAuthorizationPolicy` closes its validator set over `BuiltInPolicyType` while `PolicyEngine.registerEvaluator` and the free-string `auth_policies.type` are open, so a custom policy type cannot travel through the catalog (it denies rather than evaluating). A `PolicyDefaultValidators` registry, symmetric with the evaluators, is the fix.
- `PermissionService.delete` / `save` do not drop the three owned-permission junction cache keys, so for up to 60 seconds an introspection can report a grant on a permission the catalog no longer carries; the consumer denies it in that window.
- The introspection is scope-blind: a bearer without `global` reports its full grant list, so a resource server applies its own scope check. Documented; changing it would change released behaviour.
- `IdentityPermissionBindingPolicyEvaluator` recurses without bound on a grant policy that itself contains a `permissionBinding` node (pre-existing).
- A `hash` on the catalog (the `/schemas` `meta.hash` pattern) so a consumer can revalidate without refetching.

## Summary by CodeRabbit

* **New Features**
  * Added an identity-free authorization catalog available through `GET /authorization`.
  * Authorization grants are now provided through token introspection and evaluated alongside the catalog.
  * Added client support for fetching authorization catalogs and building permission evaluators.
  * Added catalog caching, stale-catalog refresh, and safe deny-all fallbacks in web clients.
  * Authorization endpoints now require appropriate read, update, or delete permissions.

* **Documentation**
  * Updated JavaScript SDK guidance for catalog-based authorization, caching, scopes, and evaluator usage.


tada5hi:
@coderabbitai pause

Issue #3574, opened 2026-09-08
Ungated list reads: role / scope / permission / policy findMany carry no per-row realm gate


tada5hi:
## Context

`role`, `scope`, `permission` and `policy` are the global-capable entities: a row may carry `realmId: null` (a system-level building block) or a realm's id. Their `findMany` paths apply the decode-time allow-lists and the `*_READ` pre-gate, but no per-row realm gate: `RoleService.getMany` and its three siblings hand the decoded query straight to the adapter, and the adapters carry no `applyRealmScopeSelect` because there is no per-row `evaluate` to feed.

So a `realm_admin` holding `ROLE_READ` at `ownOrNull` reach lists every realm's roles, not only its own plus the global ones, and the same for scopes, permissions and policies. The `getOne` paths are gated per row (`resourceRealmMatch`), so the single read refuses what the list shows.

## Why it is a decision, not a fix

Wave 0 (plan 039) found this while closing the projection bypass and recorded it as a gap rather than fixing it, because the list surfaces carry no secret: the four entities hold names, display names, descriptions and, for policies, their configuration. Whether a realm-scoped administrator may *see* another tenant's role names is a product question about realm isolation, not a leak of credentials. The fix, once decided, is the compiled-WHERE shape every other realm-gated list already uses (#3286 phase 3): `compile({ name: <ENTITY>_READ })` and append the reach condition, so pagination and totals stay exact. The four services currently skip `compile()` entirely.

## Proposal

1. Decide whether realm isolation covers *visibility* of global-capable rows across realms. If yes:
2. `getMany` in the four services composes `appendQueryConditions(query, compiled.condition)` on a `conditional` verdict, a constant-false condition on `deny`, and the existing per-row drop loop on `post`, exactly like `KeyService` / `TrustAnchorService`.
3. The adapters gain `applyRealmScopeSelect` for the `post` fallback.
4. A spec per entity in the `realm-isolation` family pins that an `ownOrNull` reader lists own-realm and global rows and never a foreign realm's.

## Related

- plan 039 (projection bypass), `.agents/plans/000-roadmap.md` §1 follow-up candidates
- #3286 (compile / WHERE pushdown), the shape the fix would take

PR #3598, opened 2026-09-15
fix(server-core): realm-gate the junction reads, count the device deny, state the post contract

### LINKED ISSUE #3591: A foreign-realm device deny records no authorize metric
## What happens

`OAuth2DeviceAuthorizationService.approve` records `authup_authorize_total{outcome="login_required"}` when the shared `resolve()` refuses a foreign-realm user, matching the `/authorize` ladder. `deny` runs the same `resolve()` and the same refusal but records nothing, so the realm-mismatch outcome counts approvals only (#3587 review, accepted as low).

## Fix

Move the `login_required` classification out of `approve`'s catch into `resolve()` itself (the refusal is the shared code path), or wrap `deny`'s `resolve()` the same way. One service case: a foreign-realm deny answers `login_required` and `recordAuthorize` was called with `login_required`.

## Pointers

- `apps/server-core/src/core/oauth2/device-authorization/module.ts` (`resolve`, `approve`, `deny`)
- `apps/server-core/src/core/oauth2/authorization/helpers.ts` (`classifyAuthorizeFailure`)
- `.agents/architecture.md` → *Device Authorization Grant (RFC 8628)*, the events and metrics paragraph

### LINKED ISSUE #3594: Junction collection reads carry no per-row realm gate, so include= reaches foreign-realm rows
## Context

#3574 / #3593 realm-gated both read paths of the four global-capable entities (`role`, `scope`, `permission`, `policy`). A cross-realm read of those same rows stays open one layer out, through the **junction** collections.

Five junction services apply a name-level pre-gate and then hand the decoded query to the repository with no realm predicate and no per-row drop loop:

- `core/entities/permission-policy/service.ts` `getMany`
- `core/entities/role-permission/service.ts` `getMany`
- `core/entities/client-role/service.ts` `getMany`
- `core/entities/client-scope/service.ts` `getMany`
- `core/entities/user-role/service.ts` `getMany`

Their repositories add none either, and their schemas allow the member include (`permission-policy/schema.ts` allows `['permission', 'policy']`, `role-permission/schema.ts` likewise, and so on). `core/query/relations.ts` runs `preEvaluateOneOf({ name: permissions })` with no `realmMatch` data, so the realmScope factor neutral-passes: the gate checks whether the caller may read that TYPE, never whether it may reach that ROW's realm.

So `GET /role-permissions?include=permission` and its siblings still return another realm's permission, role, scope and policy rows to any caller holding the target triple.

## Severity

Lower than #3574. The joined columns are the row's own metadata (`name`, `displayName`, `description`, `type`, `invert`, `realmId`), and the policy **configuration** does not ride along — the junction repositories call no `extendManyWithEA`, unlike `policy/repository.ts`. So this is cross-realm metadata disclosure to an authenticated junction reader, not the access-control-rule disclosure #3574 closed.

The junction WRITE paths are already gated: `JunctionEntityService.junctionResourceRealm` stamps the owner realm and the write is realm-matched (pinned by `test/unit/http/controllers/security/realm-isolation.spec.ts`). It is only the reads that are open.

## Proposal

Give each junction `getMany` the compiled-WHERE shape #3593 established, using the owner-realm key each class already declares (`permissionRealmId`, `roleRealmId`, `clientRealmId`, `userRealmId`) as the realm-match column, plus `applyRealmScopeSelect` on the matching adapters for the `post` fallback. That closes the junction row leak and the include with it.

The alternative — making the decode-time include gate reach-aware — would need the target schema to contribute a realm condition on the joined alias, which rapiq's per-relation fieldset narrowing would have to support. Worth considering, but larger.

A spec in the `realm-isolation` family, mirroring `global-entity-realm-isolation.spec.ts`: an `ownOrNull` junction reader must not surface a foreign realm's member row through `include=`.

## Related

- #3574 / #3593 (the entity-level gate, and the shape to copy)
- #3286 phase 3 (compile / WHERE pushdown)
- #3295 (the relations read gate, which checks permission not reach)

### LINKED ISSUE #3595: meta.total and page density are approximate on the post compile fallback

**Reported in:** PR #3593, review comments by @coderabbitai
**Files:** every service that falls back to the `post` compile verdict

## Description

When `IPermissionEvaluator.compile()` returns the `post` verdict, a service fetches a page through `repository.findMany(parsed)` and then runs the per-row `evaluateOneOf` drop loop over **that page only**, decrementing `meta.total` once per dropped row:

```ts
const data: T[] = [];
let { total } = meta;
for (const entity of entities) {
    try { await actor.permissionEvaluator.evaluateOneOf({ ... }); data.push(entity); }
    catch { total -= 1; }
}
```

Two consequences:

- `meta.total` still counts unauthorized rows on pages the caller has not fetched, so it overstates the accessible set.
- A page can come back shorter than its `limit`, and in the worst case empty, while later pages still hold authorized rows.

This is the documented trade-off of the `post` fallback, not a defect introduced by #3593. The same block appears verbatim in `KeyService`, `TrustAnchorService`, `SessionService`, `ConsentService`, `UserService` and `UserAuthenticatorService`, and `EventService` / `UserAttributeService` / `RoleAttributeService` reach the same fallback by other means. `.agents/architecture.md` states it: the `conditional` verdict runs the authorization as a WHERE so "pagination and totals stay exact", while `post` is "the sound fallback".

**Reachability is narrow.** With default provisioning every `admin` / `realm_admin` grant is policy-free, so the realm reach lowers to a condition and `compile()` returns `conditional`, where totals are exact. `post` is reached only when a junction grant carries a policy with no `toCondition` — today that means an `ATTRIBUTE_NAMES` policy attached as a junction `policyId`. (`global-entity-realm-isolation.spec.ts` constructs exactly that to cover the branch.)

## Why the obvious fix is worse

The review suggests evaluating the complete matching collection before paginating. That removes the bound the query surface deliberately imposes: every registered schema declares `pagination: { maxLimit: 50 }`, so an unpaginated fetch plus a per-row policy evaluation turns an authenticated list request into an unbounded read. On `auth_permissions` that is the whole catalogue; on `auth_events` or `auth_sessions` it is a table that grows without bound.

## Suggested fix

Prefer shrinking the `post` surface over unbounded fetching:

1. Give `AttributeNamesPolicyEvaluator` a `toCondition`, which is what forces `post` in practice today. It lowers naturally where the policy is a single-key allow/denylist over row attributes.
2. Where a residual `post` is unavoidable, consider a bounded evaluate-then-refill loop (fetch, drop, fetch the shortfall, up to a cap) so a page is at least dense, and document that `meta.total` is an upper bound on that path.
3. Either way, state the `post` contract once in `.agents/architecture.md` rather than leaving it implicit at nine call sites.

## Related

- #3593 (added the fallback to the four global-capable entity reads)
- #3286 phase 3 (compile / WHERE pushdown, which is the exact path)

tada5hi:
Closes #3594, closes #3591, closes #3595.

## #3594 — junction reads carry no realm gate

Every `JunctionEntityService` read applied a name-level pre-gate and then handed
the decoded query straight to the repository: no realm predicate, no per-row drop
loop. So a junction row of another realm was listed, was readable by id, and
carried its member along through `include=`, to any caller holding the target
permission triple. The relations read gate (#3295) cannot catch that — it asks
whether the caller may read that TYPE, never whether it may reach that ROW's realm.

Two deviations from the issue, both deliberate:

- **All EIGHT junctions, not the five listed.** `client-permission`,
  `user-permission` and `identity-provider-role-mapping` carried the identical
  hole; fixing five would have left three open with the same shape.
- **`getOne` ships with `getMany`.** All eight returned the row after the pre-gate
  alone, so `GET /<junction>/<uuid>` disclosed a foreign row without needing
  `include=` at all. #3593 argues that gating only the list is the more dangerous
  direction, and that applies here unchanged.

### The compiled-WHERE shape needed a seam, not a copy

A junction row carries no `realmId`, and `RealmMatchPolicyEvaluator` lowered the
scope-mode reach onto that literal column. A verbatim copy of #3593's shape would
have emitted `eq('realmId', …)` against `auth_role_permissions` and died at the
driver.

The override already half existed — that evaluator reads `attributeName` as the
lowered column, and scope mode ignores it while EVALUATING — so the column name is
threaded through as `PermissionCompileContext.realmAttributeName` →
`PolicyEvaluationContext.realmAttributeName` → the inline `{ scope }` policy the
binding evaluator builds. It can never move an `evaluate()` outcome: settled
scope-mode evaluation takes the resource realm from the `realmMatch` data key and
consults no column name at all. Each service passes `this.ownerRealmKey`, the
field `JunctionEntityService` already declares `abstract`.

### Three load-bearing details

- The **owner** key, never the member key. Gating `user-role` on the `roleRealmId`
  of a usually-global role would deny every own-reach reader.
- The post fallback passes `junctionResourceRealm(entity)` explicitly.
  `resourceRealmMatch` is presence-based on `realmId`, which a junction row never
  has, so it would return `{}` and neutral-pass every foreign row while looking
  correct.
- The adapters force-select the owner key through `applyJunctionRealmScopeSelect`
  (`applyRealmScopeSelect` prepends a `realmId` no junction table has). A client
  `fields=` projection replaces the schema default, and a stripped owner realm
  coalesces to `null`, which an `ownOrNull` reader reaches.

No schema change was needed: all eight already declare their owner-realm key in
`fields.default`, in `filters.allowed` and as an index.

## #3591 — a foreign-realm device deny recorded no authorize metric

The guard moves into `resolveForDecision`, which both decision methods call.
Sharing it rather than duplicating five lines per method is the point: the guard
sitting on `approve` and not on `deny` IS the bug.

`lookup` keeps the bare `resolve`. The issue offers moving the record into
`resolve` itself; that was implemented first and then reverted, because `lookup`
is a page render — the device analogue of the `/authorize` GET, which records
nothing either — and it can never produce the counter's other labels, so
contributing only `login_required` would skew the ratio between them.

## #3595 — `post`-verdict totals and page density

**Documentation, plus two corrections.** Both suggested code fixes are rejected,
and the reasoning is written down:

- Fetching the whole collection before paginating removes the bound
  `pagination: { maxLimit: 50 }` exists to impose, turning an authenticated list
  request into an unbounded read.
- `AttributeNamesPolicyEvaluator.toCondition` **cannot be written soundly.** That
  policy decides a predicate over which KEYS the evaluation bag carries — a write
  payload's field names, or a requested projection — never over a row's values. As
  a row predicate it is row-INDEPENDENT (it flips when the projection narrows, not
  when row data changes), and `compile()` never receives the projection it would
  have to range over. `IPolicyEvaluator.toCondition` already names `attributeNames`
  among the policies that never lower.

Two statements in `architecture.md` claimed "exact totals and pagination" without
saying which verdict that holds for. It is true of `conditional` and false of
`post`; both are corrected, and the contract is now stated once where the verdicts
are described, covering the two drop loops that reach the same shape without
calling `compile()` (`SessionTokenService`, `IdentityProviderAccountService`) and
naming the one `compile()` caller that is not a drop loop (`secretReadGate`, which
fails closed).

## Tests

`apps/server-core/test/unit/http/controllers/entities/junction-realm-isolation.spec.ts`,
mirroring `global-entity-realm-isolation.spec.ts`: a policy-free `ownOrNull` reader
(the `conditional` branch) and a second reader whose grants carry a non-lowerable
ATTRIBUTE_NAMES policy (the `post` branch). It covers all eight junctions on both
read paths, the `include=` disclosure, and the field-projection case.

Both fixes were verified differentially rather than by a green run:

- deleting one `applyJunctionRealmScopeSelect` call fails *holds the gate on the
  post branch* with the foreign row present — so the force-select coverage is not
  theatre (the trap #3593's first pass shipped);
- dropping one `realmAttributeName` fails three tests with
  `no such column: userRole.realmId`;
- reverting `deny` to the bare `resolve` fails the new metric assertion.

`packages/access/test/unit/policy/identity-permission-binding.spec.ts` pins the
lowering onto a named column and the unchanged `realmId` default.

`client-secret-projection.spec.ts` reads cross-realm junction rows to exercise the
secret field gate, so its actor's junction grants widen to `any`. What that spec
tests is the `CLIENT_READ` reach the secret gate compiles, which stays `own`.

Full `apps/server-core` suite: 2582 passed, 28 skipped, 0 failed.
`packages/access`: 196 passed. `check:types` and `eslint` clean on both, docs build
clean.

## Follow-up

The junction gate narrows `own` reach in one place worth stating: `permission-policy`
gates on `permissionRealmId`, and the built-in permission catalogue is global, so an
`own`-reach reader sees no built-in permission's policy bindings — the same narrowing
`own` already means for the permissions themselves since #3574. `realm_admin` holds
these reads at `ownOrNull` and is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced realm-based access controls across junction resource listings and single-item reads.
  * Prevented foreign-realm records from appearing through projections or included relationships.
  * Improved filtering for conditional and post-evaluated permissions, including accurate totals for filtered results.
  * Fixed authorization handling when denying device authorization requests that require login.

* **Documentation**
  * Clarified realm-scoped filtering, approximate totals, and pagination behavior for post-evaluated results.


tada5hi:
## Audit follow-up: one regression this branch introduced, now fixed

An adversarial review of the branch found a defect the green suite did not catch. Two more commits.

### `fix(access): never lower a grant policy onto a foreign row model`

Threading `realmAttributeName` rebased the grant's **reach** onto the junction's owner realm key, but not the grant's own `policyId` policy. That policy is the operator's, written against the **entity** the permission names, and nothing rebases it — so its condition was ANDed onto the junction query naming a column the table does not have:

```
and( or(eq(roleRealmId, X), eq(roleRealmId, null)),   // the reach — right column
     eq(realmId, X) )                                 // the grant policy — no such column
```

Every junction collection read answered **500 `no such column: rolePermission.realmId`** for that actor, where before the gate existed it answered 200. Reachable in two ordinary admin POSTs — `POST /policies` with a `scope` or a `query`, then `POST /client-permissions` with that `policyId` — and the architecture notes retain the `REALM_MATCH` type for exactly such operator policies, so it is a supported configuration rather than a misconfiguration.

Both lowerable built-in shapes hit it: a scope-mode `realmMatch` (the framework's own hard-coded column) and an `ATTRIBUTES` query naming `realmId` (the operator's, and the more natural one to write, since it works on entity reads).

**The class is refused rather than one instance rebased.** Under a declared `realmAttributeName` a pending grant policy contributes no condition, so the compile answers `post` and the per-row branch evaluates that same policy against the junction's real attributes — reaching the identical verdict, since the pushdown and the per-row check agree wherever the pushdown does not crash. An earlier revision of this fix rebased only the realm-match case; it left the attributes case, and the next lowerable policy type would have reopened it.

The grant's **reach** is untouched, so the policy-free grants that dominate real deployments still push down and keep exact totals. Only a policy-bound junction reader pays the `post` approximation.

Verified differentially: with the guard removed, both parameterized HTTP cases fail with the driver error above.

### `docs: correct three claims the branch overstated`

- The junction gate reads the **owner** realm, so "no longer reachable cross-realm as an include target **either**" was too strong. A junction row whose owner is **global** (`roleRealmId`, `permissionRealmId`, both nullable) is reachable by an `ownOrNull` reader, so a realm-scoped member bound onto a global owner still surfaces. Only an `any` actor can create that binding; closing it needs the reach-aware decode-time include gate #3594 calls the larger alternative. Now stated as the residual it is.
- The `IdentityProviderAccountService` rationale went stale inside this same PR — `realmAttributeName` can name `userRealmId`, so it is a conversion candidate rather than a place the fallback is required.
- "five of them OR an ownership term" is six (event, session, consent, user, user-attribute, user-authenticator); pre-existing, but it sits in a sentence this branch rewrites.

### Reviewed and rejected

- **Query cost.** Junction list endpoints go 5 → 17 statements for a `realm_admin` and 9 → 121 for a policy-bound reader, because `PermissionEvaluator` does not memoize and `PermissionDatabaseProvider.findOne` caches only the `auth_permissions` row. Real, but it is the arrival of an authorization gate on endpoints that had none, and the cost is identical to what master already charges every other realm-gated list. A memoization pass belongs in its own change.
- **`own`-reach hides global-owner junction rows.** Same semantics #3574 established for the entities themselves; `realm_admin` holds these at `ownOrNull` and is unaffected.

### State

`apps/server-core` 2585 passed / 0 failed / 28 skipped, `packages/access` 198 passed, `check:types` and `eslint` clean on both, docs build clean.

tada5hi:
@coderabbitai pause

tada5hi:
## Rebased onto #3588, and a downstream impact check

Rebased onto `bd9b7d03c` (the `GET /authorization` catalog). Two conflicts, both real:

- `packages/access/src/permission/authorization.ts` became the `authorization/` directory. Its `compile` still enumerates the fields it forwards, so `realmAttributeName` is re-added there — and it matters more now than it did: #3588's documented use case is a **resource server lowering its own collection queries** through `compile()`, and its rows need not name the column `realmId`. The option stops being internal plumbing for eight junctions.
- The SDK page rewrote that paragraph; merged into its shorter form rather than reimposing mine.

### One correction the rebase forced

The guard's justification was written as *"naming a column declares the rows are not the entity the permission names"*. True for Authup's junctions, **false for a resource server** whose rows the operator's policy really does range over. The behaviour is unchanged and sound — `post` — but the reason is now stated honestly: Authup cannot tell the two cases apart and refuses the lowering either way, with the workaround (leave the option unset, map the column in your query adapter) written down. If a real consumer ever wants that pushdown back, the fix is a separate flag, not a reinterpretation of the column name.

### PrivateAIM/hub: no impact

hub reads Authup's junction API — `clientPermission.getMany` with `relations: { permission: true }`, plus `create` and `delete` — in `analysis-client.ts`, `analysis-client-permission.ts` and `node-client.ts`.

It is unaffected, and not by luck. hub creates each client with `realmId: entity.realmId` (the analysis's or node's realm, not its own), and `ClientService.create` evaluates `CLIENT_CREATE` with `resourceRealmMatch(validated)` — so hub's credential must already reach that realm. The junction rows it then reads are gated on `clientRealmId`, the same realm, under the same permission family its writes already use (`client-permission`'s read pre-gate is `CLIENT_PERMISSION_CREATE` / `_DELETE`).

So the read gate demands exactly what the write path already demanded. A deployment where hub can create and delete these rows but not read them cannot exist. Its reads carry no `fields=` projection either, so the `post`-branch force-select is not even in play.

### Scope

Confirmed with the maintainer: keep the junction gate. What settled it is that the disclosure needs no `include=` at all — `GET /role-permissions?filter[roleRealmId]=<realm B>` plus `GET /permissions/<id>` reads another realm's capability map, because the built-in catalogue is global and `ownOrNull` reaches global rows. Owner-realm gating only; owner+member was considered and rejected (a second `compile()` per read, and the global-owner residual is correct semantics rather than a hole).

### State

`apps/server-core` 2605 passed / 0 failed / 28 skipped, `packages/access` 219 passed, `check:types` and `eslint` clean on both, docs build clean — all on the rebased tree.

Issue #3599, opened 2026-09-15
Cache GET /authorization server-side, keyed by its narrowing parameters


tada5hi:
`GET /authorization` is anonymous and identity-free since #3588: the body is a pure function of the permission and policy rows and depends on no caller. That makes it cacheable on the server, which it is not today — every request reads every permission definition, every bound policy tree and every grant-junction policy, and projects them.

Two coupled pieces of work.

## Narrowing parameters

Take `realmId` and a realm scope (`own` / `ownOrNull` / `any`, read relative to the NAMED realm rather than to a caller, which the route does not have). No parameters keeps today's behaviour: the whole catalog. They are a payload projection, never an access decision.

One correctness rule comes with them. An absent definition currently means exactly one thing to a consumer — the copy it holds is older than the definition — and that is the signal `AuthorizationCatalogStaleError` and the single refetch are built on. A narrowed document would break that: a grant naming a definition outside the requested realm would read as stale, the refetch would answer the same, and `@authup/client-web-kit` commits a deny-all evaluator after the second stale answer. So a narrowed catalog has to declare what it covers (echo the realm and the scope in the document) and `createAuthorizationEvaluator` has to DROP a grant whose definition falls outside that scope instead of reporting stale.

Worth checking whether the payload saving justifies this at all: authup's built-in permission catalogue is entirely global rows, so the document is close to constant-size regardless of realm count.

## The cache

Not a TTL. A clock-based memo makes a routine `POST /permissions` invisible for the length of the window, and a consumer that meets a grant for the new definition refetches once, gets the same stale answer and commits a deny-all evaluator — so a cache that is merely eventually consistent converts an ordinary write into denied sessions.

It needs invalidation off every write the document derives from:

- `auth_permissions` (definitions)
- `auth_permission_policies` (the definition-to-policy assignment)
- `auth_policies` and `auth_policy_attributes` (policy configuration, including the EA rows a policy's config lives in)
- `auth_role_permissions`, `auth_user_permissions`, `auth_client_permissions` (the junction policy ids `findGrantPolicies` collects)

The existing mechanism fits: store the built document under a `CachePrefix` entry in the shared `ICache` and add that key to each of those subscribers' `cache.keys`. Because the cache is redis, an invalidation from the replica that did the write reaches every other replica, which a process-local memo would not.

Missing one of those entity types leaves the catalog silently stale, with the deny-all consequence above, so the set is the load-bearing part of the change and deserves a test per entity.

[...]

PR #3603, opened 2026-09-15
feat: batch permission check at POST /authorization/check

### LINKED ISSUE #3600: Batch permission check, so a public UI client needs no authorization catalog
## Why

`GET /authorization` has two consumer classes with opposite constraints.

A **resource server** is a confidential client. It can hold `PERMISSION_READ` through one `client-permission` row, and it genuinely needs the catalog: it evaluates ABAC over its own rows (`ATTRIBUTES` policy data) and lowers collection queries through `compile()`, neither of which a server-side check can answer without being handed the row.

A **public UI client** is the opposite. It has no secret, so it can obtain no `client_credentials` token and has no credential of its own to be gated on. But what it actually needs is far less than the catalog: coarse, attribute-free gating of nav items, buttons and routes.

Serving the second case by publishing the policy configuration to everyone is a poor trade. A batch check serves it better on every axis: the answer is authoritative rather than reproduced, no policy configuration leaves the server, and the version skew between a consumer's `@authup/access` and the server stops mattering.

## What

A batch counterpart to the existing `POST /permissions/:id/check`, which is `ForceLoggedIn` with no permission gate, so it is already a self-check any bearer can make.

**Evaluate every permission and answer the passing set**, rather than taking a list of names. That is the property that matters: a client naming what it checks has to maintain that list in step with its own UI, and a name missing from the request fails silently as a control that quietly disappears. With no list, a downstream app such as PrivateAIM/hub sends nothing and maintains nothing.

Open questions:

- **Realm — validate this first, it is the one that could sink the shape.** `usePermissionCheck` may carry `realmMatch`, and a flat set of names cannot answer "may I do X in realm R". Either the request takes the realm (one call per realm in play, which a console's realm switcher bounds) or the answer is per name and realm. Read hub's `usePermissionCheck` call sites before building: if they routinely carry a realm and the realm set is not small, the catalog looks better again for that consumer.
- **Cost.** One evaluation per permission per call, over an identity's grants loaded once. Most definitions share `system.default`, so this should be dominated by the grant load, but it wants measuring before it sits on every session resolve.
- **Shape.** `POST /permissions/check` alongside the per-row `POST /permissions/:id/check`, and whether `/policies` gets the same treatment.
- **Attributes stay out**, by construction. A check that depends on a resource row belongs to whoever holds the row.

## Status

The catalog route stays GATED. #3588 briefly made it anonymous to serve the public-client case; that commit is reverted on the branch (9d308d726), so nothing ships with the policy configuration public and there is no gate to flip back later.

Until this lands, a public UI client keeps the coarse name-only gating it has today. Replacing that fallback with an authoritative answer is the payoff here.

tada5hi:
Closes #3600.

## Why

`GET /authorization` has two consumer classes with opposite constraints. A resource server is a confidential client, can hold `PERMISSION_READ` through one `client-permission` row, and genuinely needs the catalog: it evaluates ABAC over its own rows and lowers collection queries through `compile()`. A **public UI client** is the opposite. It has no secret, so it can obtain no `client_credentials` token and has no credential of its own to be gated on, while what it needs is far less than the catalog: coarse gating of nav items, buttons and routes.

Serving that second case by publishing the policy configuration to everyone is a poor trade, so this route answers **verdicts** instead.

## What

```
POST /authorization/check
{ "names": ["user_update"]?, "realms": "own" | "ownOrNull" | ["<realm id>", null]? }

-> [{ "name": "user_update", "realms": ["<realm id>", null] }]
```

`ForceLoggedInMiddleware`, **no permission gate**. It discloses strictly less than the already-ungated `POST /permissions/:id/check` discloses one name at a time: answers about the caller's own authorization, no definition, no policy configuration, and no realm key the caller did not itself supply.

**Every permission by default.** Omitting `names` checks every definition, so a client never maintains a list in step with its own UI, where a name missing from the request would fail silently as a control that quietly disappears.

**The answer is per (permission, realm).** `realms` is `own`, `ownOrNull` (the default) or an explicit list, with `null` for the global rows. A symbolic selector resolves against the caller's own identity; an explicit list is taken and echoed **verbatim**, so the server resolves no realm key, discloses no realm's existence, and the caller matches the answer against its own input. There is deliberately no `any`: a caller that cares about another realm names it.

A permission reaching none of the requested realms is **absent**, so absent means denied, and a `realmMatch` naming a realm the request never asked about is fail-closed for free.

## Why it is cheap

A loop over the existing per-name path would be hundreds of statements: `PermissionDatabaseProvider.findOne` carries an uncached junction read plus a three-query tree walk each, and the binding evaluator re-reads the grants per name. Both halves of the fix already existed for the catalog, so the definitions are read once through `findDefinitions()` into a `PermissionMemoryProvider` and the grant load is hoisted into one memoized closure: two bulk reads plus one grant load, and the rest in memory.

The evaluator is wrapped in `RequestPermissionEvaluator` rather than having its PolicyData bag assembled by hand, which is what keeps the `global`-scope condition from being restated in a second place. A scope-restricted bearer is answered nothing, pinned by a spec.

## Route placement

`POST /permissions/check` is shadowed by `POST /permissions/:id` (routup returns matches in registration order and takes the first that responds), and `check` is a legal permission name. That is the collision this repo already argued and rejected for `/roles/schema`, and every literal POST sub-route in server-core sits on a `:id`-free controller. So it lives next to the catalog it substitutes for.

## The kit adopts it

`buildAuthorization` falls through to the check when `loadCatalog` answers 403 or 404, so a console's gating becomes authoritative where the name-only view it replaces is merely coarse. The name-only view survives as the last rung, for a server serving neither route. The memo is keyed by **subject**, not by credential: the catalog is identity-free and this answer is not.

**One visible behaviour change.** A bare `usePermissionCheck({ name })` now means "in at least one requested realm", so a `realmScope: none` grant stops passing. A realm-less pre-gate neutral-passes reach, so such a grant currently enables a control it can never use. That is a fix, but it shows up as a control disappearing.

## Review findings fixed in the second commit

- **Cost.** The realm cap was 8. Measured against 73 definitions: 7 ms for `ownOrNull`, 19 ms at eight, on a synchronous walk on an ungated route. Lowered to 4 (own realm, global rows, two named), which holds the worst case near the cost of the call the route exists to serve. The comment claimed a product ceiling the code never provided and now states the real bound, including the operator-controlled scaling of the names-omitted path.
- **Cross-identity memo.** A revalidation naming a different subject does not clean up, and the two consoles share one cookie session on one origin, so the memo needed the subject in its key.
- **Silent failure.** The grant load is one memoized promise, so a single failed read denied every pair at once and answered an authoritative empty set with nothing in the log. The verdict stays fail-closed and is now reported, the way the catalog route reports the same class of failure.
- **Docs.** The new Response Versioning convention was wrong twice (`GET /` also carries a top-level `version`, and neither it nor `GET /schemas` reports a rapiq version), and superseded name-only claims were left standing in `architecture.md`, `structure.md` and two doc-comments.

## Residuals, stated rather than papered over

- The answer is an **upper bound on what may be attempted**, never an entitlement: it is a pre-gate, so a grant restricted by a junction policy needing a resource row passes here and is still decided per row. The server stays the enforcement point.
- `findDefinitions()` is two uncached full-table reads, and this is its first ungated consumer. The names-omitted path therefore scales with the deployment's own catalogue (about 240 ms at 1000 global definitions), which is operator-controlled rather than caller-controlled. Capping it would refuse the console it was built for; the answer is the definition read's own cache, which is #3599.
- No counterpart on `POST /policies/:id/check`: the shape rests on the permission universe being enumerable, and policy names are operator-created and unbounded, so a no-subset form there would have no defensible default.

## Verification

`npm run build`, `npm run lint`, `npm run check:types --workspaces`, and the full suites: `packages/access` 229, `packages/core-http-kit` 36, `packages/client-web-kit` 296, `apps/server-core` 2626 passed. Each new spec was mutation-tested: dropping `realmMatch`, the subject key, or `decorate` fails it.

## Summary by CodeRabbit

- **New Features**
  - Added batch authorization checks for multiple permissions across requested realms.
  - Added authorization-check support to the API and client libraries.
  - Public clients can use permission verdicts when the authorization catalog is unavailable.
  - Added validation for authorization requests and responses, including supported realm selectors and size limits.

- **Bug Fixes**
  - Improved authorization fallback and cache invalidation when grants or scopes change.
  - Authorization failures now safely deny access when verdicts cannot be evaluated.

- **Documentation**
  - Documented the batch authorization endpoint and response versioning conventions.
  - Clarified that check results are upper-bound gating guidance, not entitlements.


tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai full review

tada5hi:
@coderabbitai pause

PR #3605, opened 2026-09-16
fix(server-core): scope the check routes to the caller, and gate checks for other subjects behind permission_check

### LINKED ISSUE #3604: POST /permissions/:id/check ignores the token's scope, so a scope-restricted bearer is over-reported
## What happens

`POST /permissions/:id/check` (and its `/policies/:id/check` sibling) builds its
`PolicyData` bag from `actor.identity` unconditionally. Every other request path
attaches the IDENTITY key through `RequestPermissionEvaluator.extendContext`,
which sets it **only** when the bearer's scopes include `ScopeName.GLOBAL`, so a
token holding no `global` scope fails the `system.identity` child of
`system.default` and is denied every bound permission.

The single-permission check therefore answers `success` for a bearer that the
same server denies on every ordinary route.

## Reproduction

Observed with one scope-restricted bearer against both routes:

```
PROBE single-check status=202 body={"status":"success"}
PROBE batch-check  status=200 body=[]
```

The batch route (`POST /authorization/check`, #3600) composes its evaluator
inside `request.decorate` -> `RequestPermissionEvaluator`, so it inherits the
scope condition and answers the denial the request path would give. The
single-permission route does not.

## Why it matters

It is an over-report rather than an over-grant: nothing is authorized by the
answer, and every real request stays enforced. But the route exists so a caller
can ask what it may attempt, and a consumer that trusts it will render controls
that 403, or conclude a scope restriction is not in force.

## Suggested fix

Have `PermissionCheckerService` (and the policy checker) take the request
evaluator rather than assembling the identity data itself, so the
`global`-scope condition is spelled exactly once. `POST /authorization/check`
is the shape to follow: it wraps the composed `PermissionEvaluator` in
`RequestPermissionEvaluator` instead of restating the rule.

Out of scope for #3600's PR (#3603): the fix touches
`PermissionCheckerService`, which that change does not otherwise go near.

tada5hi:
Closes #3604

## What was wrong

`POST /permissions/:id/check` and `POST /policies/:id/check` placed the actor's identity into the policy data themselves. Two things followed:

- **The token's scopes were ignored.** A bearer without the `global` scope was answered `success` while every ordinary route denied it (the issue as filed).
- **A body `identity` was evaluated as given, for any authenticated caller.** The permission-binding policy loads the grants of whatever identity the data names, so anyone logged in could read whether another user or client holds a permission, in any realm. Confirmed before the fix: an ungranted user naming the admin's identity got `success` for `user_update`.

## Who a check is for: the data's own `identity` key

Both controllers pass the body through unchanged, with an actor whose identity follows the scope rule (`useRequestPolicyIdentity(event)`, the one place the `global` condition is spelled; `RequestPermissionEvaluator` uses it too). Both checkers keep `master`'s signature `check(idOrName, data, actor, realm)`. `buildCheckData` (`core/identity/permission/checker/data.ts`) decides:

1. **An identity in the data**: it is evaluated as given. If it is not the requester's own (by `type` and `id`), the requester needs the new `permission_check` permission:
   - pre-evaluated first (403), so a caller without the grant learns nothing about which subjects exist;
   - then the subject is looked up (404) and `permission_check` is evaluated with `realmMatch` set to its STORED realm. The lookup is for the gate alone: the grant lookup loads by `id` only, so a realm written into the data could not bound it, and a `realm_admin` checks subjects of its own realm only;
   - a foreign identity is named as `{ type: 'user' | 'client', id: <UUID> }` (400 otherwise), since names repeat across realms.
2. **No identity**: the requester. A token without `global` hands over no identity, so its check runs without one, and it never names "its own" identity either, so naming one always needs the gate.
3. **`identity: null`**: the check runs without any identity.

Both routes then evaluate that data on a plain `PermissionEvaluator` / `PolicyEngine`, since the data already says who is evaluated. The resolved permission row is evaluated (its `realmId` and `clientId` ride the evaluation context), not a global permission of the same name. Refusals throw from `safeCheck` as HTTP errors rather than an embedded `status: "error"`. Provisioning grants `permission_check` like every permission: `admin` in every realm, `realm_admin` for its own realm.

Rejected: wrapping the policy in a memory-provider `PermissionEvaluator` so both routes share one path. It overwrites a caller's `permissionBinding` and turns `PolicyError` into `PermissionError`.

## Behaviour change for API callers

Documented in `docs/src/guide/deployment/upgrading.md`: naming another subject needs `permission_check` reaching its stored realm and a UUID; a scope-restricted bearer is denied and needs the grant even for its own identity; a realm- or client-scoped permission is evaluated as itself.

## Tests

Every new case was watched failing on the pre-fix code:

- Both HTTP checker specs: a scope-restricted bearer, bare (`error`) and naming its own identity (403); a subject answered for its own grants; a caller without the grant refused for another subject, including an unknown UUID (403, not 404), and answered for its own identity; `identity: null` evaluated without an identity.
- Permission route only: an unknown subject (404), a malformed type and a name (400), a `permission_check` grant at `own` reach answering for a subject in its realm and refused (403) for one in another realm, also when the body claims the checker's realm, and a realm-scoped permission evaluated for the caller and for a subject.
- Core specs compose a real `RequestPermissionEvaluator` over a real `PermissionEvaluator`, and pin the requester default, `null`, and that the data's reference object is never rewritten.
- Mutations each fail the relevant cases: dropping the subject realm from the gate; gating on a realm in the data; evaluating the subject on the request evaluator; looking the subject up before the pre-gate; passing either checker the actor without the scope rule.
- New helper `createScopeRestrictedClient` (`test/utils/token.ts`) re-signs the admin's claims without `global`, since the password grant always stamps it.

Full `apps/server-core` suite: 228 files, 2658 tests passed. `check:types`, `tsc -p tsconfig.build.json` and eslint pass.

## Not in this PR (already present before it)

- Both routes resolve another realm's permission or policy ROW by UUID (or by name under `/realms/:realmId`) with no reach check, so the error type leaks existence.
- The policy route evaluates the bare policy row without extra attributes or children, so composites fail and some policy configuration is ignored.

## Summary by CodeRabbit

* **New Features**
  * Added `permission_check` authorization for checking another user’s or client’s permissions.
  * Checks now support caller, no-identity, and explicitly selected identity evaluations using UUIDs.
  * Permission and policy checks honor the subject’s stored realm and scope-specific permissions.

* **Bug Fixes**
  * Improved validation and error responses for malformed, unknown, unauthorized, or cross-realm subjects.

* **Documentation**
  * Updated upgrade and API guidance covering identity handling, authorization requirements, scopes, and response behavior.


tada5hi:
@coderabbitai pause

PR #3606, opened 2026-09-16
fix: narrow a user's client-owned grants to the token's client

### LINKED ISSUE #3597: Client-scoped permissions are not scoped to the token's client for user identities
A permission may be owned by a client (`auth_permissions.client_id`), and provisioning lets a client declare its own permissions and roles. For a **user** identity that ownership is never enforced: a user holding such a permission holds it on every token, including tokens issued to a different client. There is no way to say "this permission applies when the user acts through application X".

## What happens today

`IdentityPolicyData.clientId` means the SUBJECT's own client ownership, and it has exactly one consumer, `IdentityPermissionProvider.reduceBindingsByIdentityClient` (`apps/server-core/src/core/identity/permission/module.ts`), which keeps only permissions whose own `client_id` equals it:

```ts
return bindings.filter((binding) => binding.permission.clientId === identity.clientId);
```

Its state per identity type:

- **user**: the filter runs, but `toIdentityPolicyData` resolves `clientId` to `null` for every user (no user row carries a client), so it returns every binding unchanged.
- **client**: `getForClient` does not call it at all.
- **role**: live. The superset checks build a `{ type: 'role', clientId: role.clientId }` identity, so a client-owned role's grants are narrowed to that client's permissions.

The `identity` policy type does not read `clientId` either; it compares `types` only (`packages/access/src/policy/built-in/identity/evaluator.ts`). So for users the client dimension of a permission is inert at request time.

The token itself is not the missing piece: `client_id` is already a claim on every token payload and both introspection endpoints echo it. What is missing is anything that uses it for authorization.

## Why the obvious fix is wrong

Feeding the token's `client_id` into `IdentityPolicyData.clientId` looks like a one-line change and is not. The filter is an equality, not a disjunction, so every global permission (the whole provisioned catalogue) would drop the moment the value is non-null. That regression already shipped once on the introspection path and was fixed by projecting over the resolved identity instead (`resolveIntrospectionSubject`, `apps/server-core/src/core/oauth2/introspection/module.ts`).

## What a real fix needs

1. **The narrowing becomes a disjunction**: global OR this client's. As an equality, a client-scoped permission can never coexist with the built-in catalogue.
2. **A carrier separate from `IdentityPolicyData.clientId`**. The token's client is a property of the CREDENTIAL, not of the subject, so it belongs on the actor context next to `scope`. Overloading the subject's own field puts the two meanings back on one name, which is what produced the regression above.
3. **The same narrowing on both introspection endpoints.** `GET /authorization` is identity-free and the grants ride introspection, so a console or a resource server evaluating locally (`createAuthorizationEvaluator`) would otherwise reach a different verdict than the server.

## Scope

Admission control is a separate question and already answered: `accessPolicyId` decides who may obtain a token for a client, and `scope` narrows what a token may do. This issue is only about whether a user's grants are narrowed by the application the token was issued to.

Non-goals: changing how client identities resolve their own permissions, and changing the role-superset narrowing, which is correct as it stands.

tada5hi:
Closes #3597.

A permission or a role owned by a client now applies to a **user** only through a token issued to that client. Through a token issued to client Y, the server withholds the user's permissions owned by client X, and the X-owned roles together with the global permissions they carry. Unowned grants always apply.

## How it works: the grants a token carries, resolved once

`IdentityPermissionProvider.getForToken(token)` is the one place the narrowing happens. From the token's `sub`, `sub_kind` and `client_id` it:

1. loads the user's roles and drops those owned by another client;
2. loads those roles' permissions, plus the user's direct ones;
3. drops every permission owned by another client, however it is held.

Admin and realm_admin receive every client-scoped permission through their roles, which is why step 3 is needed.

It is a disjunction (unowned OR the token's client), since an equality drops the whole global catalogue. `getFor(identity)`, `getRolesFor(identity)` and the token issuer are unchanged, and nothing passes the client alongside the subject. Everything else reads that one grant set:

- **Request evaluation.** The authorization middleware composes the request's evaluator per request, over an engine whose grant source is `createGrantsResolver`:
  - the request's own subject resolves through `getForToken` with the verified bearer's payload, memoized for the request;
  - any other subject (a check made on another's behalf) and any request without a token resolve through `getFor`.

  `POST /authorization/check` reads the same source through a required `grants` field, which it takes for the same reason it requires `decorate`. `POST /permissions/:id/check` and `POST /policies/:id/check` evaluate on engines of their own (since #3605); `createCheckerGrantSource` makes them read the actor's grants when they evaluate the actor, and a foreign subject's as stored.
- **Delegation.** `buildActorContext` exposes that source as `ActorContext.grants`. `isSuperset(parent, child)` and `resolveJunctionGrant(bindings, options)` take grant lists, never an identity: the services pass the actor's grants, and `isSuperset` compares them with the role's grants the service loads. The services read them through `AbstractEntityService.getActorGrants`, and a missing resolver yields no grants, so a mistake denies instead of widening.
- **Introspection.** `resolveIntrospectionSubject({ token, active })` projects `getForToken` for the token it describes. `POST /token/introspect` passes the introspected token, never the caller's; `GET /sessions/@me/introspect` passes the request's own, which for the console cookie is its subject alone. A local `createAuthorizationEvaluator` therefore reaches the server's verdict.

Unchanged by design: client subjects, and the equality in `getForRole`.

## Decisions worth reviewing

- **A request without a token issued to a client is not narrowed.** That covers Basic authentication, a password grant without a client, and the served console's cookie. Making those global-only would stop the served console and Basic from binding client-scoped permissions, since the member gates evaluate the client-scoped definition. It would also close nothing the `POST /authorize` residual below leaves open.
- **The system console clients are not exempted.** A standalone-hosted admin console, and its vite dev server, holds `admin-console` tokens and is narrowed. From there, binding another client's permissions or assigning its roles now answers `403`. An exemption would be worse: `admin-console` and `account-console` are public and auto-consenting, so with the `POST /authorize` residual any application holding a user token could regain the un-narrowed grants.
- **The request's evaluator and engine are built per request.** A cheap construction, the same pattern `POST /authorization/check` already used. In exchange, a request resolves its token's grants once instead of once per evaluation.

## Behaviour change

Written up in `docs/src/guide/deployment/upgrading.md` under the next release. An application that exercised a user's grants owned by another client, or an administrator binding client-scoped permissions through another client's token (including a standalone admin console), now receives `403`. A deployment whose permissions are owned by an API client while users sign in through a separate front-end client loses those permissions on the front end's tokens, including in what `POST /token/introspect` reports to the API; the note says so.

## Stated rather than guarded

These are listed in `.agents/architecture.md`, section *A user's client-owned grants apply through that client's tokens*:

- `POST /authorize` and device approve accept any user bearer, so a holder of a Y token can mint an X token for a public client X. This is the one real bypass left. It belongs to the issuance surface and is tracked in #3608.
- The application access-policy engine still holds the bare provider; it evaluates with IDENTITY data only, so it never reads grants.
- **Pre-existing, not introduced here:** assigning a client-owned role checks none of the global permissions it carries (`getForRole` keeps only the role client's own permissions, and an empty child passes `isSuperset`), so an actor holding only `USER_ROLE_CREATE` can assign itself an owned role carrying any global permission. Reproduced locally; tracked in #3607.
- An actor holding `ROLE_UPDATE` / `PERMISSION_UPDATE` can change a row's `clientId`. An actor acting through an X token can assign itself an unowned role that carries grants it holds only through X. Either change reaches a user's grants after the 60 s owned-roles / owned-permissions cache.

## Overlap with #3604 (PR #3605, merged into this branch)

#3605 answers the check routes on engines of their own, over the bare provider, so without this branch's `createCheckerGrantSource` a Y token checking `role_read` for itself would have been told `success` while every other surface says no. A check for another subject under `permission_check` resolves that subject's own grants.

## Tests

- `test/unit/http/controllers/security/token-client-narrowing.spec.ts` (new, HTTP) sets up clients X and Y, a user holding an X-owned permission, an X-owned role and a global role, and tokens for X, Y and no client. It covers:
  - introspection by bearer, and by admin Basic (narrowed by the introspected token);
  - `GET /sessions/@me/introspect`;
  - a route gate and the batch check;
  - delegation on `/user-roles` and `/user-permissions`.

  6 of its 9 cases failed before the fix. The 3 that passed are the positive and fail-open controls.
- `test/unit/core/entities/junction-actor-grants.spec.ts` (new) checks that all 10 delegation call sites hand the provider the actor's grants. Mutation-checked: pointing one site back at a fresh identity lookup fails its row.
- `provider.spec.ts` adds `getForToken` cases: token client X, token client Z, no client, a client subject, a token naming no subject, and a delegation check over token grants.
- `authorization.spec.ts` (middleware) pins the request grants resolver: the token subject resolves once per request (actor grants included), another subject and a token-less request resolve through `getFor`.
- The HTTP spec adds `POST /permissions/:id/check` (success / error / success for X / Y / no client); mutation-checked against the checker grant source. The existing superset and junction-grant cases now pass the actor's grants. The introspection regression pin (a direct global grant survives on a token minted for a client) is kept.
- `npm run test` passes for `apps/server-core` (2687 passed, 16 skipped for the MySQL/Postgres-only specs) and `packages/access` (232). `check:types` is clean for access, server-kit, server-test-kit, server-core, server-adapter-kit and client-web-kit.

## Summary by CodeRabbit

* **Security**
  * Client-owned permissions and roles now apply to users only through tokens issued to that client.
  * Grants from other clients, including permissions inherited through roles, are withheld.
  * Cross-client role and permission assignments are rejected with `403 Forbidden`.

* **Authorization**
  * Request authorization, permission checks, policy checks, delegation, and introspection now consistently evaluate the grants carried by the active token.
  * Basic authentication, clientless password grants, and served-console sessions remain unrestricted by client narrowing.

* **Documentation**
  * Added upgrade guidance and deployment notes covering token-based grant behavior and affected admin-console workflows.


tada5hi:
@coderabbitai full review

PR #3611, opened 2026-09-18
fix(server-core): refuse a bearer issued to a client on POST /authorize and device approve

### LINKED ISSUE #3608: POST /authorize and device approve accept a bearer issued to another client
## Summary

`POST /authorize` and `POST /device_authorization/approve` accept any user bearer, including an access token issued to another client. A holder of a token issued to client Y can therefore mint an authorization code (and then a token) for a public client X. Since #3597 a user's client-owned grants apply only through that client's tokens, so this route turns a Y token back into the X-owned grants the narrowing withholds.

## Where

- `HTTPOAuth2Authorizer.authorizeWithRequest` (`apps/server-core/src/adapters/http/adapters/oauth2/authorization/module.ts`) takes `useRequestIdentityOrFail(event)` and issues the code for it, whatever client the request's bearer was issued to.
- `DeviceAuthorizationController.approve` (`apps/server-core/src/adapters/http/controllers/workflows/device-authorization/module.ts`) does the same with `useRequestIdentityOrFail(event).raw`.

Both are reachable with a bearer only: the console session cookie is already refused on the issuance surface (`isOAuth2IssuancePath`).

## Exploit shape

1. The user signs in to application Y (client Y) and Y holds the access token.
2. Y posts `/authorize` with that bearer, `client_id=X` (a public client, `builtIn` or with covering consent), its own PKCE challenge and a `redirect_uri` matching X's patterns.
3. The response body carries the redirect URL with the code, so no browser navigation is needed; Y redeems the code at `/token` with the verifier. A `builtIn` public client such as `admin-console` / `account-console` needs no consent, and its `redirect_uri` patterns cover every trusted app origin.

The resulting token is issued to X and carries X's client-owned grants for the user.

## Proposed fix

Refuse a bearer that carries a `client_id` on both routes (`login_required` on `/authorize`, the same on approve). Every hosted flow already authenticates with a clientless bearer:

- the hosted password login (`LoginForm` sends no `client_id`),
- the MFA ticket completion (the ticket inherits the password grant's client, which the hosted form never sets),
- the federated handoff (`OAuth2FederatedLoginService.completeHandoff` mints without `client_id`),
- the device page, which logs in through the same `LoginForm`.

Basic authentication carries no token and stays accepted. To check before implementing: any downstream caller posting `/authorize` with a client token (none in this repository), and the spec coverage of the three hosted token paths above so a later change cannot give them a client silently.

tada5hi:
Closes #3608.

## The hole

`POST /authorize` and `POST /device_authorization/approve` took any user bearer. A holder of a token issued to client Y could mint a code — and then a token — for a public client X, which turns the X-owned grants #3597's narrowing withholds from Y back on. For a `builtIn` client (`admin-console`, `account-console`) there is no consent step, and the POST answers with the redirect URL in its body, so no browser is involved either.

## The fix

`assertTokenMayAuthorize(event)` (`adapters/http/request/helpers/token.ts`) refuses a bearer carrying a `client_id`, called at the top of `HTTPOAuth2Authorizer.authorizeWithRequest` and of `DeviceAuthorizationController.approve`, through a new `OAuth2LoginRequiredError.tokenClientBound()` beside its three siblings (every other site of that error goes through a factory, and the shared wire code already covers four causes — `message` is the only thing separating them in a log). `login_required` / 400, the shape the realm binding next to it already answers with. Nine lines of code; no core type, no option and no gate touched.

Only the user AT the authorization server may authorize an application, and their token there carries no client. Traced end to end for all four hosted paths:

| path | where the clientless property comes from |
|---|---|
| hosted password login | `store.login` has one call site and `StoreLoginContext` has no client field; the form forwards only `codeRequest.realm_id`, never its `client_id` |
| MFA-ticket completion | inherits the ticket's, which inherits the credential POST's |
| federated handoff | `completeHandoff`'s issue payload has no such key, and neither does any type on the path |
| device page (password and, since #3610, identity-provider) | the same two mechanisms, not a third |

Basic carries no token, so a session-less authorize is unaffected, and `lookup` / `deny` keep taking any user bearer: both need the `user_code`, and neither mints anything.

This is the bearer half of the refusal `isOAuth2IssuancePath` already makes for the console cookie, which is denied on this same surface, for this same threat, with the `account-console` client named in its comment.

## Why not compare for equality?

Equality (refuse only when the bearer's client differs from the client being authorized) is the semantically obvious invariant, and it does close the reported cross-client escalation — I could not construct a bypass of it. It was not taken for three reasons, in order of weight:

1. **It leaves the same-client escalation open.** A client re-authorizing *itself* gains no grant it does not already hold, but it gains scope and lifetime. `resolveGrantedScope` admits any request asking for `global` whatever the client has bound, the server *records* consent rather than gating on it, and the exchange answers with a fresh refresh chain. So equality would let an application turn its own 15-minute `openid` token into a 3-day `global` one, past the consent the user actually gave, with no browser and no consent screen. Refusing outright closes that too.
2. **It is the larger change.** Refusing needs nothing but the request, so it lives in the HTTP adapter next to `useRequestTokenPayload`. A comparison needs the *resolved* client, which on the device path only core knows (it is behind the `user_code`, in the code blob), so it would have to be threaded as a new option through `OAuth2AuthorizationOptions` *and* `OAuth2DeviceAuthorizationApproveOptions` into the gate — six files against three, plus a fail-closed subtlety (`options.client` is optional, so the comparison must be `!== options.client?.id`, never `a && b && a !== b.id`).
3. **Nothing needs the same-client case.** There is no in-repo caller, spec or shipped client flow that authorizes with a client-bound bearer, and no documented one either: the public docs describe `/authorize` only as a hosted page the browser is navigated to. A caller that already holds an X token has nothing to gain from minting another.

Equality remains a one-line change from here (`!== options.client?.id` in the gate) if a legitimate same-client caller ever turns up.

## Tests

Three new cases, all three red before the fix and green after (verified by short-circuiting the assert: exactly these three fail, nothing else):

- `security/token-client-narrowing.spec.ts` gains `describe('issuance')` — this is #3597's own security spec and #3608 was its stated residual, so its `beforeAll` already builds clientX / clientY / the three tokens. Another client's token refused, **its own token refused** (the one case that discriminates the two rules, so the choice is pinned rather than implied), a clientless token still issued a code.
- `device-authorization.spec.ts` gains *should refuse an approval presented with a token issued to a client*, next to the existing *should refuse a console session cookie on the approval*: refused, the device code left pending (`authorization_pending` on the next poll), and the hosted page's own clientless login still approves.
- `federated-login/module.spec.ts` pins the missing half of the premise, next to the existing `acr` assertion: the handoff token carries no `client_id`. It was the one hosted path whose clientless property nothing asserted, and it is the path the fix most depends on.

One committed test needed correcting. `account-console-session.spec.ts`'s *attributes the forced account-console flow per token* drove step 2 with the admin console's own token while its own comment said the bearer is the one the shared IdP-origin cookie jar holds. The account console kicks with `window.location.href = buildAuthorizeURL(...)` — a navigation — so the consent POST rides the hosted page's bearer, i.e. the clientless login of step 1. It now presents that, which is what the comment claimed all along; what the test is actually about (per-token attribution across one shared session) is unchanged.

Full `apps/server-core` suite green (2716 passed, 16 skipped), `check:types` and `eslint` clean.

## Residuals

Found while sweeping every mint site for the same laundering class; none is introduced here, and none is in scope for #3608:

- **`resolveGrantedScope` admits `global` on request regardless of what the client has bound** (`core/oauth2/scope/helpers.ts`), which makes a client's bound scope set non-binding at `/authorize`. It is deliberate as written, documented and test-pinned, so it is left alone — but it is what turns the scope boundary into something self-liftable, and it is the reason point 1 above bites. Worth its own issue if that was not the intent.
- **`POST /device_authorization/lookup` lets an application holding a user's token burn that user's 10-per-10-minutes lookup throttle**, locking them out of their own genuine device verification. A nuisance, not an escalation — it mints nothing — which is why `lookup` stays ungated here.
- **A pre-fix exploit leaves durable state the fix does not undo**: a covering `auth_consents` row and an `AUTHORIZE` audit row that both read as a user consent at the authorization server.
- **Wiring `HTTPOAuth2IdentityGrantType`** (unregistered dead code) would mint a clientless `global` pair straight from the request identity and defeat this check as well as #3597's narrowing. The assert's doc comment says so at the mechanism.

## Docs

- `.agents/architecture.md`: new *Only the user at the authorization server may authorize an application* under *Authorize Realm Binding* (it asks which credential presented the identity, where the realm binding asks whose identity it is); a *confused deputy across clients* row in the device threats table; and the #3597 section's now-closed `POST /authorize` residual bullet rewritten into the one that remains (the served consoles authenticate with a cookie, which narrows nothing), plus the stale clause that justified not exempting the console clients by pointing at this residual.
- `docs/src/guide/deployment/upgrading.md`: an action-required entry, since this is a behaviour change on a public route.

## Summary by CodeRabbit

- **Security**
  - Authorization and device-approval requests now reject bearer tokens issued to a client, returning `login_required`.
  - Clientless user sessions continue to authorize applications normally.
  - Device lookup and denial actions remain available with bearer tokens.

- **Documentation**
  - Added upgrade guidance describing the behavior change and integrations that may require updates.

- **Tests**
  - Added coverage for client-bound token rejection and valid clientless authorization flows.


tada5hi:
@coderabbitai full review

PR #3613, opened 2026-09-18
feat(authup): add device login and authenticated CLI API access

### LINKED ISSUE #3592: `authup login`: sign the CLI in through the device authorization grant
## What it is

The device grant shipped in #3587 with the typed client surface a CLI needs and nothing else. An `authup login` command is one loop over that client and needs no server change:

1. `client.deviceAuthorization.create({ client_id })` against a client the operator lists the URN on (`grantTypes` must name `urn:ietf:params:oauth:grant-type:device_code`; a null column does not enable it).
2. Print `verification_uri_complete` and the formatted `user_code`.
3. Poll `client.token.createWithDeviceCode({ device_code, client_id })` every `interval` seconds: continue on `authorization_pending`, add 5 s on `slow_down` (`isOAuth2DeviceAuthorizationError` + `code`), stop on `expired_token`, `access_denied` or `invalid_grant`.
4. Store the grant (where, and how a later command reuses and refreshes it, is the design question of this issue).

## Open questions

- Which client the CLI presents by default (a provisioned per-realm `cli` client would need a system-client definition; an operator-supplied `--client-id` needs none).
- Credential storage on the operator's machine (the OS keychain against a dotfile).
- Whether `healthcheck` / `config validate` stay unauthenticated, so `login` is only for commands that will call the management API.

## Pointers

- `packages/core-http-kit/src/domains/workflows/oauth2/device-authorization/` (`create`, `lookup`, `approve`, `deny`)
- `packages/core-http-kit/src/domains/workflows/oauth2/token/module.ts` (`createWithDeviceCode`)
- `docs/src/guide/development/api-oauth2.md`, section 8
- `apps/authup/src/commands/` (the command shape)

tada5hi:
Adds device-flow login and authenticated API access to local or remote Authup servers, with prefixed resource commands and a raw HTTP command sharing the same authentication and refresh handling.

```sh
export AUTHUP_SERVER_URL=https://auth.example.com/
authup login --client-id <public-client-uuid>
authup resource users list --query 'page[limit]=10'
authup resource users create --data '{"name":"alice","realmId":"<realm-uuid>"}'
authup resource users update <user-uuid> --data '{"displayName":"Alice"}'
authup api 'users?filter[name]=alice'
authup logout
```

`resource` supports list/get/create/update/delete for users, clients, realms, roles, permissions, policies and scopes. `api` handles other endpoints and workflows. `--server` overrides `AUTHUP_SERVER_URL`, then defaults to `http://localhost:3000/`. The operator supplies a public client with the device grant explicitly enabled; no server provisioning changes are needed. Login prints the verification URL and code, honors polling/backoff, and stops on denial or expiry.

Credentials default to the OS keychain through optional, lazily loaded `@napi-rs/keyring`. Linux explicitly selects persistent Secret Service. Headless environments can select `--credential-store=file` or `AUTHUP_CREDENTIAL_STORE=file`; keychain errors never silently downgrade to unencrypted files. The file store retains private permissions and atomic writes. Stores are independent and scoped by server URL; logout removes the selected local entry only.

Expiring tokens are refreshed and rotated credentials are saved before an API request. Per-server locking prevents concurrent rotation races. Ctrl-C cancels polling and requests, but a completed rotation is persisted before releasing the lock. API paths cannot escape the configured base URL, redirects are refused, remote servers require HTTPS, and failed mutations are not replayed. Healthcheck and configuration validation remain independent of login.

Commands live in separate files. Shared request handling lives in `remote/request.ts`; session types, lifecycle/locking, file storage and keychain storage have dedicated modules under `remote/session/`.

Validation: 113 CLI tests pass, including mocked keychain failures and cancellation during persistence; CLI type check, build, and changed-file ESLint pass. The native keyring module loads on the development machine. A built-binary loopback smoke test covers explicit file fallback, login, refresh, resource creation, raw API redirect refusal, and logout. Actual OS keychain operations are mocked in tests; Windows and Linux keychain integration have not been exercised locally.

Closes #3592


tada5hi:
Superseded by #3614: an operator-supplied client remembered per host instead of a required --client-id on every login, gh-shaped keychain storage with --insecure-storage instead of a store selector on every command, and entity commands derived from core-http-kit's registry instead of a hand-listed resource set and a raw request passthrough.

